PR #5, opened 2015-04-27
adds CmdLine() exception handler

jhseol:
If /proc/pid/cmdline has no content, slice operation of the 'data' will be panicked. (e.g. kernel threads)


beorn7:
:+1: but let's have @grobie take a look to verify.

proc.go
```diff
@@ -95,6 +95,10 @@ func (p Proc) CmdLine() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	
+	if len(data) < 1 {
+		return nil, fmt.Errorf("could not read cmdline pid:%d", p.PID)
```
juliusv: Could you do: `"could not read cmdline for pid: %d"`

juliusv: Actually, should this really return an error, or just an empty command-line?

jhseol: Hmm, just return an empty command-line make more sense to me.

juliusv: Yeah. Curious - are you using this function from somewhere? Seems like the "prometheus" github org is not even using it anywhere yet.

grobie: :+1: for an empty string

grobie:
Please add a test case as well.

proc.go
jhseol: Okay, I'm building some kind of system event listener including docker and kernel proc events. 
As a part of the listener, I found this repository for proc fs reader. 
And we're planning to use prometheus for our system metric server. :)

jhseol:
I added a test case for an empty cmdline file.

proc.go
```diff
@@ -95,6 +95,10 @@ func (p Proc) CmdLine() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	
+	if len(data) < 1 {
+		return nil, nil
```
grobie: I think we should return `[]string{}, nil` here, as the assumption in go for `type, error` return signatures is usually: if `err` is `nil` then the returned type is an initialized value.

proc_test.go
```diff
@@ -44,12 +44,25 @@ func TestCmdLine(t *testing.T) {
...
+
+	p2, err := testProcess(26232)
+	if err != nil {
+		t.Fatal(err)
+	}
+	c2, err := p2.CmdLine()
+	if err != nil {
+		t.Fatal(err)
+	}
+	var want []string
```
grobie: What I "want" here is actually `[]string{}`.

jhseol:
Thanks for your comment.

grobie:
:+1: Great, thanks! Now comes the last pedantry thingy ;), we usually tend to squash our commits together at the end of a PR review, so that we have only a few coherent commits with all changes to code+tests+fixtures together. I'll merge immediately after squashing them to a nice commit.

jhseol:
:+1: Great!

juliusv:
@jhseol Hey hey, but the commits are still not squashed - could you squash them? Then we'll merge :)

jhseol:
Commits are squashed. Is it okay to merge?

juliusv:
@jhseol Hmm, somehow the "Update example" commit by @audebert made it into your squashed commits (https://github.com/prometheus/procfs/pull/5/commits), but under a different SHA than in master. This should fix it:

```
git checkout master
git reset --hard origin/master
git cherry-pick 99d4f8f  # The one commit we're actually interested in.
git push -f origin master
```

jhseol:
Commits are squashed again. I'm not sure why "Update example" commit was squashed with my commit. :(

juliusv:
@jhseol no worries - looking good now :) Thanks! Merging.

PR #6, opened 2015-05-02
Added AbsCmdLine to return absolute pathe of the command

armen:


grobie:
Awesome, thank you for the contribution! Can you please squash the commits together, will merge afterwards.

armen:
Sure, squashed the commits.

proc.go
```diff
@@ -103,6 +103,17 @@ func (p Proc) CmdLine() ([]string, error) {
 	return strings.Split(string(data[:len(data)-1]), string(byte(0))), nil
 }
 
+// AbsCmdLine returns absolute path to the command line of a process.
+func (p Proc) AbsCmdLine() (string, error) {
+	ac, err := p.readlink("exe")
```
juliusv: Hmm, `exe` does not contain the process's command line, just the first part of it (the executable). Maybe it would make sense to rename this method to `Executable`?

armen:
Yeah, that makes sense

proc.go
armen: Yeah, that makes sense

armen:
Sent a pull request

PR #8, opened 2015-05-04
FEAT: add basic support for /proc/<pid>/io counters.

cournape:
May be useful when one tries to track per-process IO.

### DIFF
--- a/proc_stat_io.go
+++ b/proc_stat_io.go
@@ -1,0 +1,54 @@
+package procfs
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+// ProcIO models the content of /proc/<pid>/io.
+type ProcIO struct {
+	// Chars read.
+	RChar uint64
+	// Chars written.
+	WChar uint64
+	// Read syscalls.
+	SyscR uint64
+	// Write syscalls.
+	SyscW uint64
+	// Bytes read.
+	ReadBytes uint64
+	// Bytes written.
+	WriteBytes uint64
+	// Bytes written, but taking into account truncation. See
+	// Documentation/filesystems/proc.txt in the kernel sources for
+	// detailed explanation.
+	CancelledWriteBytes int64
+}
+
+// NewIO creates a new ProcIO instance from a given Proc instance.
+func (p Proc) NewIO() (ProcIO, error) {
+	pio := ProcIO{}
+
+	f, err := p.open("io")
+	if err != nil {
+		return pio, err
+	}
+	defer f.Close()
+
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		return pio, err
+	}
+
+	ioFormat := "rchar: %d\nwchar: %d\nsyscr: %d\nsyscw: %d\n" +
+		"read_bytes: %d\nwrite_bytes: %d\n" +
+		"cancelled_write_bytes: %d\n"
+
+	_, err = fmt.Sscanf(string(data), ioFormat, &pio.RChar, &pio.WChar, &pio.SyscR,
+		&pio.SyscW, &pio.ReadBytes, &pio.WriteBytes, &pio.CancelledWriteBytes)
+	if err != nil {
+		return pio, err
+	}
+
+	return pio, nil
+}
--- a/fixtures/26231/io
+++ b/fixtures/26231/io
@@ -1,0 +1,7 @@
+rchar: 750339
+wchar: 818609
+syscr: 7405
+syscw: 5245
+read_bytes: 1024
+write_bytes: 2048
+cancelled_write_bytes: -1024
--- a/proc_stat_io_test.go
+++ b/proc_stat_io_test.go
@@ -1,0 +1,49 @@
+package procfs
+
+import "testing"
+
+func TestProcIO(t *testing.T) {
+	fs, err := NewFS("fixtures")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p, err := fs.NewProc(26231)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	s, err := p.NewIO()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, test := range []struct {
+		name string
+		want uint64
+		got  uint64
+	}{
+		{name: "RChar", want: 750339, got: s.RChar},
+		{name: "WChar", want: 818609, got: s.WChar},
+		{name: "SyscR", want: 7405, got: s.SyscR},
+		{name: "SyscW", want: 5245, got: s.SyscW},
+		{name: "ReadBytes", want: 1024, got: s.ReadBytes},
+		{name: "WriteBytes", want: 2048, got: s.WriteBytes},
+	} {
+		if test.want != test.got {
+			t.Errorf("want %s %d, got %d", test.name, test.want, test.got)
+		}
+	}
+
+	for _, test := range []struct {
+		name string
+		want int64
+		got  int64
+	}{
+		{name: "CancelledWriteBytes", want: -1024, got: s.CancelledWriteBytes},
+	} {
+		if test.want != test.got {
+			t.Errorf("want %s %d, got %d", test.name, test.want, test.got)
+		}
+	}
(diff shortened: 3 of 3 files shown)


proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (p Proc) NewIO() (ProcIO, error) {
+	var rchar int64
+	var wchar int64
+	var syscr int64
+	var syscw int64
+	var read_bytes int64
+	var write_bytes int64
+	var cancelled_write_bytes int64
```
GeertJohan: You could avoid these allocations by creating a ProcIO here;
```go
pio := ProcIO{}

// ....

_, err = fmt.Sscanf(string(data), io_format, &pio.Rchar, &pio.Wchar, &pio.Syscr,
&pio.Syscw, &pio.Read_bytes, &pio.Write_bytes, &pio.Cancelled_write_bytes)
```

proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
+package procfs
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+type ProcIO struct {
+	// chars read
+	rchar int64
```
GeertJohan: I think these fields should be exported.

GeertJohan: If the values are guaranteed to be positive, maybe consider using `uint64`.

proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
...
+	syscr int64
+	// write syscalls
+	syscw int64
+	// bytes read
+	read_bytes int64
+	// bytes written
+	write_bytes int64
+	// bytes written, but taking into account truncation. See
+	// Documentation/filesystems/proc.txt for detailed explanation
+	cancelled_write_bytes int64
```
GeertJohan: Please consider using CamelCase for field names, e.g.: `CancelledWriteBytes`

proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
...
+	// read syscalls
+	syscr int64
+	// write syscalls
+	syscw int64
+	// bytes read
+	read_bytes int64
+	// bytes written
+	write_bytes int64
+	// bytes written, but taking into account truncation. See
+	// Documentation/filesystems/proc.txt for detailed explanation
```
GeertJohan: Linked Documentation/filesystems/proc.txt is missing.

proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
...
+	// chars read
+	rchar int64
+	// chars written
+	wchar int64
+	// read syscalls
+	syscr int64
+	// write syscalls
+	syscw int64
+	// bytes read
+	read_bytes int64
```
juliusv: Instead of under_scores in Go, please use camelCase.

And yes, these fields need to be exported (capitalized first letter), as otherwise they won't be visible outside of the package.

proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
+package procfs
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+type ProcIO struct {
+	// chars read
```
juliusv: Just a style nit: could you start all comments with capital letters and period at the end? (sentence-style)

proc_stat_io.go
```diff
@@ -0,0 +1,58 @@
+package procfs
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+type ProcIO struct {
```
juliusv: Exported fields will need a docstring. `golint` (`go get -u github.com/golang/lint/golint`) will catch this.

juliusv:
:+1: besides comments by @GeertJohan and me.

GeertJohan:
Yes. Thanks @cournape for adding this! :smiley:

cournape:
As you've noticed, I am a go beginner.

Everything should look ok now.

proc_stat_io.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+// ProcIO models the content of /proc/<pid>/io.
+type ProcIO struct {
+	// Chars read.
+	Rchar uint64
```
juliusv: I'd go for `RChar`, `WChar`, `SyscR`, ... (compare with existing stats in https://github.com/prometheus/procfs/blob/master/proc_stat.go). An alternative would be more verbose names like `CharsRead`, `CharsWritten`, etc., but I guess that would actually be inconsistent with https://github.com/prometheus/procfs/blob/master/proc_stat.go.

proc_stat_io.go
```diff
@@ -0,0 +1,54 @@
...
+		return pio, err
+	}
+	defer f.Close()
+
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		return pio, err
+	}
+
+	IoFormat := "rchar: %d\nwchar: %d\nsyscr: %d\nsyscw: %d\n" +
```
juliusv: Lower-case local variables (`ioFormat`).

juliusv:
@cournape No problem, I hope the reviews aren't too annoying :)

In the end, could you squash your commits into one? We don't really follow the `PREFIX: ...` commit-name convention anywhere in Prometheus, so I'd just omit that here.

cournape:
@juliusv no worries. Should I just create a new PR with this PR's commits squashed into one ? I prefer avoiding rewriting existing history

juliusv:
It's common to just squash in the existing PR.

The not-rewriting-history is an interesting question :) If a branch is created solely for review (which on GitHub is the case most of the time), rewriting is considered fine by almost anyone I know. If you think about it, in Linux kernel development (and thus when used as envisioned by Linus), history *within* reviews is also not kept. Instead, reviews happen not via GitHub, but via email, there. GitHub is a bit of a weird case because they centralize git again, and also base reviews off of branches instead of concrete patchsets, which is ugly in several ways (like when I'm merging a PR, how do I know I'm merging what I just reviewed vs. something that has been appended/rewritten in the last minute?).

Anyways, yeah - just squash over the current branch please :)

GeertJohan:
I always rewrite history in when squashing commits in a PR. Github handles it pretty well imo.

cournape:
@juliusv interesting. In the projects I am usually involved, rewriting history is almost always avoided, including PR, partly for the reason you've just mentioned.

But when in Rome do as Roman do :) I have just rebased the PR

juliusv:
:+1: Woot, thanks!

I guess this repo is becoming an externally used library now, and we'll have to keep that in mind when making changes to it later...

cournape:
@juliusv I used that code (outside procfs) for a small internal "proc_exporter" in a similar spirit to prometheus "node_exporter", for cases where in-app monitoring is not worth it.

GeertJohan:
@juliusv You might consider doing a versioned release using gopkg.in so, when required, you can break the API's in a v2.

juliusv:
@GeertJohan We could do that, although I've had trouble go-getting from gopkg.in in the past (found some article that advised against it, IIRC). I've also seen Go libraries just breaking their APIs in slight, easy-to-update ways. I guess for a small library like this that was never officially advertised as "please use me externally", that might be acceptable.

GeertJohan:
Oh? I've never had any problems with gopkg.in.. I personally expect all popular publicly released packages having their own repository (not being a sub-package in a command that was released) to not make breaking changes in their master, unless it was required to fix a security issue. This is not always the case ofcourse, some package do break their API on master. But those mostly have a "do not use, api can change" warning in the readme.
Ofcourse at this point the package is not very popular yet. But that's why now is a good time to switch to a system like gopkg.in, or maybe add a warning that API is subject to change.

juliusv:
For now I'm kind of with Dave Cheney on this (https://lists.ubuntu.com/archives/juju-dev/2014-July/002903.html) and would rather declare this as "use externally at your own risk" for the time being:

https://github.com/prometheus/procfs/pull/10

grobie:
Thanks! I renamed the source files to not longer include `stat` to follow the naming schema of procfs.

PR #9, opened 2015-05-08
Add IPVS metrics parsing.

matthiasr:
As suggested in prometheus/node_exporter#67, move the IPVS parsing into
this repository.

@juliusv

### DIFF
--- a/ipvs.go
+++ b/ipvs.go
@@ -1,0 +1,223 @@
+package procfs
+
+import (
+	"bufio"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"strconv"
+	"strings"
+)
+
+// IPVSStats holds IPVS statistics, as exposed by the kernel in `/proc/net/ip_vs_stats`.
+type IPVSStats struct {
+	// Total count of connections.
+	Connections uint64
+	// Total incoming packages processed.
+	IncomingPackets uint64
+	// Total outgoing packages processed.
+	OutgoingPackets uint64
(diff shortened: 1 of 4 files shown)


matthiasr:
ah just realized that this needs lots more exporting to be useful from collectors.

matthiasr:
@juliusv now it's ready. exported all fields, removed the unused fixture.

ipvs.go
```diff
@@ -0,0 +1,199 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"strconv"
+	"strings"
+)
+
+type IPVSStats struct {
```
matthiasr: Something I'm really not sure about is naming here: stats vs. stat vs status vs statuses?

juliusv: Stats sounds fine to me.

juliusv: I think `ProcStat` and `Stat` are called that way because the underlying file is called `stat`. Here it's `ip_vs_stats`, so you could argue that `Stats` makes sense here :)

ipvs.go
```diff
@@ -0,0 +1,199 @@
...
+	Connections, IncomingPackets, OutgoingPackets, IncomingBytes, OutgoingBytes uint64
+}
+
+type IPVSBackendStatus struct {
```
juliusv: Try running golint (https://github.com/golang/lint) - it will tell you to have docstrings for all exported types and fields :)

juliusv:
:+1: otherwise

juliusv:
:+1:

PR #10, opened 2015-05-11
Add notice about library stability.

juliusv:


juliusv:
@grobie See https://github.com/prometheus/procfs/pull/8

grobie:
I don't see why it should be limited to prometheus use. :+1: to add a note that this project is work in progress and that the public interface might change in backwards incompatible ways without warnings though.

juliusv:
@grobie updated the wording

GeertJohan:
:+1: good move

PR #11, opened 2015-06-16
Add Proc.FileDescriptorTargets() function

nickygerritsen:
For a project we are creating we needed something that could scan /proc/ID/fd and tell us what target files are used for each process.

Before trying to code everything else I found this project which already did almost everything I wanted, so I added the missing function.

We are going to use it even if this would not get merged, but I think it is nice to at least propose a merge for something like this :). I'm of course willing to change the implementation if something should be done differently.


proc.go
```diff
@@ -133,6 +133,28 @@ func (p Proc) FileDescriptors() ([]uintptr, error) {
...
+func (p Proc) FileDescriptorTargets() ([]string, error) {
+	names, err := p.fileDescriptors()
+	if err != nil {
+		return nil, err
+	}
+
+	var targets = make([]string, len(names))
+
+	for i, name := range names {
+		var target, err = p.readlink("fd/" + name)
```
juliusv: Style nit: we do `:=` instead of `var` when the assignment happens on the same line: `target, err := ...`.

proc.go
```diff
@@ -133,6 +133,28 @@ func (p Proc) FileDescriptors() ([]uintptr, error) {
...
+		if err != nil {
+			targets[i] = ""
```
juliusv: It'll be the empty string by default, so this case could be omitted.

juliusv:
:+1: Otherwise, provided you're ok with the API-breakage warning in the README.md.

nickygerritsen:
OK I changed the code for both your comments and pushed again.

Furthermore we are probably going to create a fork that we will update when we have time to fix API-breakage stuff :)

juliusv:
Thanks!

proc_test.go
```diff
@@ -93,7 +93,54 @@ func TestFileDescriptors(t *testing.T) {
...
+	fdsBefore, err := p2.FileDescriptors()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	s, err := os.Open("fixtures")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer s.Close()
```
grobie: Why do you open fixtures, it doesn't seem to be used afterwards.

grobie: Eh, ignore.

PR #12, opened 2015-09-14
adding mdstat-parser to procfs

cherti:
Adding functionality used in node_exporter's mdadm-exporter to procfs, including unittest for the parsing-functionality.

### DIFF
--- a/mdstat.go
+++ b/mdstat.go
@@ -1,0 +1,158 @@
+package procfs
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var (
+	statuslineRE = regexp.MustCompile(`(\d+) blocks .*\[(\d+)/(\d+)\] \[[U_]+\]`)
+	buildlineRE  = regexp.MustCompile(`\((\d+)/\d+\)`)
+)
+
+// MDStat holds info parsed from /proc/mdstat.
+type MDStat struct {
+	// Name of the device.
+	Name string
+	// activity-state of the device.
+	ActivityState string
+	// Number of active disks.
+	DisksActive int64
+	// Total number of disks the device consists of.
+	DisksTotal int64
+	// Number of blocks the device holds.
+	BlocksTotal int64
+	// Number of blocks on the device that are in sync.
+	BlocksSynced int64
+}
+
+// ParseMDStat parses an mdstat-file and returns a struct with the relevant infos.
+func (fs FS) ParseMDStat() (mdstates []MDStat, err error) {
+	mdStatusFilePath := path.Join(string(fs), "mdstat")
+	content, err := ioutil.ReadFile(mdStatusFilePath)
+	if err != nil {
+		return []MDStat{}, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+	}
+
+	mdStatusFile := string(content)
+
+	lines := strings.Split(mdStatusFile, "\n")
+	var currentMD string
+
+	// Each md has at least the deviceline, statusline and one empty line afterwards
+	// so we will have probably something of the order len(lines)/3 devices
+	// so we use that for preallocation.
+	estimateMDs := len(lines) / 3
+	mdStates := make([]MDStat, 0, estimateMDs)
+
+	for i, l := range lines {
+		if l == "" {
+			// Skip entirely empty lines.
+			continue
+		}
+
+		if l[0] == ' ' {
+			// Those lines are not the beginning of a md-section.
+			continue
+		}
+
+		if strings.HasPrefix(l, "Personalities") || strings.HasPrefix(l, "unused") {
+			// We aren't interested in lines with general info.
+			continue
+		}
+
+		mainLine := strings.Split(l, " ")
+		if len(mainLine) < 3 {
+			return mdStates, fmt.Errorf("error parsing mdline: %s", l)
+		}
+		currentMD = mainLine[0]      // name of md-device
+		activityState := mainLine[2] // activity status of said md-device
+
+		if len(lines) <= i+3 {
+			return mdStates, fmt.Errorf("error parsing %s: entry for %s has fewer lines than expected", mdStatusFilePath, currentMD)
+		}
+
+		active, total, size, err := evalStatusline(lines[i+1]) // parse statusline, always present
+		if err != nil {
+			return mdStates, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+		}
+
+		//
+		// Now get the number of synced blocks.
+		//
+
+		// Get the line number of the syncing-line.
+		var j int
+		if strings.Contains(lines[i+2], "bitmap") { // then skip the bitmap line
+			j = i + 3
+		} else {
+			j = i + 2
+		}
+
+		// If device is syncing at the moment, get the number of currently synced bytes,
+		// otherwise that number equals the size of the device.
+		syncedBlocks := size
+		if strings.Contains(lines[j], "recovery") || strings.Contains(lines[j], "resync") {
+			syncedBlocks, err = evalBuildline(lines[j])
+			if err != nil {
+				return mdStates, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+			}
+		}
+
+		mdStates = append(mdStates, MDStat{currentMD, activityState, active, total, size, syncedBlocks})
+
+	}
+
+	return mdStates, nil
+}
+
+func evalStatusline(statusline string) (active, total, size int64, err error) {
+	matches := statuslineRE.FindStringSubmatch(statusline)
+
+	// +1 to make it more obvious that the whole string containing the info is also returned as matches[0].
+	if len(matches) != 3+1 {
+		return 0, 0, 0, fmt.Errorf("unexpected number matches found in statusline: %s", statusline)
+	}
+
+	size, err = strconv.ParseInt(matches[1], 10, 64)
+	if err != nil {
+		return 0, 0, 0, fmt.Errorf("%s in statusline: %s", err, statusline)
+	}
+
+	total, err = strconv.ParseInt(matches[2], 10, 64)
+	if err != nil {
+		return 0, 0, 0, fmt.Errorf("%s in statusline: %s", err, statusline)
+	}
+
+	active, err = strconv.ParseInt(matches[3], 10, 64)
+	if err != nil {
+		return 0, 0, 0, fmt.Errorf("%s in statusline: %s", err, statusline)
+	}
+
+	return active, total, size, nil
+}
+
+// Gets the size that has already been synced out of the sync-line.
+func evalBuildline(buildline string) (int64, error) {
+	matches := buildlineRE.FindStringSubmatch(buildline)
+
+	// +1 to make it more obvious that the whole string containing the info is also returned as matches[0].
+	if len(matches) < 1+1 {
+		return 0, fmt.Errorf("too few matches found in buildline: %s", buildline)
+	}
+
+	if len(matches) > 1+1 {
+		return 0, fmt.Errorf("too many matches found in buildline: %s", buildline)
+	}
+
+	syncedSize, err := strconv.ParseInt(matches[1], 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("%s in buildline: %s", err, buildline)
+	}
+
+	return syncedSize, nil
+}
--- a/mdstat_test.go
+++ b/mdstat_test.go
@@ -1,0 +1,33 @@
+package procfs
+
+import (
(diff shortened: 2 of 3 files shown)


mdstat.go
```diff
@@ -0,0 +1,157 @@
...
+
+)
+
+var (
+	statuslineRE = regexp.MustCompile(`(\d+) blocks .*\[(\d+)/(\d+)\] \[[U_]+\]`)
+	buildlineRE  = regexp.MustCompile(`\((\d+)/\d+\)`)
+)
+
+type mdStatus struct {
+	mdName       string
```
grobie: With none of these fields being exported, how are users supposed to the this type?

cherti: good point indeed. Both the struct and the fields have to be exported, right?

SuperQ:
Very cool, I don't see any Spare or Failed disks in the test fixture, maybe worth adding them.

See examle 6: https://raid.wiki.kernel.org/index.php/Mdstat

mdstat.go
grobie: At least the fields, but I'd export both. Also please add comments to each line (as `golint` would tell you anyways), so that users can understand each field by just looking at godoc.

SuperQ:
A spare example:
```
md0 : active raid1 sdc1[2](S) sdb1[1] sda1[0]
      9998264 blocks super 1.2 [2/2] [UU]
```

SuperQ:
Thinking about this some more, maybe parsing `/proc/mdstat` is the wrong way to go.  There are some more detailed things we could export from `/sys/block/md.*/...`.  See the kernel docs[0]

[0]: https://www.kernel.org/doc/Documentation/md.txt

mdstat.go
cherti: done

cherti:
@SuperQ 
Might be a good idea. I think one could still do the transition later, though.

Any more suggestions to the code so far?

grobie:
Please run and fix both `go fmt` and `golint`.

mdstat.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err != nil {
+		return 0, fmt.Errorf("%s in buildline: %s", err, buildline)
+	}
+
+	return syncedSize, nil
+}
+
+// ParseMdstat parses an mdstat-file and returns a struct with the relevant infos.
+func (fs FS) ParseMdstat() (mdstates []MDStatus, err error) {
```
grobie: My preference, and hence the style in the repository, is to order code in a file by `constants, global vars, type, exported methods, unimported methods`. That helps me at least to quickly navigate to the important code without having to scroll past a lot of helper methods first. Therefore I'd appreciate if you could move this method up below the type

mdstat.go
```diff
@@ -0,0 +1,164 @@
...
+			// We aren't interested in lines with general info.
+			continue
+		}
+
+		mainLine := strings.Split(l, " ")
+		if len(mainLine) < 3 {
+			return mdStates, fmt.Errorf("error parsing mdline: %s", l)
+		}
+		currentMD = mainLine[0]               // name of md-device
+		isActive := (mainLine[2] == "active") // activity status of said md-device
```
grobie: My intention with the procfs library was to make the information available with go. I would not change / format the information too much, but try to return the information as read. So, what do you think about returning the status as string without doing any boolean comparision here? This would also make the library future proof in case the status types can be different than `active` or `inactive`.

mdstat.go
```diff
@@ -0,0 +1,164 @@
...
+	if len(matches) < 1+1 {
+		return 0, fmt.Errorf("too few matches found in buildline: %s", buildline)
+	}
+
+	if len(matches) > 1+1 {
+		return 0, fmt.Errorf("too many matches found in buildline: %s", buildline)
+	}
+
+	syncedSize, err := strconv.ParseInt(matches[1], 10, 64)
+
```
grobie: Please no newlines between a method invocation and the `if err != nil` block. These belong together.

mdstat.go
```diff
@@ -0,0 +1,164 @@
...
+		// If device is syncing at the moment, get the number of currently synced bytes,
+		// otherwise that number equals the size of the device.
+		if strings.Contains(lines[j], "recovery") || strings.Contains(lines[j], "resync") {
+			syncedBlocks, err = evalBuildline(lines[j])
+			if err != nil {
+				return mdStates, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+			}
+		} else {
+			syncedBlocks = size
```
grobie: More idiomatic go would look like:

```go
syncedBlocks = size
if strings.Contains(lines[j], "recovery") || ... {
  syncedBlocks, err = evalBuildline(lines[j])
  if err != nil {
    ...
  }
}
```

mdstat.go
cherti: yup, that sounds like a good idea

mdstat.go
cherti: done

mdstat.go
cherti: done

mdstat.go
```diff
@@ -0,0 +1,161 @@
...
+
+var (
+	statuslineRE = regexp.MustCompile(`(\d+) blocks .*\[(\d+)/(\d+)\] \[[U_]+\]`)
+	buildlineRE  = regexp.MustCompile(`\((\d+)/\d+\)`)
+)
+
+// MDStatus holds info parsed from /proc/mdstat.
+type MDStatus struct {
+	// Name of the device.
+	MDName string
```
grobie: What about just `Name`, avoids having a duplicated `MDStatus.MDName`?

mdstat.go
```diff
@@ -0,0 +1,161 @@
...
+	// activity-state of the device.
+	ActivityState string
```
grobie: I'd usually vote for shorter names, but if you think `State` would be ambiguous (are there any other md related states?) it's fine to leave it.

mdstat.go
```diff
@@ -0,0 +1,161 @@
...
+			// Skip entirely empty lines.
+			continue
+		}
+
+		if l[0] == ' ' {
+			// Those lines are not the beginning of a md-section.
+			continue
+		}
+
+		if strings.HasPrefix(l, "Personalities") || strings.HasPrefix(l, "unused") {
```
grobie: It should be safer to use a whitelist than a blacklist. It would also simplify the code:

```go
for i, l := range lines {
  if !strings.HasPrefix(l, "md") {
    continue
  }
  // ...
}
```

mdstat.go
```diff
@@ -0,0 +1,161 @@
...
+			return mdStates, fmt.Errorf("error parsing %s: entry for %s has fewer lines than expected", mdStatusFilePath, currentMD)
+		}
+
+		active, total, size, err := evalStatusline(lines[i+1]) // parse statusline, always present
+		if err != nil {
+			return mdStates, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+		}
+
+		// Now get the number of synced blocks.
+		var syncedBlocks int64
```
grobie: The above line should be moved down and merged with the initialization.

```go
var syncedBlocks int64 = size
if strings .... {
}
```

mdstat_test.go
```diff
@@ -0,0 +1,34 @@
+package procfs
+
+import (
+	"testing"
+)
+
+func TestMdstat(t *testing.T) {
+	fs := FS("fixtures")
+	mdStates, err := fs.ParseMdstat()
+
```
grobie: No newline please.

mdstat.go
```diff
@@ -0,0 +1,161 @@
...
+	// Total number of disks the device consists of.
+	DisksTotal int64
+	// Number of blocks the device holds.
+	BlocksTotal int64
+	// Number of blocks on the device that are in sync.
+	BlocksSynced int64
+}
+
+// ParseMdstat parses an mdstat-file and returns a struct with the relevant infos.
+func (fs FS) ParseMdstat() (mdstates []MDStatus, err error) {
```
grobie: `ParseMDStat` would be more in line with the rest of the naming.

grobie:
:+1: after having a look at the last comments. When you're done, please squash all your commits together so that there is a single commit including all code and tests for this change.

mdstat.go
cherti: I thought about that, but the file parsed is actually named /proc/mdstat, so the correct camelcase here would be parseMdstat, wouldn't it? But that's not too much of an issue of course.

mdstat.go
cherti: Is the name "mdX" mandatory for mdadm-devices? this would be required for this. I'm not sure, I'll see what I can find out.

cherti: @grobie 
the mds are special block devices like sda and so on. The latter can be renamed by udev to something like /dev/backup (whether that makes sense is a different question), therefore I'd suspect that the mds can also be renamed that way, therefore I am not sure if searching for "md" is a good idea.

grobie: Alright, makes sense then. Thanks for checking.

mdstat.go
juliusv: I'd still prefer the consistency though over keeping it exactly word-boundaried like the original name. So `ParseMDStat`, but I'd also vote for renaming the `MDStatus` struct to `MDStat`.

Hmm, the other types in this package are all prefixed with `Proc...` at the moment (`ProcStat`, `ProcLimit`, ...), but that's actually an anti-pattern because the name of the package is already `proc`. So maybe don't do that here.

grobie: The reason some - not all - types have a Proc prefix is that there is a difference between `/proc/stat` and `/proc/$pid/stat`. Some types only apply to processes and need to be distinguised. The name of the package is also `procfs` and not `proc` so I don't follow your argumentation of an anti-pattern here. Interested to hear other naming suggestions though.

As `mdstat` is not bound to specific processes and generally available under `/prod/mdstat` it should not have a `Proc` Prefix.

I agree that `MDStat` seems to match the procfs naming better than `MDStatus`.

grobie: Actually, all other methods use the `New<TYPE>` naming, so it might be best to follow that there and name the method `NewMDStat`? What do you think?

grobie: On the other hand, this method returns an array of `MDStat` types, so `New...` might be confusing. I'm fine with `Parse...`.

cherti:
@grobie 
everything should be taken care of now. If you agree, we should be ready for merge.

mdstat.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func evalStatusline(statusline string) (active, total, size int64, err error) {
+	matches := statuslineRE.FindStringSubmatch(statusline)
+
+	// +1 to make it more obvious that the whole string containing the info is also returned as matches[0].
+	if len(matches) < 3+1 {
+		return 0, 0, 0, fmt.Errorf("too few matches found in statusline: %s", statusline)
+	} else {
+		if len(matches) > 3+1 {
```
grobie: One last thing (as found via `golint *.go`): `mdstat.go:117:9: if block ends with a return statement, so drop this else and outdent its block`

Go prefers as less nesting as possible, so it should become:

```go
if len(matches) < 3+1 {
	return 0, 0, 0, fmt.Errorf("too few matches found in statusline: %s", statusline)
}
if len(matches) > 3+1 {
	return 0, 0, 0, fmt.Errorf("too many matches found in statusline: %s", statusline)
}
```

grobie: I guess just one check `len(matches) != 3+1` should be fine as well, but I leave that to you.

juliusv:
:+1: besides @grobie's last comments.

mdstat.go
cherti: That's right. I initially wanted to keep the error more informative in case there is indeed an error. But as I include the statusline now, that might actually not be necessary, good point.

mdstat.go
```diff
@@ -0,0 +1,160 @@
...
+			return mdStates, fmt.Errorf("error parsing %s: entry for %s has fewer lines than expected", mdStatusFilePath, currentMD)
+		}
+
+		active, total, size, err := evalStatusline(lines[i+1]) // parse statusline, always present
+		if err != nil {
+			return mdStates, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+		}
+
+		// Now get the number of synced blocks.
+		var syncedBlocks int64 = size
```
grobie: The other golint error can be fixed as well. `size` has already the type `int64`, so this statement can be shortened to `syncedBlocks := size`. I'd expect this statement also right above the if statement `if strings.Contains(lines[j], "recovery") || strings.Contains(lines[j], "resync")`.

cherti:
@grobie 
Should all be taken care of now. Just waiting for the checks to finish.

cherti:
@grobie 
ok, should be ready for merge now.

grobie:
Thank you for sticking with me :)

PR #15, opened 2015-10-23
moved FS methods to fs.go

guidopatanella:
@grobie @armen Minor. Moving FS related methods which were located on proc.go instead of fs.go

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -38,3 +38,48 @@ func (fs FS) open(p string) (*os.File, error) {
 func (fs FS) readlink(p string) (string, error) {
 	return os.Readlink(path.Join(string(fs), p))
 }
+
+// Self returns a process for the current process.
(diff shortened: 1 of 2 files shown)


grobie:
Hm, not sure about this change. I prefer the current way. FS is a global type, most functions depend on it. I find it more logical to have all functions related to one subsystem of procfs in one file. You also only moved the functions in proc.go, but there are FS related functions in every file.

As a style preference, please keep exported functions above unexported functions (in a file, I prefer the order: global constants, global vars, main functions (exported), helper functions (not exported).

grobie:
Thanks a lot for your contribution though!

guidopatanella:
Makes sense. Sorry for my newbie Go contribution. I was thinking around my former Erlang work. I am working on some tooling that might leverage from these packages. Looking forward to make more solid contribs soon... Thanks.

PR #16, opened 2015-10-23
moved FS methods to fs.go

guidopatanella:
@grobie @armen
moving FS related methods that were on proc.go to fs.go


grobie:
Please don't open a new pull request but edit the existing one.

PR #18, opened 2015-12-11
/proc/meminfo

abhiyerra:


fixtures/meminfo
```diff
@@ -0,0 +1,42 @@
...
+VmallocChunk:   34359637840 kB
+HardwareCorrupted:     0 kB
+AnonHugePages:     12288 kB
+HugePages_Total:       0
+HugePages_Free:        0
+HugePages_Rsvd:        0
+HugePages_Surp:        0
+Hugepagesize:       2048 kB
+DirectMap4k:       91136 kB
+DirectMap2M:    16039936 kB
```
brian-brazil: The meminfo on my machine has a newline at the end

meminfo.go
```diff
@@ -0,0 +1,174 @@
...
+				reflect.ValueOf(&m).Elem().Field(i).SetInt(v)
+			}
+		}
+	}
+
+	return m, nil
+}
+
+func (m Meminfo) Regex() string {
+	return "([A-Za-z()]*): *([0-9]*).*$"
```
brian-brazil: This regex doesn't match all of my meminfo, in particular HugePages_Total.

meminfo.go
```diff
@@ -0,0 +1,174 @@
...
+				}
+				reflect.ValueOf(&m).Elem().Field(i).SetInt(v)
+			}
+		}
+	}
+
+	return m, nil
+}
+
+func (m Meminfo) Regex() string {
```
brian-brazil: Should this be private?

brian-brazil:
The output of meminfo has changed a fair bit over the years (contrast your fixture with https://www.centos.org/docs/5/html/5.1/Deployment_Guide/s2-proc-meminfo.html). How do you ploan for a user to distinguish between missing fields and 0, and in addition how will new fields be handled?

meminfo.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	defer f.Close()
+
+	st := reflect.TypeOf(m)
+
+	s := bufio.NewScanner(f)
+	for s.Scan() {
+		line := s.Text()
+
+		re := regexp.MustCompile(m.Regex())
```
grobie: It'd be better to not compile a new regex during every invocation. I'd move that into a global variable:

```
var memInfoRE = regexp.MustCompile("....")
```

meminfo.go
```diff
@@ -0,0 +1,174 @@
...
+		submatch := re.FindAllStringSubmatch(line, 1)
```
grobie: Is it actually necessary to use a regular expression here at all? Looks like simple string operations like split would be sufficient already.

meminfo.go
jimmidyson: If we can do this without regexes all the better - regexes are comparably slow & should be avoided if possible. `string.Fields` seems better for this kind of thing.

fixtures/meminfo
abhiyerra: Hmm. don't see that on mine.

meminfo.go
abhiyerra: Made a fix.

meminfo.go
abhiyerra: Yes, changed it.

abhiyerra:
@brian-brazil I feel if we want to make sure that the values exist we should use a map instead of setting the fields themselves. That would actually make this code a lot simpler since it'd just be a `map[string]int64` and doing a get on the map would let us know if the value is set.

mdlayher:
Hi, appears this has grown out of date.  Mind rebasing?

mdlayher:
Hello, it appears these PRs have grown very stale.  Closing for now due to lack of activity, but feel free to re-open if you're able to pick them back up.

PR #20, opened 2015-12-15
/proc/net/tcp

abhiyerra:


juliusv:
@abhiyerra Thanks! I just noticed that nobody looked at this PR yet.

@brian-brazil you were always against reading /proc/net/tcp from the node exporter because of O(n^2) behavior, right?

grobie:
It's on my review list.

procfs is a package independent of node_exporter. Any procfs support is welcome.

juliusv:
Makes sense. Thank you @grobie :)

brian-brazil:
It's already implemented in the node exporter, but disabled by default. It's not in general safe to use this file due to poor performance when you have thousands of connections.

mdlayher:
Hello, it appears these PRs have grown very stale.  Closing for now due to lack of activity, but feel free to re-open if you're able to pick them back up.

hubt:
Are there alternatives here? I'd like to see this(or something similar) in the package.

I tested reading /proc/net/tcp and it was pretty fast. sub 200ms to read the file on a host with 16k open connections. That seems good enough for me.

I can start a new PR based on this one, or if there is a different way of doing this, let me know and I can take a shot at it.

SuperQ:
200ms is quite slow. But that's not really a big deal, we can still provide parsing support for this data in the library. There are other use cases for parsing this file and we ant this procfs library to be useful outside of just node_exporter.

@hubt If you want to open a new PR, please go ahead.

Issue #23, opened 2016-04-11
harcoding pagesize in proc_stat_test.go causes build failure on ppc64


ngothan:
there's hardcoding of pagesize 4096 `in proc_stat_test.go` which causes build failure on `ppc64`.
The pagesize is `64k` on `ppc64`. It should use the `os.Getpagesize()` to get the correct pagesize instead hardcoding.

```
diff -up procfs-406e5b7bfd8201a36e2bb5f7bdae0b03380c2ce8/proc_stat_test.go.than procfs-406e5b7bfd8201a36e2bb5f7bdae0b03380c2ce8/proc_stat_test.go
--- procfs-406e5b7bfd8201a36e2bb5f7bdae0b03380c2ce8/proc_stat_test.go.than      2016-04-11 06:37:23.590234409 -0400
+++ procfs-406e5b7bfd8201a36e2bb5f7bdae0b03380c2ce8/proc_stat_test.go   2016-04-11 07:33:59.835360813 -0400
@@ -1,5 +1,6 @@
 package procfs
 
+import "os"
 import "testing"
 
 func TestProcStat(t *testing.T) {
@@ -66,7 +67,7 @@ func TestProcStatResidentMemory(t *testi
                t.Fatal(err)
        }
 
-       if want, have := 1981*4096, s.ResidentMemory(); want != have {
+       if want, have := 1981*os.Getpagesize(), s.ResidentMemory(); want != have {
                t.Errorf("want resident memory %d, have %d", want, have)
        }
 }
```

juliusv:
@ngothan Thanks, makes sense! Do you want to send a PR with the patch above?

ngothan:
Julius, feel free to commit my patch above into github

juliusv:
@ngothan Thanks, done: https://github.com/prometheus/procfs/pull/24

PR #24, opened 2016-04-11
Get correct pagesize instead of hardcoding

juliusv:
Fixes https://github.com/prometheus/procfs/issues/23

### DIFF
--- a/proc_stat_test.go
+++ b/proc_stat_test.go
@@ -1,6 +1,9 @@
 package procfs
 
-import "testing"
+import (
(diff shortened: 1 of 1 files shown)


juliusv:
@grobie

grobie:
Thanks! I think we should first fix the CI tests. As I wrote in another PR, it's probably easiest to just remove the old go versions from the travis config.

grobie:
Ok, it's just the linter it seems.

PR #30, opened 2016-11-29
Enable parsing of '/proc/[pid]/mountstats'

mdlayher:
Tested on my system using NFS mounts from my desktop to server.  Reduced some of the per-operation stats in the fixture to save time.  Kernel is 3.13.0-37.

### DIFF
--- a/mountstats.go
+++ b/mountstats.go
@@ -1,0 +1,552 @@
+package procfs
+
+// While implementing parsing of /proc/[pid]/mountstats, this blog was used
+// heavily as a reference:
+//   https://utcc.utoronto.ca/~cks/space/blog/linux/NFSMountstatsIndex
+//
+// Special thanks to Chris Siebenmann for all of his posts explaining the
+// various statistics available for NFS.
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Constants shared between multiple functions.
+const (
+	deviceEntryLen = 8
+
+	fieldBytesLen  = 8
+	fieldEventsLen = 27
+
+	statVersion10 = "1.0"
+	statVersion11 = "1.1"
+
+	fieldTransport10Len = 10
+	fieldTransport11Len = 13
+)
+
+// A Mount is a device mount parsed from /proc/[pid]/mountstats.
+type Mount struct {
+	// Name of the device.
+	Device string
+	// The mount point of the device.
+	Mount string
+	// The filesystem type used by the device.
+	Type string
+	// If available additional statistics related to this Mount.
+	// Use a type assertion to determine if additional statistics are available.
+	Stats MountStats
+}
+
+// A MountStats is a type which contains detailed statistics for a specific
+// type of Mount.
+type MountStats interface {
+	mountStats()
+}
+
+// A MountStatsNFS is a MountStats implementation for NFSv3 and v4 mounts.
+type MountStatsNFS struct {
+	// The version of statistics provided.
+	StatVersion string
+	// The age of the NFS mount.
+	Age time.Duration
+	// Statistics related to byte counters for various operations.
+	Bytes NFSBytesStats
+	// Statistics related to various NFS event occurrences.
+	Events NFSEventsStats
+	// Statistics broken down by filesystem operation.
+	Operations []NFSOperationStats
+	// Statistics about the NFS RPC transport.
+	Transport NFSTransportStats
+}
+
+// mountStats implements MountStats.
+func (m MountStatsNFS) mountStats() {}
+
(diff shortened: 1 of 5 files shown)


mdlayher:
Is it okay to make use of subtests now that Go 1.7 is out?  Regarding the CI failure.  Could update CI to Go 1.7.

beorn7:
About the Go 1.7 features: This is a library, so we can expect it to be used in certain contexts where older Go compilers are used. Thus, I'd shy away from using Go-1.7 features if I can avoid it.
(Travis should _also_ test with Go 1.7, but we should keep 1.6 (and probably 1.5) for a while.)

@SuperQ @brian-brazil Is this PR doing what you discussed in https://github.com/prometheus/node_exporter/issues/108 ?

@grobie as you are the maintainer of this repo, could you signal if you are able/unable to review this?

Thanks everybody.

SuperQ:
@beorn7 Yes, it looks OK to me for https://github.com/prometheus/node_exporter/issues/108

mdlayher:
@beorn7 Updated to remove subtests.  It appears there is a failure unrelated to my change in CI for Go 1.5.

grobie:
I'll have some time the next 3 days.

mdlayher:
Hey @grobie, any chance you can look this over soon?

mountstats.go
```diff
@@ -0,0 +1,552 @@
...
+	fieldEventsLen = 27
+
+	statVersion10 = "1.0"
+	statVersion11 = "1.1"
+
+	fieldTransport10Len = 10
+	fieldTransport11Len = 13
+)
+
+// A Mount is a device mount parsed from /proc/[pid]mountstats.
```
grobie: TIL `(with optional leading article)` is a thing now.

grobie: Nit: Missing slash after `[pid]`.

proc.go
```diff
@@ -192,6 +192,18 @@ func (p Proc) FileDescriptorsLen() (int, error) {
 	return len(fds), nil
 }
 
+// MountStats retrieves statistics and configuration for mount points in a
+// process's name space.
```
grobie: Nit: Isn't namespace one word?

mountstats.go
```diff
@@ -0,0 +1,552 @@
...
+		}
+
+		m, err := parseMount(ss)
+		if err != nil {
+			return nil, err
+		}
+
+		// Does this mount also possess statistics information?
+		if len(ss) > deviceEntryLen {
+			// Only NFSv3 and v4 are supported for parsing statistics
```
grobie: Do you know other possible types providing statistics? I wonder if we should just leave Stats empty in this case, but being explicit is probably the right decision.

mountstats.go
```diff
@@ -0,0 +1,552 @@
...
+// parseMount parses an entry in /proc/[pid]/mountstats in the format:
+//   device [device] mounted on [mount] with fstype [type]
+func parseMount(ss []string) (*Mount, error) {
+	if len(ss) < deviceEntryLen {
+		return nil, fmt.Errorf("invalid device entry: %v", ss)
+	}
+
+	// Check for specific words appearing at specific indices to ensure
+	// the format is consistent with what we expect
+	format := []struct {
```
grobie: What do you think, would it make sense to extract this struct to a global variable in order to not rebuild it on function invocation? Or is the compiler already smart enough to do that for us?

Have you thought about using `fmt.Scanf` here?

mountstats.go
mdlayher: The manpage mentions:

```
              (4)  Optional  statistics  and configuration information.  Currently (as at Linux 2.6.26), only NFS filesystems export information
                   via this field.
```

So as far as I can tell, no other filesystem type exposes statistics here.

mountstats.go
mdlayher: I'll make some benchmarks for this logic and see which approach is fastest.

mdlayher:
Will rebase this once #31 is merged.

mountstats.go
mdlayher: https://gist.github.com/mdlayher/0e5db14dd6a2468bab343f2890d499d4

According to these results, moving this to a package variable results in about a 2 nanosecond savings per run.  `fmt.Sscanf()` is nowhere near competitive.

My gut tells me that the very slight speedup is not worth adding a variable used in a single function to the package namespace, but I'd be okay with changing it if requested.

grobie: Sounds good, thanks for checking.

grobie:
Thanks! Given the build error is fixed already in master, in going to merge it.

grobie:
> Pull requests that have a failing status can’t be merged on a phone.

Well, or not.

mdlayher:
@grobie all fixed!

grobie:
Thanks again!

PR #41, opened 2017-04-19
Add new package sysfs

mdlayher:
I just copied and pasted `procfs.FS` and created `sysfs.FS`.  This is where functions that rely on `/sys` can live, as discussed on:

https://groups.google.com/forum/#!topic/prometheus-developers/ys1LjSlKwGw

/cc @grobie @SuperQ, since you both preferred a separate package in the same repository.

### DIFF
--- a/sysfs/fs.go
+++ b/sysfs/fs.go
@@ -1,0 +1,46 @@
+// Copyright 2017 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sysfs
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// FS represents the pseudo-filesystem sys, which provides an interface to
+// kernel data structures.
+type FS string
+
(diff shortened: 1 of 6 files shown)


grobie:
LGTM, but I'm not sure it makes sense merging a skeleton package, as it will just confuse users. Let's wait until we have the first real functions.

mdlayher:
Conveniently, I wrote the code I needed first, but removed it to make sure the skeleton was fine!

Will add a second commit a bit later that accesses the XFS data I need in /sys.

mdlayher:
Updated.  @SuperQ @grobie please take a look at the second commit.

mdlayher:
Sorry, caught a bug on my own. :) Just modified the code to ensure that the file handle is closed regardless of parsing success or failure.

sysfs/fs.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Path returns the path of the given subsystem relative to the sys root.
+func (fs FS) Path(p ...string) string {
+	return filepath.Join(append([]string{string(fs)}, p...)...)
+}
+
+// XFSStats retrieves XFS filesystem runtime statistics for each mounted XFS
+// filesystem.  Only available on kernel 4.4+.  On older kernels, an empty
+// slice of *xfs.Stats will be returned.
+func (fs FS) XFSStats() ([]*xfs.Stats, error) {
```
grobie: So I always wondered, which pattern to follow here. `NewThing`, `ParseThing` or `Thing`. Do you have a good explanation why you chose the last? I guess I'd update the other functions for consistency then.

mdlayher: My personal rules of thumb for each case:

- `NewThing`: top level constructor functions. Never methods.
- `ParseThing`: parses some input data like `[]byte` or `io.Reader`.
- `Thing`: retrieval methods for some data.  Doesn't make much sense as a name for top-level functions to me.

This fits case number 3 in my rules.

Issue #40, opened 2017-04-19
Parse ipv6 addresses in ipvs backend status


filippog:
We are running `ipvs` collector from `node_exporter` on a machine receiving ipv6 traffic too, though `node_exporter` 0.13 fails with this:

```
time="2017-04-19T09:47:26Z" level=error msg="ERROR: ipvs collector failed after 0.000758s: could not get backend status: invalid IP: [2620" source="node_exporter.go:91"
```

And the `/proc/net/ip_vs` file looks like this (after redacting the addresses)

```
# cat /proc/net/ip_vs
IP Virtual Server version 1.2.1 (size=1048576)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port Forward Weight ActiveConn InActConn
TCP  00000000:0050 sh  
  -> 00000000:0050      Route   1      33         10        
  -> 00000000:0050      Route   1      48         52        
  -> 00000000:0050      Route   1      56         92        
  -> 00000000:0050      Route   1      48         15        
  -> 00000000:0050      Route   1      32         52        
  -> 00000000:0050      Route   1      62         28        
  -> 00000000:0050      Route   1      19         30        
  -> 00000000:0050      Route   1      93         43        
TCP  [2620:0000:0000:0000:0000:0000:0000:0000]:01BB sh 
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      0          0         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      2          0         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      2          0         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      0          0         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      2          0         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      1          3         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      1          1         
  -> [2620:0000:0000:0000:0000:0000:0000:0000]:01BB      Route   1      1          3         
TCP  00000000:01BB sh  
  -> 00000000:01BB      Route   1      35         59        
  -> 00000000:01BB      Route   1      74         104       
  -> 00000000:01BB      Route   1      275        340       
  -> 00000000:01BB      Route   1      25         56        
[...]

PR #42, opened 2017-04-20
Add new IPv6 format for ipvs

SuperQ:
Add additional parsing for kernels that use a different format for
IPv6 addresses in `/proc/net/ip_vs`.

CC @matthiasr


SuperQ:
Fixes https://github.com/prometheus/procfs/issues/40

fixtures/net/ip_vs
```diff
@@ -12,3 +12,7 @@ TCP  C0A80039:0CEA wlc
 TCP  C0A80037:0CEA wlc  
   -> C0A8321A:0CEA      Tunnel  0      0          0         
   -> C0A83120:0CEA      Tunnel  100    0          0         
+TCP  [2620:0000:0000:0000:0000:0000:0000:0001]:0050 sh
```
grobie: Hm, mixing different formats in the same file is confusing, as it wouldn't happen in real-life. I'd prefer different fixture files instead. The current fixtures structure needs some clean up though, so I'd be fine postponing that.

ipvs.go
```diff
@@ -199,23 +199,37 @@ func parseIPVSBackendStatus(file io.Reader) ([]IPVSBackendStatus, error) {
...
+func parseIPPort(ipPort string) (net.IP, uint16, error) {
+	var (
+		portString string
+		ip         net.IP
+		err        error
+	)
 
-	if len(tmp[0]) != 8 && len(tmp[0]) != 32 {
-		return nil, 0, fmt.Errorf("invalid IP: %s", tmp[0])
+	if len(ipPort) == 13 || len(ipPort) == 37 {
```
grobie: A switch statement is more common to handle multiple cases.

ipvs.go
```diff
@@ -199,23 +199,37 @@ func parseIPVSBackendStatus(file io.Reader) ([]IPVSBackendStatus, error) {
 	return status, nil
 }
 
-func parseIPPort(s string) (net.IP, uint16, error) {
-	tmp := strings.SplitN(s, ":", 2)
-
-	if len(tmp) != 2 {
-		return nil, 0, fmt.Errorf("invalid IP:Port: %s", s)
-	}
+func parseIPPort(ipPort string) (net.IP, uint16, error) {
```
grobie: Given the function name is clear, it's more idiomatic to use `s` as variable name.

ipvs.go
```diff
@@ -199,23 +199,37 @@ func parseIPVSBackendStatus(file io.Reader) ([]IPVSBackendStatus, error) {
...
+		}
+	} else {
+		return nil, 0, fmt.Errorf("invalid IP:Port: %s", ipPort)
 	}
 
-	ip, err := hex.DecodeString(tmp[0])
-	if err != nil {
-		return nil, 0, err
+	if len(portString) != 4 {
+		return nil, 0, fmt.Errorf("invalid port: %s", portString)
```
grobie: I don't like the usage of "invalid" here. More likely than being an invalid port, we'll be presented a different, unexpected format. So I'd change the wording in most of the error messages to `unexpected`.

ipvs.go
```diff
@@ -199,23 +199,37 @@ func parseIPVSBackendStatus(file io.Reader) ([]IPVSBackendStatus, error) {
...
+			return nil, 0, err
+		}
+	} else if len(ipPort) == 46 && strings.HasPrefix(ipPort, "[") {
+		portString = ipPort[42:]
+		ip = net.ParseIP(ipPort[1:40])
+		if ip == nil {
+			return nil, 0, fmt.Errorf("invalid IPv6 address: %s", ipPort[1:40])
+		}
+	} else {
+		return nil, 0, fmt.Errorf("invalid IP:Port: %s", ipPort)
```
grobie: `unexpected IP:port format: %s`.

ipvs.go
```diff
@@ -199,23 +199,37 @@ func parseIPVSBackendStatus(file io.Reader) ([]IPVSBackendStatus, error) {
...
+	var (
+		portString string
+		ip         net.IP
+		err        error
+	)
 
-	if len(tmp[0]) != 8 && len(tmp[0]) != 32 {
-		return nil, 0, fmt.Errorf("invalid IP: %s", tmp[0])
+	if len(ipPort) == 13 || len(ipPort) == 37 {
+		tmp := strings.SplitN(ipPort, ":", 2)
```
grobie: This can be moved outside of the case statement. There is no need to mix IP and port handling inside of this case/ifelse block.

fixtures/net/ip_vs
SuperQ: According to the issue, this is happening in real life.

ipvs.go
SuperQ: Ok

ipvs.go
SuperQ: Sure.

ipvs.go
SuperQ: I'm not sure this is true, all of the validation for the `IP:Port` handling is part of this switch block.

SuperQ:
@grobie Ok, let me know how it looks now.

ipvs.go
```diff
@@ -200,22 +200,37 @@ func parseIPVSBackendStatus(file io.Reader) ([]IPVSBackendStatus, error) {
...
+	var (
+		portString string
+		ip         net.IP
+		err        error
+	)
 
-	if len(tmp[0]) != 8 && len(tmp[0]) != 32 {
-		return nil, 0, fmt.Errorf("invalid IP: %s", tmp[0])
+	switch len(s) {
+	case 13, 37:
```
matthiasr: I think the 37 case may not actually exist. Let's keep it out until we have proof that it exists.

fixtures/net/ip_vs
matthiasr: Yeah, this is mixing IPv4 and IPv6 which is plausible.

grobie: I thought addresses will either be hexadecimal encoded or not. Well thenot.

ipvs.go
SuperQ: Ahh, ok, I'll update to remove the hex-encoded ipv6 case.  Thanks for the history. :-)

ipvs.go
grobie: Yeah, exactly. I'd find it clearer to only do one thing at a time.

```go
parts := strings.SplitN(s, ":", 2)
if len(parts) != 2 {
  return nil, 0, fmt.Errorf("unexpected IP:Port format: %s", s)
}

port, err := strconv.ParseUint(parts[1], 16, 16)
// ...

switch len(parts[0]) {
// ip handling
}
```

Or first ip and then port handling if you prefer that.

grobie:
Added a last comment @SuperQ (github collapses it unfortunately). Continuing to do first the ip+port split and then handling port and ip separately will keep the switch statement simpler and make the len(ip) checks clearer (as there won't longer uneven numbers including colons and the port).

fixtures/net/ip_vs
matthiasr: Looks like IPv4 is straight-up hexadecimal but IPv6 is represented with colons. That the original parsing _would_ have handled straight-hexadecimal IPv6 addresses was incidental should not be construed as these actually being part of the format.

SuperQ:
@grobie Ok, I cleaned up the IP vs Port handling as you requested.

I also removed the non-existent IPv6 hex case per @matthiasr.

PR #50, opened 2017-05-29
Parse and export all fields in /proc/stat

rtreffer:
This parses _all_fields in the /proc/stat file.


SuperQ:
Nice.  `SoftirqTotal` is technically redundant, but possibly useful depending on how the library is used.

Also, I think the number of columns in the softirq line has changed over time, I'm not sure parsing with a static scanf is such a great idea.  It may be better to do a string split and check the length.

rtreffer:
According to the linked document https://0xax.gitbooks.io/linux-insides/content/interrupts/interrupts-9.html it seems to be pretty stable.

Linux 3.0 (2011) used the same softirqs (well the BlockIOPoll name changed over time, but that's about it).

I wanted to just complete the proc/stat parsing and leave everything else out.

SuperQ:
We need to be compatible with Linux back to 2.6.23 since that's what Go supports.

SuperQ:
Here's a `/proc/stat` sample from CentOS 6:

```
cpu  547 0 702 3574 206 29 3 0 0
cpu0 547 0 702 3574 206 29 3 0 0
intr 20500 161 8 0 1 2617 0 1 1 1 0 0 0 110 0 0 0 0 0 0 0 0 0 0 0 0 0 0 3109 0 143 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
ctxt 24149
btime 1496049958
processes 1595
procs_running 1
procs_blocked 0
softirq 37046 0 14299 0 143 3091 0 257 0 2 19254
```

stat.go
```diff
@@ -24,33 +79,145 @@ func NewStat() (Stat, error) {
...
+			softIrqStats, total, err := parseSoftirqStat(line)
+			if err != nil {
+				return Stat{}, fmt.Errorf("couldn't parse %s (softirq): %s", line, err)
+			}
+			stat.SoftirqTotal = total
+			stat.Softirq = softIrqStats
+		case strings.HasPrefix(parts[0], "cpu"):
+			cpuStat, cpuID, err := parseCPUStat(line)
+			if err != nil {
+				return Stat{}, fmt.Errorf("couldn't parse %s (cpu): %s", line, err)
```
grobie: That's going to wrap the error again in the same string, or am I missing something? It seems we can just return `err` here.

stat.go
```diff
@@ -3,15 +3,70 @@ package procfs
...
+	Tasklet     uint64
+	Sched       uint64
+	Hrtimer     uint64
+	Rcu         uint64
+}
+
 // Stat represents kernel/system statistics.
 type Stat struct {
 	// Boot time in seconds since the Epoch.
 	BootTime int64
```
grobie: Maybe we should change this also to uint64? The library is marked as beta.

stat.go
```diff
@@ -3,15 +3,70 @@ package procfs
...
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"strconv"
 	"strings"
 )
 
+const (
+	userHz = 100
```
grobie: I guess we should add a comment how we came up with this constant and why we consider it safe to use a constant. There should be something in the node_exporter we could just use here.

stat.go
```diff
@@ -3,15 +3,70 @@ package procfs
...
 	// Boot time in seconds since the Epoch.
 	BootTime int64
+	// Summed up cpu statistics.
+	CPUTotal CPUStat
+	// Per-CPU statistics.
+	CPU []CPUStat
+	// Number of times interrupts were handled, which contains numbered and unnumbered IRQs.
+	IrqTotal uint64
+	// Number of times a numbered IRQ was triggered.
+	Irq []uint64
```
grobie: `IRQ`

stat.go
```diff
@@ -3,15 +3,70 @@ package procfs
...
 // Stat represents kernel/system statistics.
 type Stat struct {
 	// Boot time in seconds since the Epoch.
 	BootTime int64
+	// Summed up cpu statistics.
+	CPUTotal CPUStat
+	// Per-CPU statistics.
+	CPU []CPUStat
+	// Number of times interrupts were handled, which contains numbered and unnumbered IRQs.
+	IrqTotal uint64
```
grobie: `IRQTotal`

stat.go
rtreffer: The constant is defined in proc_stat as userHZ, I'll re-use that. It has a very good comment, too!

grobie: Ah, my memory was wrong. Yes, let's only define it once.

grobie:
Thanks!

PR #51, opened 2017-06-06
Initial commit for adding bcache

ideaship:
This changeset implements a bcache package in procfs as suggested by @SuperQ and @mdlayher in prometheus/node_exporter#588.


bcache/get.go
```diff
@@ -0,0 +1,310 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// parsePseudoFloat parses the peculiar format produced by bcache's bch_hprint
+func parsePseudoFloat(str string) (float64) {
```
mdlayher: Needs `go fmt`; the `(float64)` would become `float64` when formatted.

bcache/bcache.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Stats contains bcache runtime statistics, parsed from /sys/fs/bcache/.
+//
+// The names and meanings of each statistic were taken from bcache.txt and
+// files in drivers/md/bcache in the Linux kernel source. Counters are float64
+// (in-kernel counters are mostly unsigned long).
+type Stats struct {
+	// The name of the bcache used to source these statistics.
+	Name   string
+	Bcache bcacheStats
```
mdlayher: Unexported types referenced via exported fields in exported types are a litttle odd.

Apparently it works: https://play.golang.org/p/dIKaG6eQeQ

But I think `golint` complains, and there's no way to document the fields outside of reading the source code.  I'd just export every stats type in this package.

bcache/get.go
```diff
@@ -0,0 +1,310 @@
...
+	ss := strings.Split(str, ".")
+
+	intPart, _ := strconv.ParseFloat(ss[0], 64)
```
mdlayher: Annoying as it can be to propagate errors through your code, you'll want to check every string parsing error in this file.

sysfs/fixtures/devices/platform/applesmc.768/fan1_input
```diff
@@ -0,0 +1 @@
+0
```
mdlayher: Appears that some sensors fixtures made their way into this PR.  Can you remove these?

bcache/get.go
```diff
@@ -0,0 +1,310 @@
...
+		p.err = fmt.Errorf("failed to read: %s", path)
+		return 0
+	}
+	// Remove trailing newline
+	byt = byt[:len(byt)-1]
+	return dehumanize(byt)
+}
+
+// parsePriorityStats parses lines from the priority_stats file
+func parsePriorityStats(line string, ps *priorityStats) (err error) {
```
mdlayher: No need to use a named return value for a function that only returns an error.

bcache/get.go
```diff
@@ -0,0 +1,310 @@
...
+		p.err = fmt.Errorf("failed to parse: %s (%s)", path, err)
+		return res
+	}
+	return res
+}
+
+// GetStats collects from sysfs files data tied to one bcache ID.
+func GetStats(uuidPath string) (*Stats, error) {
+	var (
+		// stats = map[string]float64{}
```
mdlayher: Remove commented out code.

bcache/get.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	// bdev stats
+
+	reg := path.Join(uuidPath, "bdev[0-9]*")
+	bdevDirs, err := filepath.Glob(reg)
+	if err != nil {
+		return nil, err
+	}
+	bs.Bdevs = []bdevStats{}
```
mdlayher: Nit: since you already know the necessary capacity for this slice (`len(bdevDirs)`), initialize it with that capacity immediately.

bcache/get_test.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+func TestParseStats(t *testing.T) {
+	// test dehumanize
+	dehumanizeTests := []struct {
+		in  []byte
+		out float64
+	}{
+		{
+			in:  []byte("542k"),
```
mdlayher: You'll want to add some nonsensical test cases to check when errors occur as well.

bcache/get_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+	}
+	for _, tst := range dehumanizeTests {
+		got := dehumanize(tst.in)
+		if got != tst.out {
+			t.Errorf("dehumanize: %s, want %f, got %f", tst.in, tst.out, got)
+		}
+	}
+
+	// test parsePseudoFloat
```
mdlayher: It looks like this test should be broken into three different test functions.

ideaship:
Thank you for the review, @mdlayher! I pushed commits that should address all the points that you raised.

bcache/bcache.go
```diff
@@ -0,0 +1,84 @@
...
+	TreeDepth             float64
+	Internal              InternalStats
+	FiveMin               PeriodStats
+	Total                 PeriodStats
+}
+
+// BdevStats contains statistics for one backing device.
+type BdevStats struct {
+	Name      string
+	DirtyData float64
```
grobie: I assume possible values are either 0 or 1? We could think about using a bool, but fine with an int as well, as it makes type conversion more involved.

bcache/bcache.go
```diff
@@ -0,0 +1,84 @@
...
+	// The name of the bcache used to source these statistics.
+	Name   string
+	Bcache BcacheStats
+	Bdevs  []BdevStats
+	Caches []CacheStats
+}
+
+// BcacheStats contains statistics tied to a bcache ID.
+type BcacheStats struct {
+	AverageKeySize        float64
```
grobie: Will the source ever return floats, or only integers?

While Prometheus only supports float64 as data type, the procfs library is also used in other programs. I'd prefer to use the most natural data type (so probably integers for all?/most bcache fields) and do the type conversion to float64 in node_exporter.

ideaship: The source does return floats for any number > 1023. However, they are ints turned into "humanized" strings such as "1.0k". Yes, we could convert these numbers into int64 or uint64 before passing them to the caller. I went with float64 mostly for safety reasons; float64 will automatically do the correct thing where uint64 and int64 would fail. The Linux function that does the printing for most of the files contains prefixes up to Yotta which won't fit into a uint64, but it's unlikely to ever occur, especially considering that the input type for said function is int64_t. Also, I have not audited all return values for signed values, so some might conceivably return negative values (although none of them look like they should). I'll take a look at using uint64 instead of float64.

bcache/bcache.go
ideaship: DirtyData is the "Amount of dirty data for this backing device in the cache." It is not a bool.

ideaship:
@grobie: I pushed an update that uses uint64 instead of float64.

bcache/get.go
```diff
@@ -0,0 +1,329 @@
...
+	// by a small amount. Unexpected bumps in time lines may occur because
+	// for bch_hprint .1 != .10 and .10 > .9 (at least up to Linux
+	// v4.12-rc3).
+
+	// Restore the proper order:
+	fracPart = fracPart / 10.24
+	return intPart + fracPart, nil
+}
+
+// dehumanize converts human-readable byte slice into float64
```
SuperQ: s/float64/uint64/

SuperQ: Also, we have a comment style for complete sentences.  Everything should start caps, and end with a period.

PR #55, opened 2017-06-09
Remove fixture files with Windows incompatible names

### LINKED ISSUE #54: bcache folders breaks transtive imports to windows
#51 The folders in the repo cannot be created on a windows machine when checking out the head. 

` git clone https://github.com/prometheus/procfs I:\tools\go\path\src\github.com\prometheus\procfs
Cloning into 'I:\tools\go\path\src\github.com\prometheus\procfs'...
fatal: cannot create directory at 'sysfs/fixtures/devices/pci0000:00': Invalid argument
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'

package github.com/prometheus/procfs: exit status 128`

I do not use your package directly  it is a transitive dependency from go-kit

grobie:
Quick fix: Some fixture files for sysfs have names which are invalid on
Windows.  This breaks git checkout on Windows systems. As procfs is a
depdency of client_golang, this breaks Windows builds on hundreds of
projects.

A nicer solution could rename all paths with characters invalid on some
filesystems. Such a rename approach should ideally work with default
tools available on OSX and Linux.

---

Fixes #54.

@brian-brazil @SuperQ @ideaship 

As we currently broke Windows builds of any project not vendoring all dependencies and using procfs/client_colang/go-kit, I quickly put that together. Happy to change to a nicer solution later, but I won't be available for the next few days.


brian-brazil:
:+1:

PR #56, opened 2017-06-14
Replace fixtures.tar.gz with regular files

ideaship:
In order to keep the fixtures in a form that can easily be reviewed,
this changeset puts the files from fixtures.tar.gz back into the repo as
regular files (in a separate directory, fixtures.src). In order to
prevent filenames that prevent checkouts on Windows, colons are encoded
as `_@colon@_` and copied to the correct location by make.

Alternatives: files could be renamed rather than copied (using -depth
in the find command to rename files before their parent directories).
However, doing so would remove files from a location where git expects
them.


Makefile
```diff
@@ -8,11 +8,17 @@ lint:
 	go get github.com/golang/lint/golint
 	golint *.go
 
-test: sysfs/fixtures/.unpacked
+test: sysfs/fixtures/.copied
```
grobie: Instead of using a marker file, which can get easily out of sync, I'd like to use each @colon@ file as dependency here. @matthiasr we do something similar in a couple of SoundCloud projects, do you have some time to use your make skills to apply that here?

I'll try to port it this weekend otherwise.

matthiasr: Working on it. The straightforward way will require GNU make, is that acceptable?

grobie:
I'd say yes. Is it available on osx?

Makefile
matthiasr: Turns out, I can't. Make [does not handle filenames with characters well](http://lists.gnu.org/archive/html/bug-make/2004-10/msg00010.html), so specifying the target names breaks it completely. Not specifying the target names means we end up in the same out-of-sync situation with the added downside of redoing all the work all the time.

matthiasr:
> Is it available on osx?

Yes, it's the default there. It's not on other BSDs, but can be installed.

Makefile
grobie: Hmm, I guess then I'd vote to remove the marker and rename them every time. Otherwise developers can run in staleness issues when others make changes to the fixtures directory.

ideaship: I have a renaming routine quite similar to the copy code in this PR. Obviously, this means that after a make test, every renamed file will show up as missing for git. Also, giving up the marker makes the test process slower. We have to walk the file try every time to check for files that need renaming. And a file removed from git won't get the renamed version removed automatically. For that, we'd have to rebuild the directory from scratch every time (also doable if that is your preference).

grobie: The problem is, the marker needs to be somehow dependent on the files in source code. Otherwise the copy command will never be executed again. That's what I asked @matthiasr to help with, but apparently it's not possible due to Make's lack of colon support in filenames.

I'd prefer slower test execution over stale fixture files. I'd like to prevent developers from unnecessary debug sessions when they forgot to remove the marker file manually to trigger an update.

matthiasr:
I agree. Renaming is a non-starter, destroying the checkout all the time
will make development impossible. Let's stick with the approach in this PR
but use a phony target instead of the .copied file

grobie:
Thank you @ideaship!

ideaship:
Sigh @ #57. I need to find a Windows box for testing.

PR #58, opened 2017-06-20
Rebuild fixtures every time

ideaship:
With sysfs/fixtures/.unpacked as a build target, developers have to
remember to delete sysfs/fixtures whenever fixtures.tar.gz changes
(e.g., because they switched to a different branch).

With this patch, the fixtures directory is rebuilt every time a
test is run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,8 @@ lint:
 	go get github.com/golang/lint/golint
 	golint *.go
 
-test: sysfs/fixtures/.unpacked
+test:
+	cd sysfs && rm -rf fixtures && tar xzf fixtures.tar.gz
 	go test -v ./...
-
-sysfs/fixtures/.unpacked: sysfs/fixtures.tar.gz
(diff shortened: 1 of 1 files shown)


ideaship:
I think the same arguments that led to an equivalent change in the abandoned copying patch apply here as well.

grobie:
Here we're relying on the power of Make :-) The `sysfs/fixtures/.unpacked` target depends on `sysfs/fixtures.tar.gz`. When one of the dependencies of a make target is younger than the target itself, make will execute the target again.

Try it out yourself with `touch sysfs/fixtures.tar.gz`. Next time you run `make test` it should extract the fixtures again.

ideaship:
Ah, I should have thought of that. Thanks, @grobie! The one case where it still fails is removing files. Once a file is untared to fixtures, it will not get removed automatically -- it will be present for tests with any checkout. But that special case may be rare enough to ignore :).

PR #59, opened 2017-07-02
Add ttar: plain text archive, replacement for tar

ideaship:
This changeset adds ttar, a plain text replacement for tar, and uses it
for the sysfs fixture archive. The syntax is loosely based on tar(1).

Using a plain text archive makes it possible to review changes without
downloading and extracting the archive. Also, when working on the repo,
git diff and git log become useful again, allowing a committer to verify
and track changes over time.

The code is written in bash, because bash is available out of the box on
all major flavors of Linux and on macOS. The feature set used is
restricted to bash version 3.2 because that is what Apple is still
shipping.

The programm also works on Windows if bash is installed. Obviously, it
does not solve the Windows limitations (path length limited to 260
characters, no symbolic links) that prompted the move to an archive
format in the first place.


ideaship:
This changeset is inspired by @grobie's comment in issue #54.

We needed an archive format because Windows fails when checking out repos with certain files (e.g., file paths that include colons or are longer than 260 characters).

For now, we are using tar: a binary format that is opaque as far as github reviews and git (diff, log -p) are concerned.

@grobie mentioned https://github.com/jtvaughan/ptar which addresses this problem but adds a dependency to another language, C.

This changeset implements a simple plain-text archive in bash 3.2, which is installed by default with all major Linux distributions and with all versions of macOS.

The script can handle plain-text files, symbolic links, and directories.

SuperQ:
Yea, this is fine.  Not my normal bash style, but reasonably good.

PR #65, opened 2017-10-27
Add support for /proc/net/dev, /proc/[pid]/net/dev, and /proc/[pid]/ns

andrewkroh:
This PR adds support for reading from /proc/net/dev, /proc/[pid]/net/dev, and /proc/[pid]/ns.

`/proc/net/dev` and `/proc/[pid]/net/dev` provide network interface metrics. These are the values that ifconfig reads.

`/proc/[pid]/ns` is directory containing symlinks that describe what namespaces the process is a member of.


net_dev.go
```diff
@@ -0,0 +1,203 @@
...
+
+		line, err := nd.parseLine(s.Text())
+		if err != nil {
+			return nd, err
+		}
+
+		nd[line.Name] = *line
+	}
+
+	return nd, nil
```
grobie: You need to check for errors from the scanner. Replace nil with `s.Err()`.

andrewkroh: Good catch. I pushed an update.

PR #66, opened 2017-11-01
Update .travis.yml

andrewkroh:
Update Go versions and add `go_import_path` so that Travis can work on forks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 sudo: false
+
 language: go
+
 go:
-    - 1.7.6
-    - 1.8.3
+- 1.9.x
+- 1.x
+
+go_import_path: github.com/prometheus/procfs


.travis.yml
```diff
@@ -1,5 +1,7 @@
 sudo: false
 language: go
+go_import_path: github.com/prometheus/procfs
 go:
     - 1.7.6
```
mdlayher: I think it'd be reasonable to do 1.x instead of explicit versions here, but will ping others for feedback too.

SuperQ: We use stuff like this in other repos
```
go:
- 1.9.x
- 1.x
```

SuperQ:
Please follow the format example in the [prometheus](https://github.com/prometheus/prometheus/blob/master/.travis.yml) repo.

PR #69, opened 2018-01-03
Procfs module for Lio backstore node_exporter

AvengerMoJo:
This module mainly create lio run time status profile for node_exporter lio to consume. Some helper functions also support checking and reading status from different backstore for sys. 
3 number Read in Byte, Write in Byte and ISCSI OPS. 
4 backstore, fileio, iblock, rbd, rdmcp.


AvengerMoJo:
iscsi/iscsi.go:27:5: cannot find package "github.com/prometheus/common/log" in any of

I can't use the common module? Should I take out all the debug? Seem we don't have any debug in the rest of the project.

AvengerMoJo:
Thanks for the review, I am working on the update and ping you when I'm done.

AvengerMoJo:
Hi @mdlayher I did golint, staticcheck, and go fmt for all my commit, added the fs_test.go for the iscsi stat test case along with the updated fixtures.ttar. I believe I still need to add another iscsi_test.go in the iscsi plugin. I just want to check up with you that I'm on the right path. And I will commit the final update to the iscsi_test.go asap. Thanks

AvengerMoJo:
Hi @mdlayher I'm done with the test case now. Please take a look. Thanks

mdlayher:
Hi, I'm very busy with work in recent days. Will request a review from others.

AvengerMoJo:
@mdlayher Thanks alot for your reply. @grobie and @SuperQ if you need me to explain over email about my PR please let me know.

iscsi/iscsi.go
```diff
@@ -0,0 +1,372 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+// for default sysfs mount point
+var (
+	sysPath = "/sys"
```
SuperQ: This should not be declared here.  Please take a look at the XFS and bcache collectors.  All of the filenames and access is declared in `sysfs/fs.go`.

SuperQ:
Please rebase this against master to pick up updated testes.

iscsi/iscsi.go
AvengerMoJo: hi @SuperQ if you look at my update in sysfs/fs.go my ISCSIStats call is basically the same as BcacheStats. This default sysfs is only here for fall back in case someone using it without going from the sysfs/fs.go directly. There are two part of sysfs I need to look into for some iscsi backstore. e.g. ceph, more information need to be match, instead of reading it directly from sysfs's structure.

discordianfish:
LGTM, what is blocking this? @SuperQ / @mdlayher

SuperQ:
There seems to be some very un-clean commit rebasing/merging here.  There are changes to `nfs/nfs.go` that appear to be unrelated, for example.

iscsi/iscsi.go
SuperQ: This is generating a very un-clean interface.  Please follow the conventions of the other collectors.

grobie:
I'm also not really happy about the package structure and inclusion. I don't think we should force everyone using the procfs package to also download the iscsi package. This pattern has been introduced before, but I don't think we should continue coupling all packages together.

AvengerMoJo:
@SuperQ about the unrelated rebase is possible my rebase error. I should start over a new PR to clean up everything. But before I do that, I need to understand which collector you suggest me to follow, I am currently using Bcache as my sample to construct this. I am willing to rework this but I really didn't get any feedback. If any pointer you can let me know what I am not doing it right or the structure you don't want it. I will fix it. Since the node_exporter lio module is already being use in our environment, I need to make sure this could push stream properly. Thanks alot. 
@grobie are you talking about the sysfs/fs.go inclusion? I'm just following what xfs and bcache here... what's your suggestion? I am not aware other package doing it dynamically. Thanks

AvengerMoJo:
Merge to a new PR https://github.com/prometheus/procfs/pull/88

PR #70, opened 2018-01-12
Add NFSd stats

SuperQ:


nfs/nfsd.go
```diff
@@ -0,0 +1,174 @@
+// /proc/net/rpc/nfsd parsing documented by https://www.svennd.be/nfsd-stats-explained-procnetrpcnfsd/
+package nfs
+
+// rc line: Reply Cache
+type NFSdReplyCache struct {
```
mdlayher: Since this package is called `nfs` (maybe it should be `nfsd`), you should omit the prefix for these types: `ReplyCache`, `FileHandles`, etc.

Not sure if `golint` clean is enforced in procfs but I'd run it on this code and update the doc comments.

nfs/util/parse.go
```diff
@@ -0,0 +1,31 @@
+package util
```
mdlayher: In general I wouldn't advise a package named `util`, but it might be okay for an internal package.  I'd make this an internal package: `procfs/internal/util` or similar.

nfs/nfsd.go
SuperQ: Yea, I started it this way because I wanted to include `nfs` and `nfsd` stats in the same package.

mdlayher: If they're very related or share some of the same structures, that seems fine.  If not, I'd just make two separate packages and clarify with doc comments.

nfsd/parse_nfsd.go
```diff
@@ -0,0 +1,282 @@
...
+		// require at least <key> <value>
+		if len(parts) < 2 {
+			return nil, fmt.Errorf("invalid NFSd metric line %q", line)
+		}
+		label := parts[0]
+
+		var values []uint64
+		var err error
+		if label == "th" {
+			values, err = util.ParseUint64s(parts[1:3])
```
mdlayher: Probably worth adding another bounds check to avoid a panic.

nfsd/parse_test.go
```diff
@@ -0,0 +1,200 @@
...
+					Verify:       3,
+					Write:        3,
+					RelLockOwner: 0,
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		stats, err := nfsd.ParseRPCStats(strings.NewReader(tt.content))
```
mdlayher: I'd do subtests: `t.Run(tt.name, func(t *testing.T) { /* ... */ })`

nfsd/parse_nfsd.go
```diff
@@ -0,0 +1,285 @@
...
+		case "proc2":
+			stats.V2Stats, err = parseV2Stats(values)
+		case "proc3":
+			stats.V3Stats, err = parseV3Stats(values)
+		case "proc4":
+			stats.V4Stats, err = parseV4Stats(values)
+		case "proc4ops":
+			stats.V4Ops, err = parseV4Ops(values)
+		default:
+			return nil, fmt.Errorf("invalid NFSd metric line %q", metricLine)
```
grobie: I'd say `unknown` makes more sense than `invalid`.

nfsd/parse_test.go
grobie: @SuperQ The test error output currently doesn't show which of the test cases failed. What do you think about using subtests? You just need to wrap the whole block into the function @mdlayher pasted.

SuperQ: Yes, sorry, got busy with other stuff, great idea.

nfsd/parse_test.go
```diff
@@ -178,23 +178,25 @@ proc4ops 72 0 0 0 1098 2 0 0 0 0 8179 5896 0 0 0 0 5900 0 0 2 0 2 0 9609 0 2 150
...
-			t.Error("expected an error, but none occurred")
-			continue
-		}
-		if !tt.invalid && err != nil {
-			t.Errorf("unexpected error: %v", err)
-			continue
-		}
+			if tt.invalid && err == nil {
+				t.Error("expected an error, but none occurred")
+				return
```
grobie: Remove these and turn `t.Error` into `t.Fatal`.

SuperQ:
Thanks @grobie.  Updated with `Fatal` returns.

PR #71, opened 2018-01-23
Add 'States' function to return the state for each process of the system

sterchelen:
This PR is link to offer a metric with the number of process by state. 
So, this PR add a function named `States` which returns a map of state as string by `Proc`.
Adding unit test also.


grobie:
Such function doesn't really belong into this library. The purpose of this repository is to provide a simple Go interface to procfs and sysfs.

sterchelen:
See last comment by @discordianfish ==> https://github.com/prometheus/node_exporter/pull/633

grobie:
Specific use cases to transform the output from procfs into other data
structures don't belong into this library. Move this states function into
node_exporter instead.

PR #72, opened 2018-01-24
Refactor NFS parsing

SuperQ:
* Rename into `nfs` package.
* Add support for both server `nfsd` and client `nfs` parsing.


discordianfish:
LGTM

PR #74, opened 2018-02-09
Add support for NFS Client stats on older kernels

SuperQ:
Some older kernels have fewer NFS v4 client stats.  The kernel behavior
(so far) has been to append new stats to the existing list.

Also cleanup some NFSd -> NFS copy-pasta mistakes.


nfs/parse.go
```diff
@@ -178,8 +178,14 @@ func parseV3Stats(v []uint64) (V3Stats, error) {
 
 func parseClientV4Stats(v []uint64) (ClientV4Stats, error) {
 	values := int(v[0])
-	if len(v[1:]) != values || values < 59 {
-		return ClientV4Stats{}, fmt.Errorf("invalid V4Stats line %q", v)
+	if len(v[1:]) != values {
+		return ClientV4Stats{}, fmt.Errorf("invalid ClientV4Stats line %q", v)
+	}
+
+	if values < 59 {
```
mdlayher: I'd say this merits a comment.

nfs/parse_nfs_test.go
```diff
@@ -32,6 +32,131 @@ func TestNewNFSdClientRPCStats(t *testing.T) {
...
+				V2Stats: nfs.V2Stats{
+					Null:     16,
+					GetAttr:  57,
+					SetAttr:  74,
+					Root:     52,
+					Lookup:   71,
+					ReadLink: 73,
+					Read:     45,
+					WrCache:  86,
+					Write:    0,
```
mdlayher: Nit: you can omit all the zeros from all of these test structures if you want.

SuperQ: I'd rather keep them in as explicit.

Issue #37, opened 2017-02-14
Many files missing license header


mdlayher:
Just noticed this when @grobie asked me to put license headers on my XFS files in #36 .

We use a pretty rudimentary shell script check in go-qemu to verify each Go source file has a license.

https://github.com/digitalocean/go-qemu/blob/master/scripts/licensecheck.sh

 It would probably be a good idea to set up something similar in this repo, and other Prometheus repos.

grobie:
See https://github.com/prometheus/prometheus/blob/master/scripts/check_license.sh as well. I guess we should open the discussion how/where to manage such scripts.

PR #77, opened 2018-02-27
Add license check

SuperQ:
* Add licence check script.
* Add to Makefile.
* Add to CI.
* Fix license in several files.

Closes: https://github.com/prometheus/procfs/issues/37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ go:
 go_import_path: github.com/prometheus/procfs
 
 script:
-- make style vet test staticcheck
+- make style check_license vet test staticcheck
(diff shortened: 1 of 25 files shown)


grobie:
The year in the license header should reflect the year the file was added AFAIK.

SuperQ:
Hmm, I thought it was usually the last modified time.

SuperQ:
Some other opinions on the subject say it's for "When the notice is first applicable".  In this case, you might have the opinion that the notice is first applicable now, since it did not exist before.

grobie:
Sounds fair, we should stick to this definition and not update the year of
existing license headers.

Issue #75, opened 2018-02-26
Breaks docker-compose


Nighthawk22:
We are using the prometheus library at one of our projects and it seems like the fixtures directory is breaking the last docker-compose build:
```
> docker-compose up --build
Building offices
Traceback (most recent call last):
  File "docker-compose", line 6, in <module>
  File "compose/cli/main.py", line 71, in main
  File "compose/cli/main.py", line 124, in perform_command
  File "compose/cli/main.py", line 1001, in up
  File "compose/cli/main.py", line 997, in up
  File "compose/project.py", line 463, in up
  File "compose/service.py", line 310, in ensure_image_exists
  File "compose/service.py", line 989, in build
  File "site-packages/docker/api/build.py", line 150, in build
  File "site-packages/docker/utils/build.py", line 14, in tar
  File "site-packages/docker/utils/utils.py", line 103, in create_archive
IOError: Can not access file in context: /Users/someuser/projects/src/somerepo.net/repo/offices/vendor/github.com/prometheus/procfs/fixtures/26231/ns/mnt
Failed to execute script docker-compose
```
Building the service with docker itself is working prefectly fine.

Docker version:
```
Client:
 Version:	18.02.0-ce
 API version:	1.36
 Go version:	go1.9.3
 Git commit:	fc4de44
 Built:	Wed Feb  7 21:13:05 2018
 OS/Arch:	darwin/amd64
 Experimental:	true
 Orchestrator:	kubernetes

Server:
 Engine:
  Version:	18.02.0-ce
  API version:	1.36 (minimum version 1.12)
  Go version:	go1.9.3
  Git commit:	fc4de44
  Built:	Wed Feb  7 21:20:15 2018
  OS/Arch:	linux/amd64
  Experimental:	true
```

Docker-compose version:
```
docker-compose version 1.19.0, build 9e633ef
```

codingconcepts:
Same issue here :(

brian-brazil:
Have you filed a bug with docker-compose?

codingconcepts:
From @Nighthawk22's message, it appears as though this was previously working with docker-compose.

Unfortunately, none of yesterday's checkins appear to have obviously broken anything...

Nighthawk22:
Yes it was previously working. I will open a docker-compose ticket.

codingconcepts:
Sorry if I'm being dense here but are you going to raise a ticket here https://github.com/docker/compose?

[...]

Issue #60, opened 2017-08-02
Symlink fixtures/26231/exe breaks git and doesn't work on windows


veqryn:
I've checkout out this repository, and also use it as a dependency in some golang work I do.
Unfortunately, git always shows there is a difference in this file:
```bash
MinGW 06:57:54 ~/.glide/cache/src/https-github.com-prometheus-procfs$ git diff
diff --git a/fixtures/26231/exe b/fixtures/26231/exe
index a91bec4..690239f 120000
--- a/fixtures/26231/exe
+++ b/fixtures/26231/exe
@@ -1 +1 @@
-/usr/bin/vim
\ No newline at end of file
+C:/usr/bin/vim
\ No newline at end of file
```

Even I check it out multiple times:
```bash
MinGW 06:57:39 ~/.glide/cache/src/https-github.com-prometheus-procfs$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   fixtures/26231/exe

no changes added to commit (use "git add" and/or "git commit -a")
MinGW 06:57:42 ~/.glide/cache/src/https-github.com-prometheus-procfs$ git checkout fixtures/26231/exe
MinGW 06:57:50 ~/.glide/cache/src/https-github.com-prometheus-procfs$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   fixtures/26231/exe

no changes added to commit (use "git add" and/or "git commit -a")
```

It appears to be a symlink, that gets changed on checkout for some reason:
```bash
MinGW 06:58:31 ~/.glide/cache/src/https-github.com-prometheus-procfs/fixtures/26231$ ls -la
total 20
drwxr-xr-x 1 cduncan 1049089    0 Aug  1 18:57 ./
drwxr-xr-x 1 cduncan 1049089    0 Aug  1 18:57 ../
-rw-r--r-- 1 cduncan 1049089   16 Aug  1 18:41 cmdline
-rw-r--r-- 1 cduncan 1049089    4 Aug  1 18:41 comm
lrwxrwxrwx 1 cduncan 1049089   14 Aug  1 18:57 exe -> /c/usr/bin/vim
[...]

PR #79, opened 2018-02-28
Package fixtures directory using ttar

### LINKED ISSUE #78: Support binary files in ttar
The procfs file structure is not compatible with non-Linux systems. For example symlinks cause issues when checked out on Windows or Mac (see #60 and #75). For similar reasons, a `ttar` solution was contributed which allows packaging the whole fixtures structure into a single file, while still allowing meaningful diffs and code reviews.

In order to use `ttar` for the `fixtures/` directory, it needs to support encodeing+decoding null bytes so that files like `cmdline` can be packaged as well.

@ideaship @SuperQ

grobie:
@ideaship @brian-brazil @SuperQ 

Fixes #78 #75 #60


brian-brazil:
:+1:

ideaship:
@grobie @brian-brazil Wow, you guys are quick. The patch was merged before I could check it out. I am not enthusiastic about this patch: it turns two strings (at least one of which is rather common in all kinds of files) into magic keywords that must not be part of the input files or ttar will break. How much are you willing to bet that "EOF" will never, ever turn up in any file that you would like to put into fixtures?

A longer code string (e.g., "EOF-gaev9AeKiv") would IMO be much less problematic. Maybe we could even work out a solution with escaping which would cover all possible files. Would you be willing to consider an improved patch? I could at least look into it.

grobie:
@ideaship Thanks for still looking into this! I'd be more than happy to accept improved solutions to this issue.

I was considering the possible clashes on EOF, but given we control the content of the fixture files here, came to the conclusion that it's good enough for now. For a general purpose ttar implementation, I can see issues with that though. My main goal was to provide a solution the reported issues with symlinks and exe files in this repo.

I could imagine that we could avoid the EOF marker at all and mark missing line endings different somehow. One thing to keep in mind with the keywords is that the created ttar file should remain human readable, so we shouldn't go too crazy I guess.

Again, thanks a ton for your continued contributions! We use ttar in https://github.com/prometheus/node_exporter as well by now. You should consider creating a repo for it so you can become famous ;-)

grobie:
Oh, and after sending the last comment I checked your profile and you already did that! I created #81 to eventually vendor ttar from your repo which should become the canonical source.

ideaship:
Cool. I already imported your fix, but I could not get myself to push it because I hope I can come up with a better solution, now that I know there is a need. By the way, I just noticed that the "wc -l | cut" part appears to need tweaking to work on macOS.

grobie:
Ah, dang. I usually try to avoid awk if not necessary. Feel free to revert
that as it seems to be more portable.

ideaship:
I'll take a look. Give me a couple of days or so.

ideaship:
@grobie I posted an update to https://github.com/ideaship/ttar.

PR #80, opened 2018-02-28
Test procfs against previous go versions as well

grobie:
Procfs is a library and not a standalone binary. It should work against
older go versions as well.


brian-brazil:
client_golang does 1.7 too. The two should probably be the same.

@beorn7

grobie:
Updated.

brian-brazil:
:+1:

PR #82, opened 2018-03-03
Sys class net

klatys:
Related to https://github.com/prometheus/node_exporter/issues/300

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -16,41 +16,43 @@ package sysfs
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"reflect"
 	"strconv"
 	"strings"
+	"syscall"
 )
 
 // NetClassIface contains info from files in /sys/class/net/<iface>
 // for single interface (iface).
 type NetClassIface struct {
 	Name             string // Interface name
-	AddrAssignType   uint64 `fileName:"addr_assign_type"`   // /sys/class/net/<iface>/addr_assign_type
-	AddrLen          uint64 `fileName:"addr_len"`           // /sys/class/net/<iface>/addr_len
+	AddrAssignType   int64  `fileName:"addr_assign_type"`   // /sys/class/net/<iface>/addr_assign_type
+	AddrLen          int64  `fileName:"addr_len"`           // /sys/class/net/<iface>/addr_len
 	Address          string `fileName:"address"`            // /sys/class/net/<iface>/address
 	Broadcast        string `fileName:"broadcast"`          // /sys/class/net/<iface>/broadcast
-	Carrier          uint64 `fileName:"carrier"`            // /sys/class/net/<iface>/carrier
-	CarrierChanges   uint64 `fileName:"carrier_changes"`    // /sys/class/net/<iface>/carrier_changes
-	CarrierUpCount   uint64 `fileName:"carrier_up_count"`   // /sys/class/net/<iface>/carrier_up_count
-	CarrierDownCount uint64 `fileName:"carrier_down_count"` // /sys/class/net/<iface>/carrier_down_count
-	DevID            uint64 `fileName:"dev_id"`             // /sys/class/net/<iface>/dev_id
-	Dormant          uint64 `fileName:"dormant"`            // /sys/class/net/<iface>/dormant
+	Carrier          int64  `fileName:"carrier"`            // /sys/class/net/<iface>/carrier
+	CarrierChanges   int64  `fileName:"carrier_changes"`    // /sys/class/net/<iface>/carrier_changes
+	CarrierUpCount   int64  `fileName:"carrier_up_count"`   // /sys/class/net/<iface>/carrier_up_count
+	CarrierDownCount int64  `fileName:"carrier_down_count"` // /sys/class/net/<iface>/carrier_down_count
+	DevID            int64  `fileName:"dev_id"`             // /sys/class/net/<iface>/dev_id
+	Dormant          int64  `fileName:"dormant"`            // /sys/class/net/<iface>/dormant
 	Duplex           string `fileName:"duplex"`             // /sys/class/net/<iface>/duplex
-	Flags            uint64 `fileName:"flags"`              // /sys/class/net/<iface>/flags
+	Flags            int64  `fileName:"flags"`              // /sys/class/net/<iface>/flags
 	IfAlias          string `fileName:"ifalias"`            // /sys/class/net/<iface>/ifalias
-	IfIndex          uint64 `fileName:"ifindex"`            // /sys/class/net/<iface>/ifindex
-	IfLink           uint64 `fileName:"iflink"`             // /sys/class/net/<iface>/iflink
-	LinkMode         uint64 `fileName:"link_mode"`          // /sys/class/net/<iface>/link_mode
-	MTU              uint64 `fileName:"mtu"`                // /sys/class/net/<iface>/mtu
-	NameAssignType   uint64 `fileName:"name_assign_type"`   // /sys/class/net/<iface>/name_assign_type
-	NetDevGroup      uint64 `fileName:"netdev_group"`       // /sys/class/net/<iface>/netdev_group
+	IfIndex          int64  `fileName:"ifindex"`            // /sys/class/net/<iface>/ifindex
+	IfLink           int64  `fileName:"iflink"`             // /sys/class/net/<iface>/iflink
+	LinkMode         int64  `fileName:"link_mode"`          // /sys/class/net/<iface>/link_mode
+	MTU              int64  `fileName:"mtu"`                // /sys/class/net/<iface>/mtu
+	NameAssignType   int64  `fileName:"name_assign_type"`   // /sys/class/net/<iface>/name_assign_type
+	NetDevGroup      int64  `fileName:"netdev_group"`       // /sys/class/net/<iface>/netdev_group
 	OperState        string `fileName:"operstate"`          // /sys/class/net/<iface>/operstate
 	PhysPortID       string `fileName:"phys_port_id"`       // /sys/class/net/<iface>/phys_port_id
 	PhysPortName     string `fileName:"phys_port_name"`     // /sys/class/net/<iface>/phys_port_name
 	PhysSwitchID     string `fileName:"phys_switch_id"`     // /sys/class/net/<iface>/phys_switch_id
-	Speed            uint64 `fileName:"speed"`              // /sys/class/net/<iface>/speed
-	TxQueueLen       uint64 `fileName:"tx_queue_len"`       // /sys/class/net/<iface>/tx_queue_len
-	Type             uint64 `fileName:"type"`               // /sys/class/net/<iface>/type
+	Speed            int64  `fileName:"speed"`              // /sys/class/net/<iface>/speed
+	TxQueueLen       int64  `fileName:"tx_queue_len"`       // /sys/class/net/<iface>/tx_queue_len
+	Type             int64  `fileName:"type"`               // /sys/class/net/<iface>/type
 }
 
 // NetClass is collection of info for every interface (iface) in /sys/class/net. The map keys
@@ -80,7 +82,7 @@ func (fs FS) NewNetClass() (NetClass, error) {
 	for _, deviceDir := range devices {
 		interfaceClass, err := netClass.parseNetClassIface(path + "/" + deviceDir.Name())
 		if err != nil {
-			continue
+			return nil, err
 		}
 		interfaceClass.Name = deviceDir.Name()
 		netClass[deviceDir.Name()] = *interfaceClass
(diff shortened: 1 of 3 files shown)


SuperQ:
Thanks for the contribution.  I'm traveling right now, but will try and get this reviewed soon.

sysfs/net_class.go
```diff
@@ -0,0 +1,127 @@
...
+			continue
+		}
+		interfaceClass.Name = deviceDir.Name()
+		netClass[deviceDir.Name()] = *interfaceClass
+	}
+	return netClass, nil
+}
+
+func (nc NetClass) parseInterfaceNetClass(devicePath string) (*interfaceNetClass, error) {
+	fields := []string{"AddrAssignType", "AddrLen", "Address", "Broadcast", "Carrier", "CarrierChanges", "CarrierUpCount", "CarrierDownCount", "DevId", "Dormant", "Duplex", "Flags", "IfAlias", "IfIndex", "IfLink", "LinkMode", "Mtu", "NameAssignType", "NetDevGroup", "OperState", "PhysPortId", "PhysPortName", "PhysSwitchId", "Speed", "TxQueueLen", "Type"}
```
grobie: Why do you list all the field names again, given that you already use the reflect package? With `reflect.ValueOf(&interfaceClass).NumField()` you can get the number of fields and with `reflect.ValueOf(&interfaceClass).Field(index)` you can access a field by index.

sysfs/net_class.go
```diff
@@ -0,0 +1,127 @@
...
+	interfaceType := reflect.TypeOf(interfaceClass)
+
+	for _, fieldName := range fields {
+		fieldType, found := interfaceType.FieldByName(fieldName)
+		if !found {
+			continue
+		}
+		fieldValue := interfaceElem.FieldByName(fieldName)
+
+		fileContents, err := ioutil.ReadFile(devicePath + "/" + fieldType.Tag.Get("fileName"))
```
grobie: There should be a check whether fileName tag is set before trying to open such file.

sysfs/net_class.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+type interfaceNetClass struct {
+	Name             string `fileName:"name"`               // Interface name
```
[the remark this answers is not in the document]
klatys: That is a correct note and mistake on my side, I'll remove the tag, thanks.

sysfs/net_class.go
```diff
@@ -0,0 +1,127 @@
...
+	AddrAssignType   uint64 `fileName:"addr_assign_type"`   // /sys/class/net/<iface>/addr_assign_type
+	AddrLen          uint64 `fileName:"addr_len"`           // /sys/class/net/<iface>/addr_len
+	Address          string `fileName:"address"`            // /sys/class/net/<iface>/address
+	Broadcast        string `fileName:"broadcast"`          // /sys/class/net/<iface>/broadcast
+	Carrier          uint64 `fileName:"carrier"`            // /sys/class/net/<iface>/carrier
+	CarrierChanges   uint64 `fileName:"carrier_changes"`    // /sys/class/net/<iface>/carrier_changes
+	CarrierUpCount   uint64 `fileName:"carrier_up_count"`   // /sys/class/net/<iface>/carrier_up_count
+	CarrierDownCount uint64 `fileName:"carrier_down_count"` // /sys/class/net/<iface>/carrier_down_count
+	DevId            uint64 `fileName:"dev_id"`             // /sys/class/net/<iface>/dev_id
```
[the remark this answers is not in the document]
klatys: Did not know that. Thanks!

sysfs/net_class_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	netClass := map[string]interfaceNetClass{
+		"eth0": {Name: "eth0", AddrAssignType: 3, AddrLen: 6, Address: "01:01:01:01:01:01", Broadcast: "ff:ff:ff:ff:ff:ff", Carrier: 1, CarrierChanges: 2, CarrierUpCount: 1, CarrierDownCount: 1, DevId: 32, Dormant: 1, Duplex: "full", Flags: 4867, IfAlias: "", IfIndex: 2, IfLink: 2, LinkMode: 1, Mtu: 1500, NameAssignType: 2, NetDevGroup: 0, OperState: "up", PhysPortId: "", PhysPortName: "", PhysSwitchId: "", Speed: 1000, TxQueueLen: 1000, Type: 1},
+	}
+
+	if want, have := len(netClass), len(nc); want != have {
+		t.Errorf("want %d parsed class/net, have %d", want, have)
+	}
+
+	fields := []string{"AddrAssignType", "AddrLen", "Address", "Broadcast", "Carrier", "CarrierChanges", "CarrierUpCount", "CarrierDownCount", "DevId", "Dormant", "Duplex", "Flags", "IfAlias", "IfIndex", "IfLink", "LinkMode", "Mtu", "NameAssignType", "NetDevGroup", "OperState", "PhysPortId", "PhysPortName", "PhysSwitchId", "Speed", "TxQueueLen", "Type"}
```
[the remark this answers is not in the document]
klatys: absolutely, did not know that. thanks!

sysfs/net_class.go
klatys: much better solution, thanks!

sysfs/net_class.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+type interfaceNetClass struct {
```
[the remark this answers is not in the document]
klatys: Ok, does `NetClassInterface` seem like a better name to you? I'm not sure whether using word `interface` is really a good idea - can be misleading language-wise, on the other hand something like `device` is misleading as well

sysfs/net_class.go
```diff
@@ -0,0 +1,127 @@
...
+	PhysPortName     string `fileName:"phys_port_name"`     // /sys/class/net/<iface>/phys_port_name
+	PhysSwitchId     string `fileName:"phys_switch_id"`     // /sys/class/net/<iface>/phys_switch_id
+	Speed            uint64 `fileName:"speed"`              // /sys/class/net/<iface>/speed
+	TxQueueLen       uint64 `fileName:"tx_queue_len"`       // /sys/class/net/<iface>/tx_queue_len
+	Type             uint64 `fileName:"type"`               // /sys/class/net/<iface>/type
+}
+
+type NetClass map[string]interfaceNetClass
+
+// NewNetDev returns kernel/system statistics read from /proc/net/dev.
```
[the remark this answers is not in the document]
klatys: I fixed the comment, but neither of those commands returned anything (neither before nor after the fix) not sure if that's correct or I'm doing sth wrong?

sysfs/net_class.go
klatys: definitely. Added, thanks for that

sysfs/net_class_test.go
```diff
@@ -0,0 +1,39 @@
...
+		t.Fatal(err)
+	}
+
+	nc, err := fs.NewNetClass()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	netClass := NetClass{
+		"eth0": {Name: "eth0", AddrAssignType: 3, AddrLen: 6, Address: "01:01:01:01:01:01", Broadcast: "ff:ff:ff:ff:ff:ff", Carrier: 1, CarrierChanges: 2, CarrierUpCount: 1, CarrierDownCount: 1, DevID: 32, Dormant: 1, Duplex: "full", Flags: 4867, IfAlias: "", IfIndex: 2, IfLink: 2, LinkMode: 1, MTU: 1500, NameAssignType: 2, NetDevGroup: 0, OperState: "up", PhysPortID: "", PhysPortName: "", PhysSwitchID: "", Speed: 1000, TxQueueLen: 1000, Type: 1},
```
grobie: It'd help readability to have a line break after each attribute/value pair.

sysfs/net_class.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return fs.NewNetClass()
+}
+
+// NewNetClass returns info for all net interfaces read from /sys/class/net/<interface>.
+func (fs FS) NewNetClass() (NetClass, error) {
+	return newNetClass(fs.Path("class/net"))
+}
+
+// NewNetClass returns info for all net interfaces read from /sys/class/net/<interface>.
```
grobie: This should start with a lowercase n to document the private function. I wonder why not just merging this fuction with the public NewNetClass function, given it's the only caller.

sysfs/net_class.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+// NetClassInterface contains info from files in /sys/class/net/<interface>
```
grobie: The kernel documentation uses `<iface>`, what about using the same?

sysfs/net_class.go
```diff
@@ -0,0 +1,135 @@
...
+// for single interface
```
grobie: What about adding `for a single interface (iface).` in parenthesis, to emphasize that it's not about a language/golang interface. Please end all comment sentences with full stops (following the golang style guide).

sysfs/net_class.go
```diff
@@ -0,0 +1,135 @@
...
+			continue
+		}
+
+		fileContents, err := ioutil.ReadFile(devicePath + "/" + fieldType.Tag.Get("fileName"))
+		if err != nil {
+			continue
+		}
+		value := strings.TrimSpace(string(fileContents))
+
+		if fieldValue.Kind() == reflect.Uint64 {
```
grobie: I'd suggest to use a switch statement here and error handling in the default case, to ensure all value types are handled.

```go
switch fieldValue.Kind() {
case reflect.Uint64:
// ...
default:
  return nil, fmt.Errorf("unhandled type %q", fieldValue.Kind())
}

File: sysfs/net_class.go [hunk 13]
```diff
@@ -0,0 +1,135 @@
...
+	PhysPortID       string `fileName:"phys_port_id"`       // /sys/class/net/<iface>/phys_port_id
+	PhysPortName     string `fileName:"phys_port_name"`     // /sys/class/net/<iface>/phys_port_name
+	PhysSwitchID     string `fileName:"phys_switch_id"`     // /sys/class/net/<iface>/phys_switch_id
+	Speed            uint64 `fileName:"speed"`              // /sys/class/net/<iface>/speed
+	TxQueueLen       uint64 `fileName:"tx_queue_len"`       // /sys/class/net/<iface>/tx_queue_len
+	Type             uint64 `fileName:"type"`               // /sys/class/net/<iface>/type
+}
+
+// NetClass is collection of info for every interface in /sys/class/net. The map keys
+// are interface names.
```
grobie: Probably here again to make it clearer `interface (iface)`.

sysfs/net_class.go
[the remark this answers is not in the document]
grobie: There are indeed no go vet errors in the packages fortunately. A couple of linter errors though in other files (nfs and bcache). Before your changes, golint also complained about the missing or wrong godoc strings. You should see some output with `golint ./...` executed in the procfs repository root.

sysfs/net_class.go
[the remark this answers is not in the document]
grobie: I believe following the terminology chosen by the kernel developers makes the most sense. Reading the kernel documentation, it should be either `iface` or `interface`. I'm generally not a fan of abbreviations and would slightly lean towards `interface`. In the doc strings we could use `interface (iface)` to make the context clear.

sysfs/net_class.go
klatys: makes sense. Omits the possible confusion, thanks!

sysfs/net_class.go
klatys: absolutely, merging.

SuperQ:
@grobie Do you have any suggestions to replace `reflect` to avoid the huge amount of copy-paste necessary for things like [this](https://github.com/prometheus/node_exporter/blob/master/collector/nfsd_linux.go#L330-L398) vs [with reflect](https://github.com/prometheus/node_exporter/blob/master/collector/nfs_linux.go#L158-L169)?

grobie:
@SuperQ I often find static code easier to reason about then trying to abstract it away, my editor helps me with managing such problems. For a middleground between reflect usage and manually typed code I would look into `go generate`. I only have used it once in my life so far, but it should be possible to auto-generate your linked functions.

klatys:
@grobie Thanks for your thorough code review, I'm really glad for that!

As for the error handling - yes I did use the `continue` on purpose, for two specific reasons:

When testing this on one of our debian machines, I got to a problem parsing the `phys_*` files which cannot be read (e.g. `$ cat /sys/class/net/eth0/phys_port_id` produces `Operation not supported` error, go returns error `read ...: operation not supported`). From what our linux guys are telling me this is probably caused by the driver of the device and might not be something that unusual. The other thing is, as you mentioned, nonexistent file due to older kernel (which I could find on any of our machines) and in my opinion should not produce error as well.

The other one is rather being cautious on my side, as I can imagine an interface dir not being accessible for some reason. In such case I don't think it is correct to return an error and ignoring all other _ok_ ifaces, but that's really my point of view...

SuperQ:
@klatys We had a similar IO problem with `hwmon`.  Some kernels would return `EAGAIN`.  We wrote a special [reader function](https://github.com/prometheus/node_exporter/pull/728) for this use case.  We may want to do the same here.

grobie:
In my experience strict error handling makes it easier to debug issues. I've witnessed many lengthy debug sessions due to silenced error swallowing. I'd suggest to handle the "file not exist" and "operation not supported" cases gracefully (and probably the EAGAIN error as well), but bubble up errors in all other cases.

klatys:
Ok, np. I implemented that and changed the error handling. besides the three file errors I had to add one other - `invalid argument`: 
```
$cat /sys/class/net/lo/name_assign_type 
cat: /sys/class/net/lo/name_assign_type: Invalid argument
```
Hopefully there won't be any other errors for these "files".

Also I ran this on several machines and discovered that some values (e.g. speed) can be negative (e.g. -1) thus the type is now Int64.

PR #87, opened 2018-04-17
change values to pointers to differ missing value from 0

klatys:
For implementation in node_exporter there's a need to differentiate 0 from nil
blocks https://github.com/prometheus/node_exporter/pull/851

### DIFF
--- a/sysfs/net_class_test.go
+++ b/sysfs/net_class_test.go
@@ -29,35 +29,56 @@ func TestNewNetClass(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	var (
+		AddrAssignType   int64 = 3
+		AddrLen          int64 = 6
+		Carrier          int64 = 1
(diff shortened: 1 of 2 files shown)


sysfs/net_class_test.go
```diff
@@ -29,35 +29,56 @@ func TestNewNetClass(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	var (
+		AddrAssignType   int64 = 3
```
mdlayher: Very minor nit: all variables in the scope of a function should start with a lowercase letter.

klatys: fixed. thanks!

klatys:
Thanks for the reviews guys, can we merge this? I've got the node_exporter code ready for push, only blocked by this. Thanks a lot!

SuperQ:
Thanks for the help reviewing.

PR #88, opened 2018-04-18
Lio rework replace old PR to avoid conflict and better communication

AvengerMoJo:
Please let me know if it is better? 
The sysPath need to keep to make sure test case calling it async after the initial Stat being create. Other then that it should be following what bcache as my example. If this is still not good enough, please do let me know how to improve it. Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ point is initialized, and then the stat information is read.
 
 ```go
 fs, err := procfs.NewFS("/proc")
-stats, err := fs.NewStat()
+stats, err := fs.Stat()
 ```
 
 Some sub-packages such as `blockdevice`, require access to both the proc and sys filesystems.
--- a/proc_limits.go
+++ b/proc_limits.go
@@ -77,6 +77,13 @@ var (
 	limitsDelimiter = regexp.MustCompile("  +")
 )
 
+// NewLimits returns the current soft limits of the process.
+//
+// Deprecated: use p.Limits() instead
+func (p Proc) NewLimits() (ProcLimits, error) {
+	return p.Limits()
+}
+
 // Limits returns the current soft limits of the process.
 func (p Proc) Limits() (ProcLimits, error) {
 	f, err := os.Open(p.path("limits"))
--- a/proc_stat.go
+++ b/proc_stat.go
@@ -104,6 +104,13 @@ type ProcStat struct {
 	proc fs.FS
 }
 
+// NewStat returns the current status information of the process.
+//
+// Deprecated: use NewStat() instead
+func (p Proc) NewStat() (ProcStat, error) {
+	return p.Stat()
+}
+
 // Stat returns the current status information of the process.
 func (p Proc) Stat() (ProcStat, error) {
 	f, err := os.Open(p.path("stat"))
--- a/stat.go
+++ b/stat.go
@@ -20,6 +20,8 @@ import (
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/prometheus/procfs/internal/fs"
 )
 
 // CPUStat shows how much time the cpu spend in various stages.
@@ -139,9 +141,21 @@ func parseSoftIRQStat(line string) (SoftIRQStat, uint64, error) {
 	return softIRQStat, total, nil
 }
 
-// Stat returns an information about current kernel/system statistics.
+// NewStat returns information about current cpu/process statistics.
+// See https://www.kernel.org/doc/Documentation/filesystems/proc.txt
+//
+// Deprecated: use fs.Stat() instead
+func NewStat() (Stat, error) {
+	fs, err := NewFS(fs.DefaultProcMountPoint)
+	if err != nil {
+		return Stat{}, err
+	}
+	return fs.Stat()
+}
+
+// Stat returns information about current cpu/process statistics.
+// See https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 func (fs FS) Stat() (Stat, error) {
-	// See https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 
 	f, err := os.Open(fs.proc.Path("stat"))
 	if err != nil {


discordianfish:
(Still) LGTM. @grobie Can you be more specific what you don't like about this? This is blocking https://github.com/prometheus/node_exporter/pull/776 and I'm not very happy with taking so long to merge these PRs.

AvengerMoJo:
@SuperQ @mdlayher It has been 5 months since I submitted my lio module in node_exporter and you recommend me to split it into procfs module and I did. Now this is blocking my lio module and I have been asking for months to understand what I need to do to pass this? It still unclear to me what is missing or what do I need to do to move this forward. Please help.

AvengerMoJo:
Hi anyone update ? It has been a year.

pgier:
@AvengerMoJo sorry for the very long delay on this one!  Any chance you can rebase this on the latest master and we'll get it merged?

AvengerMoJo:
Of course. I have been waiting for a long time to do it. Give me a couple days to rebase it back to the latest version. Thanks

AvengerMoJo:
@pgier thanks for the notification, I just rebase my branch. We should able to merge to the master. It has been awhile I looked at this code, please let me know if I need to refactoring anything accordingly. I build this procfs module mainly was trying to fulfill the request from node_exporter request for my lio module. If I did something wrong please let me know. Thanks

iscsi/get.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	return &istats, nil
+}
+
+// isPathEnable is a utility function
+// check if the file "enable" contain enable message
+func isPathEnable(path string) (bool, error) {
+	var isEnable bool
+	isEnable = false
+	tmp, err := ioutil.ReadFile(filepath.Join(path, "enable"))
```
pgier: Should use a more meaningful variable name here instead of `tmp`

iscsi/get.go
```diff
@@ -0,0 +1,314 @@
...
+			real_udev, err := ioutil.ReadFile(lunPath + "/" + target + "/udev_path")
+			if err != nil {
+				return "", "", "", fmt.Errorf("udev_path read file err %v", err)
+			}
+			udev_number, err := os.Readlink(strings.TrimSpace(string(real_udev)))
+
+			if err != nil {
+				return "", "", "", fmt.Errorf("udev_path readlink err %v", err)
+			}
+			tmp := strings.Split(typeWithNumber, "_")
```
pgier: Give this a more meaningful name than `tmp`.

iscsi/get_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	sysfs, err := iscsi.NewFS("../fixtures")
+	if err != nil {
+		t.Fatalf("failed to access xfs fs: %v", err)
+	}
+	sysfsStat, err := sysfs.ISCSIStats()
+	if err != nil {
+		t.Errorf("unexpected test fixtures")
+	}
+
```
pgier: Should add a check here to make sure that the returned stats instance is the expected length.  It looks like the fixtures were accidentally removed in one of the recent commits, but the test still passes because a zero length stats object just skips the detailed checks.

iscsi/iscsi.go
```diff
@@ -0,0 +1,119 @@
...
+package iscsi
+
+import (
+	"path/filepath"
+
+	"github.com/prometheus/procfs/internal/fs"
+)
+
+// TARGETPATH is static sys path for iscsi target
+const TARGETPATH = "/kernel/config/target/iscsi"
```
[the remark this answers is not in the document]
AvengerMoJo: Done, thanks for the reminder.

iscsi/iscsi.go
```diff
@@ -0,0 +1,119 @@
...
+// FS represents the pseudo-filesystem sysfs, which provides an interface to
+// kernel data structures.
+type FS struct {
+	sys *fs.FS
+}
+
+// NewFS returns a new FS mounted under the given mountPoint. It will error
+// if the mount point can't be read.
+func NewFS(mountPoint string) (FS, error) {
+	sysfs, err := fs.NewFS(mountPoint)
```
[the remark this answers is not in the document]
AvengerMoJo: Are you suggesting I should add configfs into the sysfs module? or more like something like below? 

```
// ConfigFsPath is static configfs mount point
const ConfigFsPath= "/kernel/config"

// TargetPath is static sys path for iscsi target
const TargetPath = "/target/iscsi"

// TARGETCORE static sys path for backstore
const TargetCore = "/target/core"

// FS represents the pseudo-filesystem configfs, which provides an interface to
// iscsi kernel data structures.
type FS struct {
    configfs *fs.FS
}

// NewFS returns a new configfs mounted under the given mountPoint. It will error
// if the mount point can't be read.
func NewFS(configfsMountPoint string) (FS, error) {
    if strings.TrimSpace(configfsMountPoint) == "" {
        // /sys/kernel/config
        configfsMountPoint = fs.DefaultSysMountPoint + ConfigFsPath
    }
    sysfs, err := fs.NewFS(configfsMountPoint)
    if err != nil {
        return FS{}, err
    }
    return FS{&configfs}, nil
}
```

iscsi/iscsi.go
```diff
@@ -0,0 +1,119 @@
...
+	matches, err := filepath.Glob(fs.sys.Path(TARGETPATH, "/iqn*"))
+	if err != nil {
+		return nil, err
+	}
+
+	stats := make([]*Stats, 0, len(matches))
+	for _, iqnPath := range matches {
+		name := filepath.Base(iqnPath)
+		// stats
+		s, err := GetStats(fs.sys.Path(), iqnPath)
```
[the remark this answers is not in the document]
AvengerMoJo: You are correct, I believe this silly setup is my workaround for the test fixture. If my path was not actually /sys/ then the test won't pass. I believe the new call path for the test fixture should work and I get just cal iqnPath directly. I need to re-test the logic behind this.

iscsi/get.go
```diff
@@ -0,0 +1,314 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var sysPath = ""
```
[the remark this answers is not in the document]
AvengerMoJo: This is also the side effect product of the include module from iscsi to sysfs. I believe with the new setup I should no longer need this to workaround the test fixture case.

iscsi/iscsi.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			return nil, err
+		}
+		s.Name = name
```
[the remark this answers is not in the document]
AvengerMoJo: Ok, I'm rewriting it

iscsi/get.go
AvengerMoJo: done, enableReadout it is

iscsi/get.go
AvengerMoJo: done, splitTypeWithNumber it is

iscsi/get_test.go
AvengerMoJo: I need to rewrite this and add the fixture back

PR #89, opened 2018-04-18
Update ttar and fix some typos

mjtrangoni:
Hi @grobie,

I was trying to make [codespell](https://github.com/lucasdemarchi/codespell) happy, and got it working. See,
```
$ codespell -S .git*,fixture*
```
We could add this as a **CircleCI** job (node_exporter is using it right now), of course if you agree!

Thanks!

PS: The `nd => netDev` renaming action is because *codespell* identify it as `nd  ==> and, 2nd`, and renaming it is cheaper than excluding `nd`.


pgier:
LGTM

PR #91, opened 2018-05-31
ignore regular files only (symlinks are acceptable)

klatys:


grobie:
Thanks @klatys! Please use more descriptive commit messages in the future, which explain what package/functionality changes.

PR #92, opened 2018-06-11
add /proc/net/unix parsing

peonone:
Signed-off-by: madk <luc.mdk@gmail.com>

### DIFF
--- a/net_unix.go
+++ b/net_unix.go
@@ -66,7 +66,7 @@ type NetUnixLine struct {
 	Flags     NetUnixFlags
 	Type      NetUnixType
 	State     NetUnixState
-	Inode     string
+	Inode     uint64
 	Path      string
 }
 
@@ -143,6 +143,14 @@ func (u *NetUnix) parseLine(line string, hasInode bool) (*NetUnixLine, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Parse Unix domain state(%s) failed: %s", fields[netUnixStateIdx], err)
 	}
+	var inode uint64
+	if hasInode {
+		inodeStr := fields[len(fields)-2]
+		inode, err = u.parseInode(inodeStr)
+		if err != nil {
+			return nil, fmt.Errorf("Parse Unix domain inode(%s) failed: %s", inodeStr, err)
+		}
+	}
 
 	nuLine := &NetUnixLine{
 		KernelPtr: kernelPtr,
@@ -151,10 +159,9 @@ func (u *NetUnix) parseLine(line string, hasInode bool) (*NetUnixLine, error) {
 		Flags:     flags,
 		State:     state,
 		Path:      fields[len(fields)-1],
+		Inode:     inode,
 	}
-	if hasInode {
-		nuLine.Inode = fields[len(fields)-2]
-	}
+
 	return nuLine, nil
 }
 
@@ -195,6 +202,10 @@ func (u NetUnix) parseState(hexStr string) (NetUnixState, error) {
 		return 0, err
 	}
 	return NetUnixState(st), nil
+}
+
+func (u NetUnix) parseInode(inodeStr string) (uint64, error) {
+	return strconv.ParseUint(inodeStr, 10, 64)
 }
 
 func (t NetUnixType) String() string {
--- a/net_unix.go
+++ b/net_unix.go
@@ -15,6 +15,7 @@ package procfs
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -45,6 +46,8 @@ const (
 	netUnixStateConnected    = 3
 	netUnixStateDisconnected = 4
 )
+
+var errInvalidKernelPtrFmt = errors.New("Invalid Num(kernal the kernel table slot number) format")
 
 // NetUnixType is the type of the type field.
 type NetUnixType uint64
@@ -118,6 +121,10 @@ func NewNetUnixByPath(path string) (*NetUnix, error) {
 
 func (u *NetUnix) parseLine(line string, hasInode bool) (*NetUnixLine, error) {
 	fields := strings.Fields(line)
+	kernelPtr, err := u.parseKernelPtr(fields[netUnixKernelPtrIdx])
+	if err != nil {
+		return nil, fmt.Errorf("Parse Unix domain num(%s) failed: %s", fields[netUnixKernelPtrIdx], err)
+	}
 	users, err := u.parseUsers(fields[netUnixRefCountIdx])
 	if err != nil {
 		return nil, fmt.Errorf("Parse Unix domain ref count(%s) failed: %s", fields[netUnixRefCountIdx], err)
@@ -136,7 +143,7 @@ func (u *NetUnix) parseLine(line string, hasInode bool) (*NetUnixLine, error) {
 	}
 
 	nuLine := &NetUnixLine{
-		KernelPtr: u.parseKernelPtr(fields[netUnixKernelPtrIdx]),
+		KernelPtr: kernelPtr,
 		RefCount:  users,
 		Type:      typ,
 		Flags:     flags,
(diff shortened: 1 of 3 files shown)


net_unix.go
```diff
@@ -0,0 +1,225 @@
...
+
+	netUnixFlagListen = 1 << 16
+
+	netUnixStateUnconnected  = 1
+	netUnixStateConnecting   = 2
+	netUnixStateConnected    = 3
+	netUnixStateDisconnected = 4
+)
+
+// NetUnixType is the type of the type field
```
SuperQ: Minor style nit, all comments should be full sentences. End these with a `.`.

net_unix.go
```diff
@@ -0,0 +1,225 @@
...
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	nu := &NetUnix{
+		Rows: make([]*NetUnixLine, 0, 32),
+	}
+	scanner := bufio.NewScanner(f)
+	// omit the header line
```
SuperQ: Similar here, but the sentences need to start with a capitol letter as well.

net_unix.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	if hasInode {
+		nuLine.Inode = fields[len(fields)-2]
+	}
+	return nuLine, nil
+}
+
+func (u NetUnix) parseKernelPtr(str string) string {
+	length := len(str)
+	if length > 1 && str[length-1] == ':' {
```
SuperQ: I would use `strings. HasSuffix(str, ":")`.

net_unix.go
```diff
@@ -0,0 +1,225 @@
...
+		State:     state,
+		Path:      fields[len(fields)-1],
+	}
+	if hasInode {
+		nuLine.Inode = fields[len(fields)-2]
+	}
+	return nuLine, nil
+}
+
+func (u NetUnix) parseKernelPtr(str string) string {
```
SuperQ: Shouldn't this function also return an error if this field doesn't end with a `:`? According to `af_unix.c#L2815`, it should always.

SuperQ:
Looks great, mostly minor nits.

peonone:
Fixed those nits.

Also updated the comment related to `Inode` field, and updated the type of this field to `uint64`, because it's  `unsigned long` in the code which generates the proc file.

net_unix.go
```diff
@@ -0,0 +1,244 @@
...
+	Protocol  uint64
+	Flags     NetUnixFlags
+	Type      NetUnixType
+	State     NetUnixState
+	Inode     uint64
+	Path      string
+}
+
+// NetUnix holds the data read from /proc/net/unix.
+type NetUnix struct {
```
grobie: What's the motivation of using a struct over a simple slice `type NetUinx []*NetUnixLine`?

net_unix.go
```diff
@@ -0,0 +1,244 @@
...
+			return nu, err
+		}
+		nu.Rows = append(nu.Rows, item)
+	}
+
+	return nu, nil
+}
+
+func (u *NetUnix) parseLine(line string, hasInode bool) (*NetUnixLine, error) {
+	fields := strings.Fields(line)
```
grobie: Please check the length of the fields first to prevent out of bounds panics. We've seen all kinds of unexpected errors in procfs, from malformed lines, to unexpected blank lines and unexpected format changes. The library should handle these gracefully.

net_unix.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+	}
+
+	nuLine := &NetUnixLine{
+		KernelPtr: kernelPtr,
+		RefCount:  users,
+		Type:      typ,
+		Flags:     flags,
+		State:     state,
+		Path:      fields[len(fields)-1],
```
grobie: The path attribute can be optional, this needs to be handled.

net_unix.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Parse Unix domain type(%s) failed: %s", fields[netUnixTypeIdx], err)
+	}
+	state, err := u.parseState(fields[netUnixStateIdx])
+	if err != nil {
+		return nil, fmt.Errorf("Parse Unix domain state(%s) failed: %s", fields[netUnixStateIdx], err)
+	}
+	var inode uint64
+	if hasInode {
+		inodeStr := fields[len(fields)-2]
```
grobie: This doesn't work on lines without a path. Count from the beginning.

net_unix.go
peonone: There are few methods defined on `NetUinx` which can be called during the parsing process.
With a simple slice, the `NetUnix` itself changes once a line get parsed.
I think this way is not good and might cause potential bugs in future if we add more logic to the method, for example, pass the `NetUnix` instance to another place, but it points to the partial result. With a struct, we have the unique `NetUnix` instance.

peonone:
@grobie 's comments make sense, I added validation for the length and handled the case of a blank path.

net_unix.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// NewNetUnix returns data read from /proc/net/unix.
+func (fs FS) NewNetUnix() (*NetUnix, error) {
+	return NewNetUnixByPath(fs.Path("net/unix"))
+}
+
+// NewNetUnixByPath returns data read from /proc/net/unix by file path.
+// It might returns an error with partial parsed data, if an error occur after some data parsed.
+func NewNetUnixByPath(path string) (*NetUnix, error) {
```
mdlayher: Nit: I would accept `r io.Reader` in this function and open a file in the method above instead.  This gives you and your caller the most flexibility.

net_unix.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	defer f.Close()
+	nu := &NetUnix{
+		Rows: make([]*NetUnixLine, 0, 32),
+	}
+	scanner := bufio.NewScanner(f)
+	// Omit the header line.
+	scanner.Scan()
+	header := scanner.Text()
+	// From the man page of proc(5), it dose not contain an Inode field,
```
mdlayher: *does

net_unix.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (u NetUnix) parseKernelPtr(str string) (string, error) {
+	if !strings.HasSuffix(str, ":") {
+		return "", errInvalidKernelPtrFmt
+	}
+	return str[:len(str)-1], nil
+}
+
+func (u NetUnix) parseUsers(hexStr string) (uint64, error) {
```
mdlayher: Nit: it feels a little weird to have these be methods when they don't actually do anything with the method receiver at all, but I suppose it makes sense to avoid littering the package with unexported functions that are only used for a very specific file.

net_unix.go
```diff
@@ -0,0 +1,268 @@
...
+	case netUnixTypeSeqpacket:
+		return "seqpacket"
+	}
+	return "unknown"
+}
+
+func (f NetUnixFlags) String() string {
+	switch f {
+	case netUnixFlagListen:
+		return "accepton"
```
mdlayher: Nit: should this be "listen" to match the flag name?  I'm not familiar with this file.

peonone: I referred to the [source code](https://elixir.bootlin.com/linux/v4.17/source/net/unix/af_unix.c#L2819) which writes the proc file, `listen` also makes sense and easier to understand, I will update it.

net_unix.go
peonone: Yes, the purpose of making them as methods is preventing package level functions with a prefix.

pgier:
@peonone any chance you can rebase this one against latest master?  Sorry this wasn't merged earlier.  After fixing the conflict I think we can merge it.

peonone:
> @peonone any chance you can rebase this one against latest master? Sorry this wasn't merged earlier. After fixing the conflict I think we can merge it.

@pgier Sure, I will do this later this week.

peonone:
@pgier Done

PR #94, opened 2018-06-12
Add parser for sysfs devices/system/cpu/...

SuperQ:
* Add parser for `devices/system/cpu/cpu*/cpufreq/...`.
* Add helper `util.ReadUintFromFile()`.
* Move `sysReadFile()` helper from `sysfs` to `internal/util`.

### DIFF
--- a/internal/util/parse.go
+++ b/internal/util/parse.go
@@ -15,8 +15,10 @@ package util
 
 import (
 	"io/ioutil"
+	"os"
 	"strconv"
 	"strings"
+	"syscall"
 )
 
 // ParseUint32s parses a slice of strings into a slice of uint32s.
@@ -59,4 +61,27 @@ func ReadUintFromFile(path string) (uint64, error) {
 		return 0, err
 	}
 	return value, nil
+}
+
+// sysReadFile is a simplified ioutil.ReadFile that invokes syscall.Read directly.
+// https://github.com/prometheus/node_exporter/pull/728/files
+func SysReadFile(file string) ([]byte, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	// On some machines, hwmon drivers are broken and return EAGAIN.  This causes
+	// Go's ioutil.ReadFile implementation to poll forever.
+	//
+	// Since we either want to read data or bail immediately, do the simplest
+	// possible read using syscall directly.
+	b := make([]byte, 128)
+	n, err := syscall.Read(int(f.Fd()), b)
+	if err != nil {
+		return nil, err
+	}
+
+	return b[:n], nil
 }
--- a/internal/util/parse.go
+++ b/internal/util/parse.go
@@ -51,6 +51,7 @@ func ParseUint64s(ss []string) ([]uint64, error) {
 	return us, nil
 }
 
+// ReadUintFromFile reads a file and attempts to parse a uint64 from it.
 func ReadUintFromFile(path string) (uint64, error) {
 	data, err := ioutil.ReadFile(path)
 	if err != nil {
@@ -63,7 +64,7 @@ func ReadUintFromFile(path string) (uint64, error) {
 	return value, nil
 }
 
-// sysReadFile is a simplified ioutil.ReadFile that invokes syscall.Read directly.
+// SysReadFile is a simplified ioutil.ReadFile that invokes syscall.Read directly.
 // https://github.com/prometheus/node_exporter/pull/728/files
 func SysReadFile(file string) ([]byte, error) {
 	f, err := os.Open(file)
--- a/internal/util/parse.go
+++ b/internal/util/parse.go
@@ -13,7 +13,11 @@
 
 package util
 
-import "strconv"
+import (
+	"io/ioutil"
+	"strconv"
+	"strings"
+)
 
 // ParseUint32s parses a slice of strings into a slice of uint32s.
 func ParseUint32s(ss []string) ([]uint32, error) {
@@ -43,4 +47,16 @@ func ParseUint64s(ss []string) ([]uint64, error) {
 	}
 
 	return us, nil
+}
+
+func ReadUintFromFile(path string) (uint64, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+	value, err := strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64)
+	if err != nil {
+		return 0, err
+	}
+	return value, nil
 }
--- a/internal/util/parse.go
+++ b/internal/util/parse.go
@@ -14,6 +14,7 @@
 package util
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 	"strconv"
@@ -62,10 +63,10 @@ func ReadUintFromFile(path string) (uint64, error) {
 
 // SysReadFile is a simplified ioutil.ReadFile that invokes syscall.Read directly.
 // https://github.com/prometheus/node_exporter/pull/728/files
-func SysReadFile(file string) ([]byte, error) {
+func SysReadFile(file string) (string, error) {
 	f, err := os.Open(file)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 	defer f.Close()
 
@@ -77,8 +78,8 @@ func SysReadFile(file string) ([]byte, error) {
 	b := make([]byte, 128)
 	n, err := syscall.Read(int(f.Fd()), b)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
-	return b[:n], nil
+	return string(bytes.TrimSpace(b[:n])), nil
 }
--- a/internal/util/parse.go
+++ b/internal/util/parse.go
@@ -57,11 +57,7 @@ func ReadUintFromFile(path string) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
-	value, err := strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64)
-	if err != nil {
-		return 0, err
-	}
-	return value, nil
+	return strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64)
 }
 
 // SysReadFile is a simplified ioutil.ReadFile that invokes syscall.Read directly.
--- a/sysfs/system_cpu.go
+++ b/sysfs/system_cpu.go
@@ -24,9 +24,15 @@ import (
 
 // SystemCpuCpufreq contains stats from devices/system/cpu/cpu[0-9]*/cpufreq/...
 type SystemCpuCpufreq struct {
-	CurrentFrequency uint64
-	MinimumFrequency uint64
-	MaximumFrequency uint64
+	CurrentFrequency   uint64
+	MinimumFrequency   uint64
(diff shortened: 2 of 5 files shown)


SuperQ:
Ping @mdlayher :-)

mdlayher:
Ack, sorry, out of town. Will try to get to this on Monday.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+// NewSystemCpufreq returns CPU frequency metrics for all CPUs.
+func (fs FS) NewSystemCpufreq() (SystemCpufreq, error) {
+	cpus, err := filepath.Glob(fs.Path("devices/system/cpu/cpu[0-9]*"))
+	if err != nil {
+		return SystemCpufreq{}, err
+	}
+
+	systemCpufreq := SystemCpufreq{}
+	for _, cpu := range cpus {
+		_, cpuName := filepath.Split(cpu)
```
discordianfish: Any reason to not use filepath.Base() instead?

SuperQ: Oh, no reason, just overlooked.

SuperQ:
Hrm, tests are failing due to something else:
```
# honnef.co/go/tools/lint/lintutil
../../../honnef.co/go/tools/lint/lintutil/util.go:273: undefined: types.SizesFor
make: *** [/home/travis/gopath/bin/staticcheck] Error 2
```

SuperQ:
@grobie Poke? Please? 😀

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+	TransitionLatency  uint64
+	AvailableGovernors string
+	Driver             string
+	Govenor            string
+	RelatedCpus        string
+	SetSpeed           string
+}
+
+// SystemCpufreq is a collection of SystemCpuCpufreq for every CPU.
+type SystemCpufreq map[string]SystemCpuCpufreq
```
mdlayher: Since there are no methods on this type, I'd remove the type.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// SystemCpuCpufreq contains stats from devices/system/cpu/cpu[0-9]*/cpufreq/...
+type SystemCpuCpufreq struct {
```
mdlayher: *CPU throughout, as golint will mention too.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+	fs, err := NewFS(DefaultMountPoint)
+	if err != nil {
+		return nil, err
+	}
+
+	return fs.NewSystemCpufreq()
+}
+
+// NewSystemCpufreq returns CPU frequency metrics for all CPUs.
+func (fs FS) NewSystemCpufreq() (SystemCpufreq, error) {
```
mdlayher: TBH I think my preference would be to return a slice with the CPU's name as a member of the structure.  Maps are usually a little gross in APIs.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+	systemCpufreq := SystemCpufreq{}
+	for _, cpu := range cpus {
+		cpuName := filepath.Base(cpu)
+		cpuNum := strings.TrimPrefix(cpuName, "cpu")
+
+		cpufreq := SystemCpuCpufreq{}
+		cpuCpufreqPath := filepath.Join(cpu, "cpufreq")
+		if _, err := os.Stat(cpuCpufreqPath); os.IsNotExist(err) {
+			continue
+		} else {
```
mdlayher: You're continuing above, so drop the else entirely and remove a level of indentation from this whole block.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	systemCpufreq := SystemCpufreq{}
+	for _, cpu := range cpus {
+		cpuName := filepath.Base(cpu)
+		cpuNum := strings.TrimPrefix(cpuName, "cpu")
+
+		cpufreq := SystemCpuCpufreq{}
+		cpuCpufreqPath := filepath.Join(cpu, "cpufreq")
+		if _, err := os.Stat(cpuCpufreqPath); os.IsNotExist(err) {
```
mdlayher: Need to handle the default "err != nil" case.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+			continue
+		} else {
+			if _, err = os.Stat(filepath.Join(cpuCpufreqPath, "scaling_cur_freq")); err == nil {
```
mdlayher: Nit: I'd personally go with a loop with the file names as the slice elements over if/else if.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			systemCpufreq[cpuNum] = cpufreq
+		}
+	}
+
+	return systemCpufreq, nil
+}
+
+func parseCpufreqCpuinfo(prefix string, cpuPath string) (SystemCpuCpufreq, error) {
+	current, err := util.ReadUintFromFile(filepath.Join(cpuPath, prefix+"_cur_freq"))
```
mdlayher: Nit: you can probably shorten this up some by using loops for all of the types that need "ReadUintFromFile" and then "SysReadFile".

sysfs/system_cpu.go
mdlayher: My personal preference would be a slice with field Name in both cases.  Maps can be clunky to use and the undefined map iteration order behavior bites a lot of people on a regular basis.

internal/util/parse.go
```diff
@@ -44,3 +50,39 @@ func ParseUint64s(ss []string) ([]uint64, error) {
...
 	return us, nil
 }
+
+// ReadUintFromFile reads a file and attempts to parse a uint64 from it.
+func ReadUintFromFile(path string) (uint64, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+	value, err := strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64)
```
mdlayher: Nit: you could just `return strconv.ParseUint(/* ...  */)` at this point.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,159 @@
...
+	TransitionLatency  uint64
+	AvailableGovernors string
+	Driver             string
+	Govenor            string
+	RelatedCpus        string
+	SetSpeed           string
+}
+
+// SystemCpufreq is a collection of SystemCPUCpufreq for every CPU.
+type SystemCpufreq map[string]SystemCPUCpufreq
```
mdlayher: Did you see my previous thought?

> Since there are no methods on this type, I'd remove the type.

internal/util/parse.go
```diff
@@ -44,3 +50,39 @@ func ParseUint64s(ss []string) ([]uint64, error) {
...
+	//
+	// Since we either want to read data or bail immediately, do the simplest
+	// possible read using syscall directly.
+	b := make([]byte, 128)
+	n, err := syscall.Read(int(f.Fd()), b)
+	if err != nil {
+		return nil, err
+	}
+
+	return b[:n], nil
```
mdlayher: Since it seems like a common need, we could `bytes.TrimSpace` on the return value here too.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,159 @@
...
+	TransitionLatency  uint64
+	AvailableGovernors string
+	Driver             string
+	Govenor            string
+	RelatedCpus        string
+	SetSpeed           string
+}
+
+// SystemCpufreq is a collection of SystemCPUCpufreq for every CPU.
+type SystemCpufreq []SystemCPUCpufreq
```
mdlayher: Since this type has no methods, I would probably remove it and just return the slice directly.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,159 @@
...
+			return SystemCpufreq{}, err
+		}
+		cpufreq.Name = cpuNum
+		systemCpufreq = append(systemCpufreq, *cpufreq)
+	}
+
+	return systemCpufreq, nil
+}
+
+func (s SystemCpufreq) parseCpufreqCpuinfo(prefix string, cpuPath string) (*SystemCPUCpufreq, error) {
```
mdlayher: Here's an idea to tighten up this function: https://play.golang.org/p/T35yJhQlsmZ

Same logic could apply to the SysReadFile parts with a second loop.

sysfs/system_cpu.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// SystemCPUCpufreq contains stats from devices/system/cpu/cpu[0-9]*/cpufreq/...
+type SystemCPUCpufreq struct {
```
mdlayher: Nit: I would go with something like `SystemCPUFrequency` as that's more human friendly than the Linux directory name.

sysfs/system_cpu.go
SuperQ: It does now. `func (s SystemCpufreq) parseCpufreqCpuinfo(...)`

sysfs/system_cpu.go
SuperQ: Ahh, good idea.

SuperQ:
@mdlayher Any other issues to fix up?

sysfs/system_cpu.go
SuperQ: I prefer the literal Linux names most of the time.

mdlayher:
Sorry, not much spare time this week.  @grobie and @discordianfish ?

sysfs/system_cpu.go
discordianfish: I think @mdlayher means getting rid of `SystemCpufreq` and returning `[]SystemCPUCpufreq` instead and I agree on that.

sysfs/system_cpu.go
discordianfish: tbh, I don't like either of these names. The name should convey what this is about. So maybe (System)CPUStats?

SuperQ: I strongly prefer to use the names we get from Linux, this is the least surprising mapping between the library and the kernel.

discordianfish: I get where you're coming from but if you just look at the code without knowing the linux names it's just hard to read. Let's append at least `Stats` or something because this struct isn't a "frequency" then?

internal/util/parse.go
SuperQ: Also returning it as a `string` makes sense.

SuperQ:
Ok, I've cleaned up a bunch of things. I think I've resolved all the comments, PTAL.

metalmatze:
Looking at this I can't find any issues regarding the Go code itself (except some nits that don't change anything, probably personal preference). As I said earlier today, I'm not an expert when it comes to these internal APIs.
So LGTM.

SuperQ:
@discordianfish Any last issues?

PR #101, opened 2018-07-02
Add iscsi information parser

hassanshabbirahmed:
Adding a very simple iscsi status parser

### DIFF
--- a/sysfs/iscsi.go
+++ b/sysfs/iscsi.go
@@ -1,0 +1,85 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


SuperQ:
If you look at the way other sysfs collectors work, you can see that they implement the `filepath.Glob()` in the `fs.go` file.  Please take a look at how `net_class.go` is implemented. This is the required design pattern.

hassanshabbirahmed:
Hey @SuperQ I've morphed it according to the design pattern you mentioned. Would you mind taking a look again.

SuperQ:
It looks like your fork is far enough behind the current master that you're missing the change from Travis to CircleCI.

Would you please sync up your fork and rebase this change?

SuperQ:
Also, this change is missing tests and fixtures like net_class_test.go has.

pgier:
I think it might make sense to move this code into the new iscsi package added in #171 
Also, this needs some unit tests.

sysfs/iscsi.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2015 The Prometheus Authors
```
pgier: Should be 2019

AvengerMoJo:
@hassanshabbirahmed this is reading from the iscsi initiator point of view. My node_exporter module is from the iscsi target point of view. I think we can merge the two into one.

discordianfish:
I'm going to close this PR. Let's reopen if somebody has time to move it to the iscsi package.

Issue #95, opened 2018-06-27
Add parser for CIFS client stats


SuperQ:
The CIFS kernel module includes client stats in `/proc/fs/cifs/Stats`, but only if the kernel is configured with `CONFIG_CIFS_STATS`.

It includes per mountpoint info, so it should be pretty useful.

SuperQ:
Here's a sample stats file:
```
Resources in use
CIFS Session: 1
Share (unique mount targets): 2
SMB Request/Response Buffer: 1 Pool size: 5
SMB Small Req/Resp Buffer: 1 Pool size: 30
Operations (MIDs): 0

0 session 0 share reconnects
Total vfs operations: 79 maximum at one time: 2

1) \\server\share
SMBs: 279
Negotiates: 0 sent 0 failed
SessionSetups: 0 sent 0 failed
Logoffs: 0 sent 0 failed
TreeConnects: 0 sent 0 failed
TreeDisconnects: 0 sent 0 failed
Creates: 0 sent 2 failed
Closes: 0 sent 0 failed
Flushes: 0 sent 0 failed
Reads: 0 sent 0 failed
Writes: 0 sent 0 failed
Locks: 0 sent 0 failed
IOCTLs: 0 sent 0 failed
Cancels: 0 sent 0 failed
Echos: 0 sent 0 failed
QueryDirectories: 0 sent 6 failed
ChangeNotifies: 0 sent 0 failed
QueryInfos: 0 sent 0 failed
SetInfos: 0 sent 0 failed
OplockBreaks: 0 sent 0 failed
```

shibumi:
This is what I get on CentOS7:
```
Resources in use
CIFS Session: 1
Share (unique mount targets): 1
SMB Request/Response Buffer: 1 Pool size: 5
SMB Small Req/Resp Buffer: 1 Pool size: 30
Operations (MIDs): 0

0 session 0 share reconnects
Total vfs operations: 21 maximum at one time: 2

1) \\server\\share
SMBs: 13 Oplocks breaks: 0
Reads:  0 Bytes: 0
Writes: 0 Bytes: 0
Flushes: 0
Locks: 0 HardLinks: 0 Symlinks: 0
Opens: 0 Closes: 0 Deletes: 0
Posix Opens: 0 Posix Mkdirs: 0
Mkdirs: 0 Rmdirs: 0
Renames: 0 T2 Renames 0
FindFirst: 2 FNext 0 FClose 0
```

@SuperQ which Distribution is this?

[...]

PR #103, opened 2018-07-16
CIFS client support for Linux

shibumi:
Supported are any client statistics in SMB1 and SMB2. This should fix https://github.com/prometheus/procfs/issues/95

### DIFF
--- a/cifs/parse_cifs.go
+++ b/cifs/parse_cifs.go
@@ -1,0 +1,124 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cifs
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+)
+
+// parseHeader parses our SMB header
+func parseHeader(line string, header map[string]uint64) error {
+	for _, regexpHeader := range regexpHeaders {
+		match := regexpHeader.FindStringSubmatch(line)
+		if match == nil {
+			continue
+		}
+		for index, name := range regexpHeader.SubexpNames() {
+			if index == 0 || name == "" {
+				continue
+			}
+			value, err := strconv.ParseUint(match[index], 10, 64)
+			if nil != err {
+				return fmt.Errorf("invalid value in header")
+			}
+			header[name] = value
+		}
+	}
+	return nil
+}
+
+// parseSMBStats parses a SMB block
+func parseSMBStats(line string, stats map[string]uint64, sessionIDs *SessionIDs) error {
+	for _, regexpSMB := range regexpSMBs {
+		match := regexpSMB.FindStringSubmatch(line)
+		if match == nil {
+			continue
+		}
+		for index, name := range regexpSMB.SubexpNames() {
+			if index == 0 || name == "" {
+				continue
+			}
+			switch name {
+			case "sessionID":
+				value, err := strconv.ParseUint(match[index], 10, 64)
+				if nil != err {
+					return fmt.Errorf("type mismatch for sessionID")
+				}
+				sessionIDs.SessionID = value
+			case "server":
+				if match[index] != "" {
+					sessionIDs.Server = match[index]
+				}
+			case "share":
+				if match[index] != "" {
+					sessionIDs.Share = match[index]
+				}
+			default:
+				value, err := strconv.ParseUint(match[index], 10, 64)
+				if nil != err {
+					return fmt.Errorf("invalid value in SMB Statistics")
+				}
+				stats[name] = value
+			}
+		}
+		return nil
+	}
+	return nil
+}
+
+// ParseClientStats returns stats read from /proc/fs/cifs/Stats
+func ParseClientStats(r io.Reader) (*ClientStats, error) {
+	stats := &ClientStats{}
+	stats.Header = make(map[string]uint64)
+	scanner := bufio.NewScanner(r)
+	var currentSMBBlock *SMBStats
+	var currentSMBMetrics map[string]uint64
+	var currentSMBSessionIDs *SessionIDs
+	for scanner.Scan() {
+		line := scanner.Text()
+		// if line is empty we can go back to start
+		if line == "" {
+			continue
+		}
+		parseHeader(line, stats.Header)
+		// If we see a new SMB block we are initializing all necessary structs and hashmaps
+		if strings.Contains(line, ") \\") {
+			currentSMBMetrics = make(map[string]uint64)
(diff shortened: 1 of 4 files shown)


cifs/parse_cifs.go
```diff
@@ -0,0 +1,228 @@
...
+	"bufio"
+	"fmt"
+	"io"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+// ParseClientStats returns stats read from /proc/fs/cifs/Stats
+func ParseClientStats(r io.Reader) (*ClientStats, error) {
```
mdlayher: Please refactor this into at least four or five smaller functions.  There is too much context to keep track of here, and any indentation beyond 3 or so levels becomes unreadable.

cifs/parse_cifs.go
```diff
@@ -0,0 +1,228 @@
...
+							tmpSMB1Stats.SessionIDs.Server = tmpSessionIDs.Server
+							tmpSMB1Stats.SessionIDs.Share = tmpSessionIDs.Share
+
+						}
+						tmpSMB1Stats.Stats[name] = value
+					}
+				}
+				break
+			}
+		} else {
```
mdlayher: In my general experience, the use of "else" in Go means that you're not doing enough early returns.

cifs/parse_cifs.go
```diff
@@ -0,0 +1,228 @@
...
+	var tmpSessionIDs *SessionIDs
+	// The legacy variable sets the current context. True for SMB1, False for SMB2
+	legacy := true
+	for scanner.Scan() {
+		line := scanner.Text()
+		if legacy {
+			// This part manages the parsing of SMB1
+			for _, regexpSMB1 := range regexpSMB1s {
+				match := regexpSMB1.FindStringSubmatch(line)
+				if 0 == len(match) {
```
mdlayher: I understand this is subjective, but I really am not a fan of comparisons with a constant on the left side as opposed to the right.  It is quite uncommon in Go.

shibumi:
@mdlayher Sure. I am full with you about the part about readability. About the regex part: I don't think there is an easier way for this. You can have a look at the example files in the issue: https://github.com/prometheus/procfs/issues/95 Any other idea is more than welcome :)

cifs/parse_cifs.go
shibumi: sure absolutly!

shibumi:
I will overwork this and open a new PR, when it's ready.

mdlayher:
No need to close it, just following up with more commits is totally fine!

shibumi:
@mdlayher I have finally some time to code further on it. So I have reopen the pull Request and did  further research. I've found another solution for getting CIFS statistics. The `cifsiostat` command provides the following output:

```
Linux 3.10.0-862.3.3.el7.x86_64 (SERVER) 	14.09.2018 	_x86_64_	(16 CPU)

Filesystem:                    rB/s         wB/s    rops/s    wops/s         fo/s         fc/s         fd/s
\\SERVER\SHARE         0,00         0,00      0,00      0,00         0,00         0,00         0,00 
```

This is much less output as we get via `/proc/fs/cifs/Stats`:
```
Resources in use
CIFS Session: 1
Share (unique mount targets): 1
SMB Request/Response Buffer: 1 Pool size: 5
SMB Small Req/Resp Buffer: 1 Pool size: 30
Operations (MIDs): 0

0 session 0 share reconnects
Total vfs operations: 85 maximum at one time: 2

1) \\SERVER\SHARE
SMBs: 81 Oplocks breaks: 0
Reads:  0 Bytes: 0
Writes: 0 Bytes: 0
Flushes: 0
Locks: 0 HardLinks: 0 Symlinks: 0
Opens: 0 Closes: 0 Deletes: 0
Posix Opens: 0 Posix Mkdirs: 0
Mkdirs: 0 Rmdirs: 0
Renames: 0 T2 Renames 0
FindFirst: 0 FNext 0 FClose 0
```

So I would prefer for writing a parser for `/proc/fs/cifs/Stats` (although it's far more complicated with a lot of context switches and other nasty things. I can't understand that there is no easy machine readable output in the kernel at the moment)

shibumi:
@mdlayher @SuperQ Can one of you have a look at the newest commit? If this commit is okay I would rebase, squash and commit with sign-off message.

I am unhappy with two snippets at the moment:

* I need to move the flush regex line for SMB2 very early in the Regex Array otherwise we will have a mismatch and we will pollute our SMB2 statistics with a SMB1 flush statistic
* I need to check for a null pointer before calling `parseSMBStats`
* I am using the currentSMBSessionIDs variable just as check for a nullpointer.. I am pretty sure there is a more convenient way to do this.

shibumi:
@mdlayher @SuperQ Could you review again? It builds fine now, but it still uses the regex approach to get the statistics. A better solution would be to write an own parser for cifs statistics (maybe in a seperate project?). However this approach works fine, feedback to my code is welcome. I am still a golang newbie and need to learn a lot.

cifs/parse_cifs.go
```diff
@@ -0,0 +1,124 @@
...
+	"bufio"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+)
+
+// parseHeader parses our SMB header
+func parseHeader(line string, header map[string]uint64) error {
+	for _, regexpHeader := range regexpHeaders {
```
discordianfish: Can we use a single multline regex instead of this array? I think it would make the code be more readable.

cifs/parse_cifs_test.go
```diff
@@ -0,0 +1,362 @@
...
+Writes: 0 Bytes: 0
+Flushes: 0
+Locks: 0 HardLinks: 0 Symlinks: 0
+Opens: 0 Closes: 0 Deletes: 0
+Posix Opens: 0 Posix Mkdirs: 0
+Mkdirs: 0 Rmdirs: 0
+Renames: 0 T2 Renames 0
+FindFirst: 1 FNext 0 FClose 0
+
+2) \\server2\share2
```
discordianfish: Can you use different numbers in the second fixture? I think there might be an issue with the parsing.

discordianfish:
Sorry for not responding earlier! So this looks really hard to parse.. :-/

I don't think the loop over the file is doing the right thing. After it finds the 'new smb block', it always run the parse block function. Either way, I'd suggest you first try to split the file into the 'global' and per mount blocks and parse these with big multiline regex. I think that would make the code much more readable.

discordianfish:
Try using different numbers in the different fixtures, that should expose the issue. Or I'm wrong, quite possible :)
One giant regexp for the whole file probably doesn't work because the sections are repeating but dunno, maybe there is a way.

shibumi:
Ok I tried using different numbers in the second fixture and you are correct. The tests will fail, but I have no idea why. When I introspect the two stats `want` and `have` they are the same. I compared them. I guess you are talking about the `stats` hashmap for the second fixture, correct? They have the same values. I will upload a new version with new values in the second fixtures.

About parsing it as one file:
Yes, it's correct. The sections **can** repeat. So you can either have only SMB1 blocks or only SMB2 blocks or a mixed file with SMB1 and SMB2 blocks in no particular order. It's also possible to have 20 SMB2 blocks and just 2 SMB1 blocks after the 5th SMB2 block.. 

this makes parsing of the cifsstats really difficult. It's strange that no kernel developer tried changing this already. I mean human readable output is nice, but it would  be nice for us to have more machine readable output (the NFS stats are awesome for example)

discordianfish:
I'm going to close this for now. Let's re-open when the changes are addressed.

PR #105, opened 2018-08-18
Initial support for parsing /proc/net/arp

skottler:
This introduces ARP support for this package. Ultimately this will be used to keep track of more complex neighbor state for IPv6 support, but in the meantime should lay the foundation for removing `/proc` parsing logic from `collector/arp_linux.go` in `node_exporter`.

/cc https://github.com/prometheus/node_exporter/pull/540 where this was originally discussed
/cc https://github.com/prometheus/node_exporter/issues/535

### DIFF
--- a/arp.go
+++ b/arp.go
@@ -75,7 +75,6 @@ func parseARPEntries(file io.Reader) ([]ARPEntry, error) {
 }
 
 func parseARPEntry(columns []string) (ARPEntry, error) {
-
 	ip := net.ParseIP(columns[0])
 	mac := net.HardwareAddr(columns[3])
 
--- a/arp.go
+++ b/arp.go
@@ -53,14 +53,21 @@ func parseARPEntries(file io.Reader) ([]ARPEntry, error) {
 
 	for scanner.Scan() {
 		columns := strings.Fields(scanner.Text())
+		width := len(columns)
+		expectedDataWidth := 6
+		expectedHeaderWidth := 9
 
-		if columns[0] != "IP" {
+		if width == expectedHeaderWidth {
+			break
+		} else if width == expectedDataWidth {
 			entry, err := parseARPEntry(columns)
 			if err != nil {
 				return []ARPEntry{}, fmt.Errorf("failed to parse ARP entry: %s", err)
 			}
 
 			entries = append(entries, entry)
+		} else {
+			return []ARPEntry{}, fmt.Errorf("%d columns were detected, but %d were expected", width, expectedDataWidth)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


arp.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2018 The Prometheus Authors
+// Copyright 2018 Sam Kottler <shkottler@gmail.com>
+
```
grobie: Please keep the license as one contiguous block and prefix this empty line with `//`.

arp.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"strings"
+)
+
+type ARPEntry struct {
```
grobie: Needs a documentation.

arp.go
```diff
@@ -0,0 +1,86 @@
...
+	// IP address
+	IPAddr net.IP
+	// MAC address
+	HWAddr net.HardwareAddr
+	// Name of the device
+	Device string
+}
+
+func (fs FS) GatherARPEntries() ([]ARPEntry, error) {
```
grobie: Needs a documentation (which seems to have been pushed down to the next function).

arp.go
```diff
@@ -0,0 +1,86 @@
...
+			entry, err := parseARPEntry(columns)
+			if err != nil {
+				return []ARPEntry{}, fmt.Errorf("Failed to parse ARP entry: %s", err)
+			}
+
+			entries = append(entries, entry)
+		}
+	}
+
+	return entries, nil
```
grobie: You need to check for errors returned by the scanner. Can be easily done by changing this line to `return entries, scanner.Err()`.

arp.go
```diff
@@ -0,0 +1,86 @@
...
+// GatherARPEntries retrieves all the ARP entries, parse the relevant columns,
+// and then return a slice of ARPEntry's.
+func parseARPEntries(file io.Reader) ([]ARPEntry, error) {
+	scanner := bufio.NewScanner(file)
+	entries := make([]ARPEntry, 0)
+
+	for scanner.Scan() {
+		columns := strings.Fields(scanner.Text())
+
+		if columns[0] != "IP" {
```
grobie: This will panic in case `scanner.Text()` returned an empty string. You need to check that `columns` has the expected length first, and return an error otherwise.

arp.go
```diff
@@ -0,0 +1,86 @@
...
+			entry, err := parseARPEntry(columns)
+			if err != nil {
+				return []ARPEntry{}, fmt.Errorf("Failed to parse ARP entry: %s", err)
```
grobie: Error strings should not be capitalized https://github.com/golang/go/wiki/CodeReviewComments#error-strings

discordianfish:
@grobie Can we get this merged?

pgier:
@skottler would you mind rebasing this on the latest master and squashing the commits?

skottler:
I don't plan on finishing this so I'm going to close the pull request. Would love it if someone else could pick things up and get them across the finish line, but I don't have time right now.

discordianfish:
@skottler Understandable and sorry again for leaving this linger around for so long.

PR #106, opened 2018-08-21
Add VulnerabilityStats() to for kernel vulnerabilities

bobrik:
See https://github.com/prometheus/node_exporter/issues/1046


mjtrangoni:
@bobrik @discordianfish shouldn‘t it be 0 or 1 depending if this is mitigated (mitigation) or not (vulnerable).

bobrik:
@mjtrangoni, it's not that simple:

```
$ cat /sys/devices/system/cpu/vulnerabilities/l1tf
Mitigation: PTE Inversion; VMX: conditional cache flushes, SMT vulnerable
```

I think raw value is a good tradeoff: you want to know what kind of setups you have and possibly alert if mitigation is not what you expect it to be (think slow kernel fix vs fast microcode).

grobie:
Thanks a lot for your contribution!

The library aims to provide a simple interface towards the procfs/sysfs filesystems. The provided interface should mirror the naming of these filesystems, and take away the burden from go developers to parse the content on their own.

The function name should mirror the naming of file system names wherever possible. The full path is quite long, I propose `CPUVulnerabilities() ([]CPUVulnerability, error)`.

Furthermore, the intent of this library is to take away the burden of parsing the specific file formats and providing idiomatic golang datatypes instead. Therefore, I suggest to return a struct for each vulnerability instead of a map of strings which is not a clear interface.

```go
type CPUVulnerability struct {
	CodeName   string
	State      string
	Mitigation string
}
```

Following the sysfs documentation at https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu, `CodeName` would be the filename, `State` one of `Not Affected`, `Vulnerable`, or `Mitigated`, and `Mitigation` the value of `$M`. It might even make sense to define a type `CPUVulnerabilityState` and three constants for the three states to guarantee type safety for the state comparison.

bobrik:
I've updated the code. Let me know if that's what you had in mind and I'll work on docs and tests.

SuperQ:
Looks like the tests failed due to missing copyright headers.

pgier:
@bobrik Any chance you can rebase this one on latest master and add the copyright header?

discordianfish:
I've created a new PR with these changes rebased on master.

PR #109, opened 2018-09-19
Support Proc.Cwd() and Proc.RootDir()

pas2k:
Addressing @grobie, as per contributing guidelines for trivial improvement.

The PR implements reading /proc/pid/cwd and /proc/pid/root similar to how /proc/pid/exe is implemented.

Naming is inspired by https://talks.golang.org/2014/names.slide and golang standard library.

### DIFF
--- a/proc.go
+++ b/proc.go
@@ -156,6 +156,26 @@ func (p Proc) Executable() (string, error) {
 	return exe, err
 }
 
+// Cwd returns the absolute path to the current working directory of the process.
+func (p Proc) Cwd() (string, error) {
+	wd, err := os.Readlink(p.path("cwd"))
+	if os.IsNotExist(err) {
+		return "", nil
+	}
+
+	return wd, err
+}
+
+// RootDir returns the absolute path to the process's root directory (as set by chroot)
+func (p Proc) RootDir() (string, error) {
+	rdir, err := os.Readlink(p.path("root"))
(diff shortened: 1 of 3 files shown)


proc.go
```diff
@@ -156,6 +156,26 @@ func (p Proc) Executable() (string, error) {
 	return exe, err
 }
 
+// Cwd returns the absolute path to the current working directory of the process.
+func (p Proc) Cwd() (string, error) {
+	exe, err := os.Readlink(p.path("cwd"))
```
grobie: Please rename the variable to `cwd` or at least a generic `v`.

proc.go
```diff
@@ -156,6 +156,26 @@ func (p Proc) Executable() (string, error) {
...
+	if os.IsNotExist(err) {
+		return "", nil
+	}
+
+	return exe, err
+}
+
+// RootDir returns the absolute path to the process's root directory (as set by chroot)
+func (p Proc) RootDir() (string, error) {
+	exe, err := os.Readlink(p.path("root"))
```
grobie: Please rename the variable to `root` or at least a generic `v`.

pas2k:
Got it. Nice fixture system, by the way.

Since the introduced functions only operate on symlinks, the returned data should be as reliable as the one returned by os.Readlink(). Tests are included for the missing, broken and valid symlinks.

PR #110, opened 2018-09-27
add max open files and current open files information to NewStat()

peterloeffler:
We want to add information about max open files and current open files to the node_exporter.
There is already a pull request for that functionality: https://github.com/prometheus/node_exporter/pull/1069
To do it as efficient as possible I think it would be a good idea to do the file parsing here...


peterloeffler:
Ok, I got it.
Can also call this functions from outside the library.
In this case everything I'm doing here makes absolutely no sense.
Let's close this stupid pull request.

Issue #111, opened 2018-10-04
Build failure on windows after commit 6bfc2c7


okleitsa:
Build output:

vendor\github.com\prometheus\procfs\internal\util\parse.go:79:28: cannot use int(f.Fd()) (type int) as type syscall.Handle in argument to syscall.Read

jonnenauha:
Please fix this asap. These are no tags or "stable" branches in this repo. This is breaking builds all over the place that include prometheus golang client. Not having tags will force people to disable windows builds or vendor the whole repo just to remove it when this is fixes.

Maybe should use some other branch than master to do dev and run tests.

okleitsa:
I'm importing client_golang and using dep (still). This means that procfs is a transitive dependency to my project and thus will only get the required packages. So client_golang requires procfs package which in turns requires xfs and nfs packages which, finally, require the util package.

So, moving this function (which is OS-specific if I understand correctly) to the util package, makes every attempt to build to another OS unsuccessful.

I hope that was helpful.

SuperQ:
I would appreciate some help from some more experienced Windows developers.

One thing I'm thinking about is how does the [processCollector](https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go) work on Windows?

Does this bit of code work at all? Perhaps we can fix the dependency tree there, so that on windows it doesn't use any Linux-specific stuff like procfs.

jonnenauha:
@SuperQ as described, this seems to break all builds that fetch latest prometheus go client.

If you don't yet know how to fix it properly, please revert the commit. Do the change later when you figure it out.

Do this to fix this now:
```
git revert 6bfc2c70c4ee6da6946888bb96dec33f825df32b
git push origin/master
```

When you try again locally, just run `GOOS=windows go build` in this repos folder to test if the bug is fixes, you will get what was reported in the first message. You don't need access to windows to figure this out.

beorn7:
> One thing I'm thinking about is how does the processCollector work on Windows?

It doesn't report metrics, but an error. The code in question is https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go#L128-L135 .

Thus, it depends on the `NewStat` constructor to return an error on platforms that aren't covered by prometheus/procfs. The latter should have all the build guards to not try anything on (e.g.) MS Windows that wouldn't compile.

beorn7:
[...]

PR #112, opened 2018-10-05
Test on various Go OS versions.

SuperQ:
Setup a test matrix to make sure we avoid platform issues.

Closes: https://github.com/prometheus/procfs/issues/111


brian-brazil:
Hmm, that doesn't seem to be taking effect

SuperQ:
Yea, not sure why the env isn't getting used.

SuperQ:
@brian-brazil Ok, I had to do some refactoring of the build pipeline.

SuperQ:
@beorn7 Yea, promu is focused around building/testing binaries, not libraries. I think this works well enough for now.

Issue #107, opened 2018-09-01
Support for /proc/[pid]/fdinfo/[fd]


pirxthepilot:
Hello,

Any plans to support reading the contents of `/proc/[pid]/fdinfo/[fd]`? If not, but you think it's a good idea, I can try and submit a PR for it.

Thanks!

pirxthepilot:
In my case I would like to keep track of the number of inotify watches. From the [proc man page](http://man7.org/linux/man-pages/man5/proc.5.html):

```
              For inotify file descriptors (see inotify(7)), we see (since
              Linux 3.8) the following fields:

                  pos: 0
                  flags:    00
                  mnt_id:   11
                  inotify wd:2 ino:7ef82a sdev:800001 mask:800afff ignored_mask:0 fhandle-bytes:8 fhandle-type:1 f_handle:2af87e00220ffd73
                  inotify wd:1 ino:192627 sdev:800001 mask:800afff ignored_mask:0 fhandle-bytes:8 fhandle-type:1 f_handle:27261900802dfd73

              Each of the lines beginning with "inotify" displays informa‐
              tion about one file or directory that is being monitored.  The
              fields in this line are as follows:

              wd     A watch descriptor number (in decimal).

              ino    The inode number of the target file (in hexadecimal).

              sdev   The ID of the device where the target file resides (in
                     hexadecimal).

              mask   The mask of events being monitored for the target file
                     (in hexadecimal).
```

grobie:
@pirxthepilot I'd be happy to reasonable PR to add this functionality.

pgier:
Fixed in #115

PR #115, opened 2018-10-16
Parse fdinfo and inotify stats from /proc/[pid]/fdinfo/[fd]

pirxthepilot:
First attempt at adding functionality to support #107. Not quite sure if my approach here is correct - guidance/feedback would be very much appreciated. Also thought I'd let you guys review first before I write the tests.

Thanks!


inotify.go
```diff
@@ -0,0 +1,97 @@
...
+
+package procfs
+
+import (
+	"bufio"
+	"os"
+	"regexp"
+	"strings"
+)
+
```
pgier: I think it would be good to create an additional struct here called `FDInfo` which contains pos, flags, and mnt_id since those fields appear to be consistently available.  The `FDInfo` struct could contain a slice of `InotifyStat` containing the detailed information for inotify file descriptors.
And to be more consistent with the rest of the API, the file should be called `proc_fdinfo.go`.

inotify.go
```diff
@@ -0,0 +1,97 @@
...
+// InotifyStat represents a file descriptor's inotify watch count.
+type InotifyStat struct {
+	// File descriptor
+	FD string
```
pgier: This could be moved into the `FDInfo` struct.

inotify.go
```diff
@@ -0,0 +1,97 @@
...
+	// List of inotify lines in the fdinfo file
+	Lines []string
```
pgier: Instead of just loading the line here, I think it would be better to parse the values into their expected fields/types.  Or at least add the values to map for easy lookup.

proc.go
```diff
@@ -256,3 +256,18 @@ func (p Proc) fileDescriptors() ([]string, error) {
 func (p Proc) path(pa ...string) string {
 	return p.fs.Path(append([]string{strconv.Itoa(p.PID)}, pa...)...)
 }
+
+func (p Proc) fileDescriptorInfo(fd string) ([]byte, error) {
```
pgier: This needs to capitalized to be available from outside of the procfs package.  Also, I think this should return a `FDInfo` with the data already parsed, to make it easier to use.

discordianfish:
Agree with pgier and don't have more to add. And sorry for not responding earlier!

pirxthepilot:
Hi folks, just pushed some updates based on your feedback. I replaced `fileDescriptorInfo` with `FileDescriptorsInfo`, which now returns a slice of `FDInfo` types. Also modified and moved `InotifyWatchLen` to `proc.go` as it felt more correct to me.

I'm sure this still needs some tweaks. Also, my go-foo isn't too good to begin with :) Hope it at least makes some sense!

pgier:
@pirxthepilot Looks good, but would you mind rebasing on latest master?  I think the staticcheck CI failures are related to an outdated Makefile or some other build config.

pirxthepilot:
@pgier whoops, sorry bout that! done!

proc_fdinfo.go
```diff
@@ -0,0 +1,110 @@
...
+	// File access mode and status flags
+	Flags string
+	// Mount point ID
+	MntID string
+	// List of inotify lines (structed) in the fdinfo file (kernel 3.8+ only)
+	InotifyInfos []InotifyInfo
+}
+
+// FDInfo constructor. On kernels older than 3.8, InotifyInfos will always be empty.
+func (p Proc) newFDInfo(fd string) (*FDInfo, error) {
```
pgier: Can you change this to just `FDInfo`?  We did some refactoring recently to remove `new` from the functions where it's not needed.

pirxthepilot: Will try to have this done over the weekend. Do you mean just outright rename this function to `FDInfo`, i.e. 
```
func (p Proc) FDInfo(fd string) ...
```
And should I do the same for `newInotifyInfo` (further down this file)?

Thanks!

pgier: Right, just rename to `func (p Proc) FDInfo(fd string)`.  You can take a look at some of the other `proc_<thing>.go` files for example.  
I'm thinking `newInotifyInfo` should be `parseInotifyInfo` instead.  It can't be just `InotifyInfo` because then it would conflict with the struct name.

pgier:
@pirxthepilot I think this still needs unit tests before we merge.

pirxthepilot:
@pgier sorry for the delayed response. I pushed a new commit with the requested changes. Just want to run it by you as I start writing the tests.

As for the tests, how should I go about adding fixtures? Do I just append data by hand in `fixtures.ttar`?

proc.go
```diff
@@ -265,3 +265,37 @@ func (p Proc) fileDescriptors() ([]string, error) {
...
+		fdinfo, _ := p.FDInfo(n)
+		fdinfos[i] = *fdinfo
+	}
+
+	return fdinfos, nil
+}
+
+// InotifyWatchLen returns the total number of inotify watches used
+// by the process.
+func (p Proc) InotifyWatchLen() (int, error) {
```
pgier: It might be better to make this a regular function which takes a slice of `ProcFDInfo` as an argument, instead of a method of Proc.  Since the process is not used other than to get the FDInfos.  Or you could define a new type which is a `[]ProcFDInfo` and make this a method of that type.

proc_fdinfo.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2018 The Prometheus Authors
```
pgier: Should be 2019

pirxthepilot:
Changes made and tests written. I think the test in `proc_test.go` should provide coverage for the functions there, except for `InotifyWatchLen`, which I wrote a separate test for in `proc_fdinfo_test.go`. Let me know if you need anything else.

proc.go
```diff
@@ -279,3 +279,21 @@ func (p Proc) fileDescriptors() ([]string, error) {
...
+	names, err := p.fileDescriptors()
+	if err != nil {
+		return nil, err
+	}
+
+	fdinfos := make(ProcFDInfos, len(names))
+
+	for i, n := range names {
+		fdinfo, _ := p.FDInfo(n)
+		fdinfos[i] = *fdinfo
```
pgier: If `p.FDInfo` returns an error and a nil fdinfo won't this cause a nil pointer dereference?  It seems like this could happen for example if a file is closed while iterating through the files.  I think it would be better to check for an error and then use `append` to add to the slice instead of using the index.

Also, if you make `ProcFDInfos` a slice of pointers instead of a slice of structs, then it should be a bit more efficient because the full struct don't need to be copied with each assignment.

fixtures.ttar
```diff
@@ -2079,70 +2147,78 @@ Mode: 755
...
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1
+Directory: fixtures/sys/kernel/config/target/core/fileio_1
 Mode: 755
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/enable
+Directory: fixtures/sys/kernel/config/target/core/fileio_1/file_lio_1G
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/core/fileio_1/file_lio_1G/enable
```
pgier: There seem to be a lot of changes to the fixtures that are unrelated to this PR. You might need to rebase on latest master and then update the fixtures again.

proc_fdinfo.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read %s: %s", f.Name(), err)
+	}
+
+	var text, pos, flags, mntid string
+	var inotify []InotifyInfo
+	rPos := regexp.MustCompile(`^pos:\s+(\d+)$`)
+	rFlags := regexp.MustCompile(`^flags:\s+(\d+)$`)
+	rMntID := regexp.MustCompile(`^mnt_id:\s+(\d+)$`)
+	rInotify := regexp.MustCompile(`^inotify`)
```
pgier: These regexps should be moved to top level package variables so that they don't need to be compiled for each function call.

pgier:
I think the fixtures issue is a separate issue and should be fixed by PR #185, so you can ignore that part.

proc_fdinfo.go
pirxthepilot: Sorry, I'm not yet familiar on how this is set. Do I just declare these at the top of the same file, after the imports?

pgier: Yes, you can put them in a [`var ()` block](https://golang.org/doc/effective_go.html#variables) outside the function.  The location is mostly style preference, but either after the imports or just before the function where they are used are fine.

pirxthepilot:
Updated!

pgier:
Ug, sorry, the commit message lost the signoff when I did a squash merge.  I had to amend the commit and do a force push in #189.

PR #117, opened 2018-10-20
Add Support for /sys/class/power_supply

shaardie:
This corresponds to [ACPI Support Issue](https://github.com/prometheus/node_exporter/issues/676) in Node exporter.

The implementation uses [netclass.go](https://github.com/prometheus/procfs/blob/master/sysfs/net_class.go) as a model.


SuperQ:
Thanks! I have two overall review items.

I think the file should be named `class_power_supply.go` to match the `/sys/class/power_supply` hierarchy.

Please add a tests like net_class_test.go and system_cpu_test.go.

sysfs/power_supply_class.go
```diff
@@ -0,0 +1,188 @@
...
+	"io/ioutil"
+	"os"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// PowerSupply contains info from files in /sys/class/power_supply for a single power supply
```
SuperQ: Please end all comments with a full stop (`.`).

shachaf:
As far as I can tell all the information in the individual files this reads is also available in `/sys/class/power_supply/*/uevent`, which is much more straightforward to read and parse. It might be worth considering doing that instead? I linked to some code that does that in a comment on the issue.

shaardie:
@shachaf Yes, all those attributes are also present via the `uevent` interface in `/sys/class/power_supply/*/uevent`, but 

* I thought this project was about parsing the `sysfs` and `procfs` interfaces, so it appears canonical to me to read those interfaces.
* I do not really see why parsing the `uevent` file is more straightforward. Yes it is only one file, but you actually have to parse it. Using the other files is only reading the file and using the correct type.

PR #119, opened 2018-11-29
*: support Go modules

simonpasquier:


.circleci/config.yml
```diff
@@ -1,49 +1,48 @@
...
+        name: linux-1-10
+        os: linux
+        go_version: "1.10"
+    - test:
+        name: linux-1-11
+        os: linux
+        go_version: "1.11"
+    - test:
+        name: windows-1-10
+        os: windows
```
grobie: Why do we run procfs tests on Windows? That OS isn't supported. When was this introduced?

SuperQ: Due to package requirements, client_golang requires procfs. There's basically one piece of code in client_golang that causes all of procfs to get pulled in as a requirement. I've complained about this several times, but no progress has been made to clean it up.

https://github.com/prometheus/procfs/pull/112

simonpasquier:
Configuration is fixed now (stupid typo eating the environment variable).

.circleci/config.yml
grobie: Well, that part of client_golang was the reason I created the procfs repo :-) The reason that we pull in even sysfs and all the ne packages created in the procfs repo is something I would like to change. There is no need that we couple all these packages together.

Thanks for the link to the pr.

PR #120, opened 2018-11-30
override default "build" target to fix makefile

pgier:
This fixes the default "make" and "make build" commands failing due to
lack of a .promu.yml file.  Overrides the "build" target with
basic "go build".

Signed-off-by: Paul Gier <pgier@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,5 +23,8 @@ update_fixtures: fixtures.ttar sysfs/fixtures.ttar
 	rm -v $(dir $*)fixtures/.unpacked
 	./ttar -C $(dir $*) -c -f $*fixtures.ttar fixtures/
 
+.PHONY: build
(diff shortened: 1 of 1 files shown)


SuperQ:
Do we need a build step for procfs? There's no main here.

pgier:
It's not a big deal, but currently if you just run "make" the build fails because it tries to run "promu build" and doesn't find a .promu.yml.  An alternative would be to either make the "build" target to nothing, or override the "all" target so that it doesn't include build.

grobie:
Thanks for the catch! As go build is a noop without any main package, I'd make that explicit by overwriting `build` to do nothing as well.

pgier:
Updated the commit to make the build target a no-op instead of a basic go build.

PR #123, opened 2018-12-07
Collect both cpuinfo and scaling files

SuperQ:
Collect cpufreq `cpuinfo` and `scaling` files. They have different
meanings.
* Ignore not found and unreadable files.
* Return pointers for uint64 values.


SuperQ:
1.10 fails due to go fmt differnces.

PR #125, opened 2019-01-04
Collect cpufreq in parallel

SuperQ:
Use wait group to collect the cpufreq data in parallel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ commands:
     - run: make style check_license vet staticcheck
   no-style-test:
     steps:
-    - run: go get golang.org/x/sync/errgroup
+    - run: go get -t ./...
     - run: make check_license vet staticcheck
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,7 @@ commands:
     - run: make style check_license vet staticcheck
   no-style-test:
     steps:
+    - run: go get golang.org/x/sync/errgroup
     - run: make check_license vet staticcheck
 
 jobs:
--- a/sysfs/system_cpu.go
+++ b/sysfs/system_cpu.go
@@ -62,7 +62,7 @@ func (fs FS) NewSystemCpufreq() ([]SystemCPUCpufreqStats, error) {
 
 	cpus, err := filepath.Glob(fs.Path("devices/system/cpu/cpu[0-9]*"))
 	if err != nil {
-		return []SystemCPUCpufreqStats{}, err
+		return nil, err
 	}
 
 	systemCpufreq := make([]SystemCPUCpufreqStats, len(cpus))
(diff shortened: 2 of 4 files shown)


SuperQ:
This is currently causing a data race, I thought that it was OK to write to the indexed value, but it's not working the way I thought it would.

SuperQ:
In new versions, the cpufreq subsystem waits 50ms to let you read the file, this causes the node_exporter scrape times to be very long on large bare metal nodes.

.circleci/config.yml
```diff
@@ -13,6 +13,7 @@ commands:
     - run: make style check_license vet staticcheck
   no-style-test:
     steps:
+    - run: go get golang.org/x/sync/errgroup
```
juliusv: Hmm, I *think* this shouldn't be necessary when build, vet, and staticcheck run with Go module support enabled? E.g. https://github.com/prometheus/procfs/blob/master/Makefile.common#L129

But my Go module foo is weak.

sysfs/system_cpu.go
```diff
@@ -76,12 +77,19 @@ func (fs FS) NewSystemCpufreq() ([]SystemCPUCpufreqStats, error) {
...
+			if err == nil {
+				cpufreq.Name = cpuName
+				systemCpufreq[i] = *cpufreq
+			}
+			return err
+		})
+	}
+
+	if err = g.Wait(); err != nil {
+		return []SystemCPUCpufreqStats{}, err
```
juliusv: This can just be `return nil, err`

sysfs/system_cpu.go
```diff
@@ -76,12 +77,19 @@ func (fs FS) NewSystemCpufreq() ([]SystemCPUCpufreqStats, error) {
...
 		}
 
-		cpufreq, err = parseCpufreqCpuinfo(cpuCpufreqPath)
-		if err != nil {
-			return []SystemCPUCpufreqStats{}, err
-		}
-		cpufreq.Name = cpuNum
-		systemCpufreq = append(systemCpufreq, *cpufreq)
+		i := i // https://golang.org/doc/faq#closures_and_goroutines
+		g.Go(func() error {
```
juliusv: Could you add a comment somewhere to explain why we are doing this in parallel, because without the knowledge that the kernel intentionally puts in delays it seems like this wouldn't be needed.

.circleci/config.yml
simonpasquier: You're right regarding Go modules but this command is only called for Go 1.10.x. Otherwise I guess that you can simply do `go get -t ./...`.

sysfs/system_cpu.go
SuperQ: Sure.

.circleci/config.yml
juliusv: Ah, didn't see that context. But yeah, doing a generate `go get` is much better than hardcoding the dependencies here.

SuperQ: Yes, I added this because the Go 1.10 build was broken.

@simonpasquier I'll change this.

sysfs/system_cpu.go
SuperQ: Done.

SuperQ:
@juliusv Cleaned up, PTAL.

juliusv:
:+1:

PR #127, opened 2019-01-05
procfs: Add support for /proc/pressure/*

### LINKED ISSUE #126: Add pressure stall information
I'd like to add support for `/proc/pressure/*` that the Linux kernel exposes (to in turn add support for that to the node exporter).

I started looking at it but I'm stumbling on a thing; currently `Proc` and its `path` helper always assumes that what we want is `/proc/<pid>/<path...>` however what I need access to is `/proc/pressure/{cpu,memory,io}` and I can't seem to find a good way to do that.

Did I miss something in the `Proc` API or should I just read the contents of those three files without leveraging the `Proc` struct and its methods?

daenney:
Closes #126

### DIFF
--- a/proc_psi.go
+++ b/proc_psi.go
@@ -1,0 +1,110 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procfs
+
+// The PSI / pressure interface is described at
+//   https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/accounting/psi.txt
+// Each resource (cpu, io, memory, ...) is exposed as a single file.
+// Each file may contain up to two lines, one for "some" pressure and one for "full" pressure.
+// Each line contains several averages (over n seconds) and a total in µs.
+//
+// Example io pressure file:
+// > some avg10=0.06 avg60=0.21 avg300=0.99 total=8537362
+// > full avg10=0.00 avg60=0.13 avg300=0.96 total=8183134
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+const lineFormat = "avg10=%f avg60=%f avg300=%f total=%d"
+
+// PSILine is a single line of values as returned by /proc/pressure/*
+// The Avg entries are averages over n seconds, as a percentage
+// The Total line is in microseconds
+type PSILine struct {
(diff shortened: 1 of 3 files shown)


proc_psi.go
```diff
@@ -0,0 +1,120 @@
...
+			if err != nil {
+				parseErrors = append(parseErrors, err.Error())
+				continue
+			}
+			psiStats.Full = &psi
+		default:
+			// If we encounter a line with an unknown prefix, ignore it and move on
+			// Should new measurement types be added in the future we'll simply ignore them instead
+			// of erroring on retrieval
+			continue
```
daenney: I'm not a 100% sure if this is a good idea, but it seems saner than throwing a parse error on a line we can predict we can't make sense of.

proc_psi.go
```diff
@@ -0,0 +1,120 @@
...
+
+// parsePSIStats parses the specified file for pressure stall information
+func parsePSIStats(resource string, file io.Reader) (PSIStats, error) {
+	psiStats := PSIStats{}
+	stats, err := ioutil.ReadAll(file)
+	if err != nil {
+		return psiStats, fmt.Errorf("psi_stats: unable to read data for %s", resource)
+	}
+
+	parseErrors := []string{}
```
daenney: See the code in the loop. The idea I had was to not error out when parsing a single line but instead continuing with the next one, accumulating errors and then return that as a whole. So if we ran into a parse error on the 'some' line the returned object would at least still have `stats.Full` if we managed to parse the 'full' line.

I'm not sure if this is desirable, since we do end up returning an error so effectively the caller should disregard the `PSIStats` returned.

daenney:
There is a few error conditions this code allows for but shouldn't occur in reality:

* `PSIStatsForResource("cpu")` should never have a non-nil `.Full` according to the kernel docs as a 'full' line is never reported for it. Could handle that but if that were to happen we would effectively have a kernel bug
* `PSIStatsForResource()` should never return with a nil `.Some` but a non-nil `.Full` according to the kernel docs b/c there should always be at least a 'some' line for each resource. This again isn't handled here b/c that would be a kernel bug

daenney:
@SuperQ Any chance I can trouble you for a review? I'd love to get feedback on the open questions and hopefully merge this so I can get the counterpart in the node-exporter going.

SuperQ:
Yes, sorry, this fell off my review radar. Thanks for the ping.

proc_psi.go
SuperQ: Typically we want to error early and return the error rather than try and return partial results. The exception to this is where we have optional files/fields. For example, in NFS there are different stats returned by different versions. But if we fail to parse the oldest supported stats, we return nothing and error.

daenney:
No worries, things take time. Just wanted to ensure someone would eventually get to it.

proc_psi.go
discordianfish: +1 on @SuperQ, let's error early.

proc_psi.go
discordianfish: Yes, I think that's reasonable. In that case this says compatible if a new prefix is introduced.

discordianfish:
Looking forward to this. Let's just return early on error to make the result consistent and error handling a bit simpler and we can merge this.

daenney:
Alright! I'll have a go at that either today or Monday

daenney:
@SuperQ @discordianfish All yours!

PR #130, opened 2019-01-14
Parse mountaddr from mountstats

mknapphrt:
This PR will export the mountaddr from the opts field. The test case is a little weird, I wasn't sure how much of the opts field to add considering only 1 value is being parsed out of it, but I wanted to have some other fields to prove that it properly grabbed the right field.

 I have a use case for this with: https://github.com/prometheus/node_exporter/pull/994 

Signed-off-by: Mark Knapp <mknapp@hudson-trading.com>

### DIFF
--- a/mountstats.go
+++ b/mountstats.go
@@ -69,6 +69,8 @@ type MountStats interface {
 type MountStatsNFS struct {
 	// The version of statistics provided.
 	StatVersion string
+	// The optional mountaddr of the NFS mount.
+	MountAddress string
 	// The age of the NFS mount.
 	Age time.Duration
 	// Statistics related to byte counters for various operations.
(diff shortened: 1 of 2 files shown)


mountstats.go
```diff
@@ -69,6 +69,8 @@ type MountStats interface {
 type MountStatsNFS struct {
 	// The version of statistics provided.
 	StatVersion string
+	// The mount address of the NFS mount.
```
SuperQ: Minor nit, I would make this comment `The optional mountaddr of the NFS mount.`

mknapphrt:
I'm not sure sure this will work anymore. If this is an optional field then can it be used as a label for the mountstats? I was hoping to use it for the deduplication problem in the linked issue but I would need some kind of default value when the mountaddr isn't there. Which seems like bad practice, especially for a field that isn't just a number that could default to 0.

mknapphrt:
Right, I was mostly worried about having an empty string for a label, whereas an int value would default  to "0" at least. But now that I think about it more adding a label wouldn't introduce any more duplication possibilities than already exist, it can only serve to distinguish two mounts that are already identical. Cool, I'll make fix that nit and squash the commits. Thanks!

mknapphrt:
@discordianfish Gentle nudge :)

mknapphrt:
@discordianfish Slightly less gentle nudge :)

PR #131, opened 2019-01-15
ProcStat VSize is unsigned

flixr:
see `man 5 proc`

This should fix the integer overflow of this field on 32bit platforms.

See also https://github.com/ncabatoff/process-exporter/issues/80

### DIFF
--- a/proc_stat_test.go
+++ b/proc_stat_test.go
@@ -38,7 +38,7 @@ func TestProcStat(t *testing.T) {
 		{name: "user time", want: 1677, have: int(s.UTime)},
 		{name: "system time", want: 44, have: int(s.STime)},
 		{name: "start time", want: 82375, have: int(s.Starttime)},
-		{name: "virtual memory size", want: 56274944, have: s.VSize},
+		{name: "virtual memory size", want: 56274944, have: int(s.VSize)},
 		{name: "resident set size", want: 1981, have: s.RSS},
 	} {
 		if test.want != test.have {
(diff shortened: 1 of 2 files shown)


SuperQ:
I didn't find where it says in the man page about being unsigned, but I think I found the [reference in the kernel source](https://github.com/torvalds/linux/blob/fe76fc6aaf538df27708ffa3e5d549a6c8e16142/fs/proc/array.c#L433).

I also wonder about the `RSS` field, if that should also be uint.

proc_stat_test.go
```diff
@@ -38,7 +38,7 @@ func TestProcStat(t *testing.T) {
 		{name: "user time", want: 1677, have: int(s.UTime)},
 		{name: "system time", want: 44, have: int(s.STime)},
 		{name: "start time", want: 82375, have: int(s.Starttime)},
-		{name: "virtual memory size", want: 56274944, have: s.VSize},
+		{name: "virtual memory size", want: 56274944, have: int(s.VSize)},
```
SuperQ: You should cast the literal here. `uint(56274944)`.

flixr: sure, I can change this... just did the same as for the other cases above....

flixr:
In the manpage it states the scanf format, there you can see that it is unsigned.

proc_stat_test.go
flixr: Using uint here makes the test compilation fail...
Removing the last commit again...

SuperQ: Hrm, My Go testing fu is failing I guess.

SuperQ:
Ohhhhh, Yes, my brain skipped over the `scanf` info.

baryluk:
FYI.

From manpage (i.e. http://man7.org/linux/man-pages/man5/proc.5.html ):

```

       /proc/[pid]/stat
              Status information about the process.  This is used by ps(1).  It is defined in the kernel source file fs/proc/array.c.


              (23) vsize  %lu
                        Virtual memory size in bytes.
```


%lu - long unsigned.

Unfortunately it is not mentioned at all in the Documentation/filesystems/proc.txt in kernel docs itself.

PR #138, opened 2019-02-28
Consolidate fixtures

### LINKED ISSUE #43: Consolidate fixtures structure
So the current fixtures structure doesn't make too much sense. It's mixing "root" level files/folders like `stat` with special purpose subdirectories (like `buddyinfo/short/buddinfo`). I was thinking about moving all "root" files into a `default` subdirectory so that we'd end up with `fixtures/{default,specialcasedir,likebuddyinfo}/actualpath`. Given we support now both procfs and sysfs, maybe it should become `fixtures/{proc,sys}/{default,special,otherspecial}/...`?

@mdlayher wdyt?

pgier:
fixes #43


pgier:
@SuperQ @mdlayher wdyt?

Issue #44, opened 2017-04-21
Consolidate method/function names


grobie:
The current naming is not stringent nor idiomatic.

From @mdlayher 

> My personal rules of thumb for each case:
>
> * NewThing: top level constructor functions. Never methods.
> * ParseThing: parses some input data like []byte or io.Reader.
> * Thing: retrieval methods for some data. Doesn't make much sense as a name for top-level functions to me.

It might make sense to just use the procfs names wherever possible.

mdlayher:
I'd also like to suggest that we get rid of constructor functions like `NewBuddyInfo` in favor of methods like: `FS.BuddyInfo`, because:
- it will shrink the API surface
- it will reduce redundancy
- it will enable much easier testing

grobie:
@mdlayher You mean all package level shorthand methods? Even `Self()`?

mdlayher:
Good question. I suppose it depends how much some of them are used.  While they may be convenient, I imagine that most applications will want to construct a `FS` instance anyway for testing that doesn't only work on Linux machines.

pgier:
Is this something that we still want to do?  After the refactoring in #149, we still have a lot of method names that look like `fs.NewSomething()`.  Using the BuddyInfo example, we have `NewBuddyInfo()` and `fs.NewBuddyInfo()`.  We could change the fs method to `fs.BuddyInfo()` however the top level function, can't be changed to just `BuddyInfo()` because then it conflicts with the BuddyInfo struct.  Maybe we can change top level functions to something like `ReadBuddyInfo()` instead?

discordianfish:
I think the idiomatic way would be to call it BuddyInfo() and rename the struct instead..

discordianfish:
See the linked PRs. I think `fs.BuddyInfo()` is the way to go and I tend to say we should call the package level functions that do the same should also be called the same but that's somewhat against prefixing "top level constructor functions" with New.. @mdlayher Would love your 2 cents on this.

mdlayher:
I still agree with my previous comment: https://github.com/prometheus/procfs/issues/44#issuecomment-296304689

I'd like to see zero package level functions, and everything is only accessible via a method on the FS type.

pgier:
[...]

PR #140, opened 2019-03-11
Adding parsing of /proc/<pid>/status

samber:
@grobie

Adding some useful data, such as context switches count, full memory usage...

### DIFF
--- a/proc_status.go
+++ b/proc_status.go
@@ -1,0 +1,162 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procfs
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// ProcStat provides status information about the process,
+// read from /proc/[pid]/stat.
+type ProcStatus struct {
+	// The process ID.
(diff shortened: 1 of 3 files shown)


proc_status.go
```diff
@@ -0,0 +1,158 @@
...
+		kv := strings.SplitN(line, ":", 2)
+
+		// removes spaces
+		k := string(strings.TrimSpace(kv[0]))
+		v := string(strings.TrimSpace(kv[1]))
+		// removes "kB"
+		v = string(bytes.Trim([]byte(v), " kB"))
+
+		// value to int when possible
+		vKBytes, _ := strconv.ParseUint(v, 10, 64)
```
discordianfish: Check error

proc_status.go
```diff
@@ -0,0 +1,158 @@
...
+		// convert kB to B
+		vBytes := vKBytes * 1024
+
+		if k == "Name" {
```
discordianfish: Use switch instead

proc_status.go
samber: Hi @discordianfish 

This error check is not necessary (see comment).

samber:
```
// we can skip error check here, 'cause vKBytes is not used when value is a string
```

proc_status.go
```diff
@@ -0,0 +1,164 @@
...
+	VmSwap uint64
+	// Size of hugetlb memory portions
+	HugetlbPages uint64
+
+	// Number of voluntary context switches.
+	VoluntaryCtxtSwitches uint64
+	// Number of involuntary context switches.
+	NonVoluntaryCtxtSwitches uint64
+
+	fs FS
```
pgier: Is this `fs` necessary?  I don't see it used anywhere.

samber:
Rebased ;)

pgier:
Great!  Could you also squash these commits with a descriptive commit log message?  Just want to avoid commit logs like "Oops".

samber:
@pgier done

pgier:
LGTM

discordianfish:
LGTM but again naming.. We should decide on #44 asap.

pgier:
Merged.  We can fix any naming issues related to #44 in a separate PR.

PR #142, opened 2019-03-13
net_class: Split device discovery

daenney:
This splits the discovery of device/interfaces out from the main
`NewNetClass()` function.

I want to do this b/c on systems that have a lot of devices in
`/sys/class/net/*` calling `parseNetClassIface` in serial takes a lot of
time. On my system with 55 devices it currently takes about 150ms.

By splitting the device discovery I can call `parseNetClassIface` using
a worker pool and process them much faster. `NewNetClass` is updated to
use the new `NetClassDevices()` method so its behaviour does not change.

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -19,12 +19,15 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
 
 	"github.com/prometheus/procfs/internal/util"
 )
+
+const netclassPath = "class/net"
 
 // NetClassIface contains info from files in /sys/class/net/<iface>
 // for single interface (iface).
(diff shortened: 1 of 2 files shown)


SuperQ:
Seems like a good idea to me. We did a similar split and used `golang.org/x/sync/errgroup` in `sysfs/system_cpu.go`.

sysfs/net_class.go
```diff
@@ -72,26 +74,42 @@ func NewNetClass() (NetClass, error) {
...
+func (fs FS) NewNetClass() (NetClass, error) {
+	devices, err := fs.NetClassDevices()
+	if err != nil {
+		return nil, err
+	}
+
+	path := fs.Path(netclassPath)
+	netClass := NetClass{}
+	for _, deviceDir := range devices {
+		interfaceClass, err := netClass.parseNetClassIface(path + "/" + deviceDir)
```
SuperQ: We should probably fix this to use `filepath.Join()`. Not asking you to do it, but just came to mind when reviewing.

daenney:
Follow-up question; with this landing should I update `NewNetClass()` to use an `errgroup` or do we want to keep it as-is here and implement that in the node exporter instead?

SuperQ:
I think we should implement the `errgroup` here, to keep the node_exporter's interface the same.

daenney:
Fixed to use `filepath.Join` now.

PR #143, opened 2019-03-14
Don't build sysreadfile for app engine

carlpett:
Google App Engine does not allow compiling with `import "syscall"`. See discussion in https://github.com/prometheus/client_golang/issues/549.

### DIFF
--- a/internal/util/sysreadfile_linux.go
+++ b/internal/util/sysreadfile_linux.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !windows
+// +build !windows !appengine
 
 package util
 


carlpett:
Actually, does it even need the `!windows` constraint? The file is named `_linux`, so it should already be excluded on Windows?

cc @discordianfish @grobie

SuperQ:
Hrm, I'm not sure about the `!windows` thing.

beorn7:
I believe all the files that use `util.SysReadFile` have to get they build tag updated accordingly. Otherwise, they will now in turn create a compile error on appengine.

SuperQ:
@beorn7 Do you mean all of the `package util` files?

SuperQ:
Oh, no, you're right. I get it.

SuperQ:
@carlpett Would you mind updating these files too:

```
sysfs/class_power_supply.go
sysfs/class_thermal.go
sysfs/net_class.go
sysfs/system_cpu.go
```

brian-brazil:
We could also put in a stub implementation that's only used for appengine.

carlpett:
I went with the stub implementation, see #144

PR #144, opened 2019-03-17
Better App Engine build tags fix

carlpett:
This fixes the patch from #143.
The "full implementation" is in `sysreadfile.go`, and is used for Linux (but not on App Engine), and `sysreadfile_compat.go` provides a fallback implementation for other platforms.

ping @SuperQ @beorn7 @brian-brazil.


internal/util/sysreadfile_compat.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ: New file, please set this to 2019.

Issue #145, opened 2019-03-22
/sys/class/power_supply/<power_supply>/scope is no uint64


shaardie:
The file is currently parsed as uint64, but is actually a string.

discordianfish:
Fixed by #146. Thanks!

PR #146, opened 2019-03-22
sysfs: change power supply scope to string

shaardie:
/sys/class/power_supply/<power_supply>/scope is currently parsed as
uint64, but is is actually a string, see #145.

Signed-off-by: Sven Haardiek <sven.haardiek@iotec-gmbh.de>


SuperQ:
It looks like we need to update the build Go versions.

SuperQ:
Ok, #147 is merged, please rebase.

Issue #135, opened 2019-02-20
Refactor interpackage dependencies to minimize client dependency tree


pgier:
@SuperQ brought up the issue that the prometheus client_golang project uses only a couple of functions from procfs for gathering process data, but the procfs package has dependencies on the xfs and nfs (and recently also iostats) packages, so these subpackages are pulled in to the client's dependency tree even if they are not used.  Ideally, any clients of procfs should only see minimal additions to their dependency tree.

Before looking at possible solutions, I think it's important to note that even with the current dependency tree, the go compiler seems to be smart enough to not actually bring in code from the xfs or nfs packages into the resulting binary if they are not used.

However, if we do decide to remove the dependency from procfs to the xfs and nfs packages, there are a few options.

(1) We could move the XFSStats method into the xfs package (and do something similar with the NFS methods).  However, since XFSStats is currently defined as a method of the FS type, it makes it a little more tricky to decouple it.  Since golang doesn't allow methods to be defined on a struct from another package, we'd have to either change it to a normal function and have a dependency on procfs, or duplicate the FS type and a couple other associated methods in the xfs package.

(2) Another option is to just merge the xfs and nfs packages into procfs.  This doesn't really reduce the client dependencies, but it does make the tree appear smaller and simplifies the procfs packages structure.  This solution would likely also require making the sysfs package dependent on the procfs package to use some of the common data structures and functions.

(3) We could do a larger package reorganization to make each package very small and specific.  We could do something like create a package structure similar to the path structure of the /proc and /sys filesystems.  We would probably also need a set of packages for common data structures which are used by both proc and sys.

pgier:
Fixed by #136

PR #148, opened 2019-04-03
Remove FS custom type

### LINKED ISSUE #44: Consolidate method/function names

pgier:
I think this also fixes #44


mdlayher:
Will review today, but it might not be for a few hours yet.

mdlayher:
I see where you're going with this, but I think I stand by my prior comments in that issue:

> I'd also like to suggest that we get rid of constructor functions like NewBuddyInfo in favor of methods like: FS.BuddyInfo ...

> While they may be convenient, I imagine that most applications will want to construct a FS instance anyway for testing that doesn't only work on Linux machines.

I'm all for getting rid of the top-level constructors that implicitly use the default proc/sys filesystem, but I think it makes more sense to keep the FS types around rather than this kind of unusual variadic defaults approach for every top level function.

pgier:
@mdlayher My reasoning for removing the FS type was try to match the signature of the functions in the refactored subpackages in #135.  I could add an FS type to each of those subpackages, but it seemed like unnecessary duplication.  I also considered using a normal string instead of the variadic arg, but that would mean the mount point is always a required parameter and we'd lose some backwards compatibility.

mdlayher:
What do you think about making these functions always take no arguments (default to `/proc`), and then having internal shims that can specify a path?

```go
func Foo() (*Bar, error) {
    return foo("/proc")
}

func foo(path string) (*Bar, error) {}
```

I feel like the variadic argument API could be confusing, especially because any arguments after the first are totally ignored.

So then: could /proc be mounted anywhere else? Is this something we use in node_exporter for example when running in a docker container?

If we still wanted to specify, I think I'd probably prefer empty string as "default".

pgier:
> What do you think about making these functions always take no arguments (default to /proc), and then having internal shims that can specify a path?

If we did it this way, would the client do something like set an exported variable such as "MountPoint" or add a "SetMountPoint" function to override the default path?  This might be confusing if we had a client that wanted to access two different proc filesystems.  For example, the in-container /proc and the mounted host's proc directory.

> So then: could /proc be mounted anywhere else? Is this something we use in node_exporter for example when running in a docker container?

Yes, in node_exporter it's common to use a non-default path for proc, because we want to access the host filesystem instead of /proc in the container.  Even though it's recommended to not use node_exporter in a container, it seems fairly common.  The common pattern in the collectors using the procfs API is to create a new FS each time.

```go
func (c *buddyinfoCollector) Update(ch chan<- prometheus.Metric) error {
    fs, err := procfs.NewFS(*procPath)
    if err != nil {
        return fmt.Errorf("failed to open procfs: %v", err)
    }

    buddyInfo, err := fs.NewBuddyInfo()
```
https://github.com/prometheus/node_exporter/blob/b8b0195d6d76358dfb592cf7e49781a443bf1849/collector/buddyinfo.go#L52-L58

Since a new FS is created for each collector during each update, I thought it made sense to just set the mount point directly in the function when reading the data.

In client_golang the default mount point is used to get process info instead of creating an FS instance.

```go
if _, err := procfs.NewStat(); err == nil {
```
https://github.com/prometheus/client_golang/blob/7490f0a74525a1f863eaf81b42f3ead3a1ecc43a/prometheus/process_collector.go#L129

So I was thinking it would be nice to be able to maintain this default behaviour, but still be able to override the mount point using the same function.

> If we still wanted to specify, I think I'd probably prefer empty string as "default".

I thought the optional arg would be simpler from the client's perspective, but I'm fine with using an empty string for default if you and @SuperQ prefer that.

discordianfish:
I don't like the variadic either, so would prefer empty string over that.

But to be honest, I don't really understand why we want to get rid of the fs struct in the first place? I probably missed some conversation around that.

I agree with #44 but that should merely rename `NewBuddyInfo()` to `BuddyInfo()` etc.

mdlayher:
> My reasoning for removing the FS type was try to match the signature of the functions in the refactored subpackages in #135. I could add an FS type to each of those subpackages, but it seemed like unnecessary duplication.

I'm not totally sure the FS type needs to go away either, even if the other packages no longer match. Perhaps the sub-packages can go without, and we can just keep it here for simplicity.

pgier:
> But to be honest, I don't really understand why we want to get rid of the fs struct in the first place? I probably missed some conversation around that.

@discordianfish There wasn't really conversation around it, just me responding to @grobie asking if procfs and sysfs packages would change to match the refactored subpackages.
https://github.com/prometheus/procfs/pull/136#issuecomment-476866062

So I was trying to come up with a way that the packages could have a consistent API with less boilerplate code.  After refactoring, the code in various node_exporter collectors would go from this:
```go
    fs, err := procfs.NewFS(*procPath)
    if err != nil {
        return fmt.Errorf("failed to open procfs: %v", err)
    }

    buddyInfo, err := fs.NewBuddyInfo()
```
to this:
```go
    buddyInfo, err := procfs.ReadBuddyInfo(*procPath)
```
So it would save a few lines and avoid repeatedly checking if the proc file path exists.

mdlayher:
I think it'd make sense to have the struct types in node_exporter construct the FS instance once and then re-use it throughout, as it cannot change at runtime.

discordianfish:
Agreeing with @mdlayher. Not sure why we didn't commented that on the initial PR.

I'd construct FS in `NewBuddyinfoCollector` and then do:
```
    buddyInfo, err := c.fs.BuddyInfo()
```

Seems like the nicer interface to me. But yeah #136 does it like this here.. So not sure but I feel like #136 should be converted to a similar API instead of the other way around.

To avoid the dependency it addressed in the first place, nfs for example would use it's own struct with the procfs path. Or we could also add read functions that prepend the proc path to the root FS type and embed a reference in all subpackages like nfs. Might be a nice option.

Either way, would like some more input from @grobie or @SuperQ on this. And sorry that I didn't brought that up in #135

pgier:
> So not sure but I feel like #136 should be converted to a similar API instead of the other way around.

I had originally experimented with creating FS in each package in #136, but I wasn't crazy about copying the fs.go file to each package, and we can't reuse the existing FS type since go doesn't allow defining methods on a type in a different package.  The other thing I didn't really like is that the fs.Path method is really just path.Join and is only needed because FS is a redefined string.  And it means that we can't use other functions from the standard `path` package since they all operate on strings.

> Or we could also add read functions that prepend the proc path to the root FS type and embed a reference in all subpackages like nfs. Might be a nice option.

Can you give an example of this?  I'm not sure I understand this option.  Do you mean multiple Read functions like this?

```go
func ReadBuddyInfo() {...}
func ReadBuddyInfoWithPath(mountPoint string) {...}
```
> And sorry that I didn't brought that up in #135

No problem!  Thanks for all the feedback on this one!

pgier:
> I think it'd make sense to have the struct types in node_exporter construct the FS instance once and then re-use it throughout, as it cannot change at runtime.

I agree that seems better.  Here is a PR for the change to node_exporter, PTAL:
https://github.com/prometheus/node_exporter/pull/1315

discordianfish:
> Can you give an example of this? I'm not sure I understand this option. Do you mean multiple Read functions like this?

No, I'd make BuddyInfo() a function on FS and add an Open() function which opens files relative to the procfs path.

For submodules, which we would use when it's non-trivial to parse and pulls in dependencies, like it's the case with NFS, I'd store a reference to a FS instance in a struct and make all functions members of that. Those functions then can use FS.Open(..) to read the procfs entries.

```
type NFS struct {
  *procfs.FS
}

func (n *NFS) ClientRPCStats() (*ClientRPCStats, error) {
  f, err := n.FS.Open(..)
```

Not 100% sure this is better, but what I'd do.

pgier:
We also need to be able to handle the xfs and the new blockdevice submodules which can use both proc and/or sys.  In that case I guess we'd need to initialize a struct with both proc and sys paths?

```go
type XFS struct {
  proc *procfs.FS
  sys  *sysfs.FS
}
```

grobie:
Hey! I don't have much time to look into Prometheus at the moment. I'll try to quickly explain my reasoning for the original library design, hoping that helps.

- I wanted to avoid the repetition of having a path argument on every function, so I created the FS type
- as procfs is commonly mounted under `/proc` I wanted to provide helper functions to make it easier for new users to get started with the library, saving 3 lines of initializing the fs type first
- It wasn't my intention to have these helper functions for all procfs functions, only for the most common ones like `Self()`, `Stat()` etc.
- I never documented this and the definition of "common" is arguably very subjective
- Seeing the explosion of public functions in this package, I don't think it was a good idea to create the helper package functions in the first place

When I saw @pgier's pr I felt it is moving the project in the right direction, reducing the coupling between packages and removing the number of public functions. I'm a strong proponent of consistency and asked for future plans around aligning the function signatures in the packages in this repo. I don't have the capacity to think about a good package redesign at the moment.

I'd suggest to try keeping the behavior of the various procfs/sysfs/... packages consistent, but I trust you all to come to a good decision. I'd also suggest to try reducing the number of functions available in the package, e.g. avoid having things like `ReadBuddyInfo/ReadBuddyInfoWithPath` or `pkg.ReadBuddyInfo/FS.ReadBuddyInfo` pairs. 

Sorry I'm not more helpful at the moment.

SuperQ:
Thanks for the historical info @grobie.

pgier:
After thinking about this a bit more, I would still prefer if we can remove the FS type because (1) it results in slightly fewer lines of code both inside the API and for users, (2) it's a bit simpler to understand because a user doesn't need to think about the FS type, and (3) it means one less bit of state to maintain in user/client code such as the node_exporter collectors.

Would everyone be ok with removing the FS type if I change the Read functions to use a regular string for the mount point instead of the variadic string?  An empty string would use the default path as @mdlayher suggested.

discordianfish:
Dunno, I like the current interface more. Not sure whether we should embed FS references in the sub modules or just have a path to proc in there but I find it much nicer to not pass in an empty string or whatever.

pgier:
@discordianfish ok, I created #149 to try out the approach using embedded FS references, please take a look when you have a chance!

pgier:
Closing in favor of #149

PR #149, opened 2019-04-11
use FS struct for proc and sys mount points in submodules

pgier:
Based on discussion in #148, instead of removing the FS type completely, this PR moves the FS type to a shared internal package and exposes a more consistent API for each of the subpackages.

### DIFF
--- a/bcache/get.go
+++ b/bcache/get.go
@@ -26,31 +26,31 @@ import (
 	"github.com/prometheus/procfs/sysfs"
 )
 
-// Bcache represents the pseudo-filesystem proc, which provides an interface to
+// Handle represents the pseudo-filesystem proc, which provides an interface to
 // kernel data structures.
-type Bcache struct {
+type Handle struct {
 	sysfs *sysfs.FS
 }
 
 // DefaultSysMountPoint is the common mount point of the sys filesystem.
 const DefaultSysMountPoint = "/sys"
 
-// NewBcache returns a new Bcache using the given sys fs mount point. It will error
+// New returns a new Bcache using the given sys fs mount point. It will error
 // if the mount point can't be read.
-func NewBcache(mountPoint string) (Bcache, error) {
+func New(mountPoint string) (Handle, error) {
 	if strings.TrimSpace(mountPoint) == "" {
 		mountPoint = DefaultSysMountPoint
 	}
 	fs, err := sysfs.NewFS(mountPoint)
 	if err != nil {
-		return Bcache{}, err
+		return Handle{}, err
 	}
-	return Bcache{&fs}, nil
+	return Handle{&fs}, nil
 }
 
 // Stats retrieves bcache runtime statistics for each bcache.
-func (b Bcache) Stats() ([]*Stats, error) {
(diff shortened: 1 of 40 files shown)


bcache/get.go
```diff
@@ -22,11 +22,35 @@ import (
...
+
+	"github.com/prometheus/procfs/sysfs"
 )
 
-// ReadStats retrieves bcache runtime statistics for each bcache.
-func ReadStats(sysfs string) ([]*Stats, error) {
-	matches, err := filepath.Glob(path.Join(sysfs, "fs/bcache/*-*"))
+// Bcache represents the pseudo-filesystem proc, which provides an interface to
+// kernel data structures.
+type Bcache struct {
```
mdlayher: I'm mobile and will do more review tomorrow, but naming not on these changes: bcache.Bcache and the like are stuttering names which are usually avoided in Go. Same with bcache.NewBcache.

How about calling these types "Handle" and the constructors just "New"?

bcache.Handle, bcache.New
xfs.Handle, xfs.New
Etc.

Will give a closer look later.

pgier: Yeah, makes sense.  I'm wondering if I should just call them FS and NewFS to match the main procfs and sysfs modules.

pgier:
One drawback of this approach is that the affected subpackages now have a dependency on the main procfs and/or sysfs packages.  It might be better to add the FS type to the internal/util package and have the sub-modules use that instead of the FS from procfs and sysfs.

pgier:
@mdlayher @discordianfish PTAL.  I've updated this PR to move the FS/string type to the internal/util package.  This allows the sub-packages bcache, blockdevice, nfs, and xfs to work without a dependency on the main procfs and sysfs packages, which was the original purpose of #136.  And I made changes to the procfs and sysfs packages to be consistent with the subpackages.

discordianfish:
Great, I like that!
One suggestion though (and I'm open for different opinions on this): I would create a fs/ module and move the functions there instead of util/. Util packages are seen as anti-pattern and while I wouldn't avoid them at all cost, it seems reasonable to use a fs module for these functions.

bcache/get.go
```diff
@@ -22,11 +22,35 @@ import (
...
-func ReadStats(sysfs string) ([]*Stats, error) {
-	matches, err := filepath.Glob(path.Join(sysfs, "fs/bcache/*-*"))
+// FS represents the pseudo-filesystem proc, which provides an interface to
+// kernel data structures.
+type FS struct {
+	sys *fs.FS
+}
+
+// DefaultSysMountPoint is the common mount point of the sys filesystem.
+const DefaultSysMountPoint = "/sys"
```
SuperQ: What about having this in the `fs` package? Then we can use `fs.DefaultSysMountPoint` here.

pgier: Good idea, saves a bit of duplication.  Done!

SuperQ:
@mdlayher Any other issues?

pgier:
@discordianfish everything look good to you?

discordianfish:
Great, thanks! :)

PR #154, opened 2019-05-07
sysfs: add parsing of clocksource

pgier:
Read current and available clocksource data from
/sys/devices/system/clocksource

Related to https://github.com/prometheus/node_exporter/issues/1336

### DIFF
--- a/sysfs/clocksource.go
+++ b/sysfs/clocksource.go
@@ -22,32 +22,23 @@ import (
 	"github.com/prometheus/procfs/internal/util"
 )
 
-// Clocksource contains metrics related to the clock source
-type Clocksource struct {
+// ClockSource contains metrics related to the clock source
+type ClockSource struct {
 	Name      string
 	Available []string
 	Current   string
 }
 
-// NewClocksource returns clocksource information including current and available clocksources.
-func NewClocksource() ([]Clocksource, error) {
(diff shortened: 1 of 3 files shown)


sysfs/clocksource.go
```diff
@@ -0,0 +1,85 @@
...
+	fs, err := NewFS(DefaultMountPoint)
+	if err != nil {
+		return []Clocksource{}, err
+	}
+
+	return fs.NewClocksource()
+}
+
+// NewClocksource returns clocksource information including current and available clocksources.
+func (fs FS) NewClocksource() ([]Clocksource, error) {
```
discordianfish: This should probably be Clocksources(), following #44 
I'm not sure though about the top level function.. tricky.. But since the structs are mainly just holding data for for consistency with the interface on `FS` we should probably call both Clocksources() (or maybe ClockSources?)

discordianfish:
But LGTM beside the naming.

sysfs/clocksource.go
pgier: Based on the comments here and in issue #44, I removed the top level constructor function and changed the FS method name to `ClockSources()`.

pgier:
@discordianfish Does this look ok now?

discordianfish:
Great, LGTM!

PR #162, opened 2019-05-21
Refactor toplevel functions

pgier:
See issue #44

### DIFF
--- a/mdstat.go
+++ b/mdstat.go
@@ -42,65 +42,64 @@ type MDStat struct {
 	BlocksSynced int64
 }
 
-// ParseMDStat parses an mdstat-file and returns a struct with the relevant infos.
-// TODO: this should just be called "MDStat" and use a separate parseMDStat function
-func (fs FS) ParseMDStat() (mdstates []MDStat, err error) {
-	mdStatusFilePath := fs.proc.Path("mdstat")
-	content, err := ioutil.ReadFile(mdStatusFilePath)
+// MDStat parses an mdstat-file (/proc/mdstat) and returns a slice of
+// structs containing the relevant info.  More information available here:
+// https://raid.wiki.kernel.org/index.php/Mdstat
+func (fs FS) MDStat() ([]MDStat, error) {
+	data, err := ioutil.ReadFile(fs.proc.Path("mdstat"))
 	if err != nil {
-		return []MDStat{}, fmt.Errorf("error parsing %s: %s", mdStatusFilePath, err)
+		return nil, fmt.Errorf("error parsing mdstat %s: %s", fs.proc.Path("mdstat"), err)
 	}
+	mdstat, err := parseMDStat(data)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing mdstat %s: %s", fs.proc.Path("mdstat"), err)
+	}
+	return mdstat, nil
+}
 
-	mdStates := []MDStat{}
-	lines := strings.Split(string(content), "\n")
+// parseMDStat parses data from mdstat file (/proc/mdstat) and returns a slice of
+// structs containing the relevant info.
+func parseMDStat(mdstatData []byte) ([]MDStat, error) {
+	mdStats := []MDStat{}
+	lines := strings.Split(string(mdstatData), "\n")
 	for i, l := range lines {
-		if l == "" {
-			continue
-		}
-		if l[0] == ' ' {
-			continue
-		}
-		if strings.HasPrefix(l, "Personalities") || strings.HasPrefix(l, "unused") {
+		if strings.TrimSpace(l) == "" || l[0] == ' ' ||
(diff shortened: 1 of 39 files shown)


mdstat.go
```diff
@@ -43,6 +43,7 @@ type MDStat struct {
 }
 
 // ParseMDStat parses an mdstat-file and returns a struct with the relevant infos.
+// TODO: this should just be called "MDStat" and use a separate parseMDStat function
```
mdlayher: Nit: this comment will appear in godoc documentation, consider moving it to line 48 as the first line of the function body instead.

proc_psi.go
```diff
@@ -51,19 +51,8 @@ type PSIStats struct {
...
-	if err != nil {
-		return PSIStats{}, err
-	}
-
-	return fs.NewPSIStatsForResource(resource)
-}
-
-// NewPSIStatsForResource reads pressure stall information from /proc/pressure/<resource>
-func (fs FS) NewPSIStatsForResource(resource string) (PSIStats, error) {
+// PSIStatsForResource reads pressure stall information from /proc/pressure/<resource>
```
mdlayher: You've lost some of the information here about which types of resources are valid, although IMO this should take an enum of some kind instead of a raw string.

mdlayher: For now, I'd say restore the missing doc comment and maybe add a TODO in the body here too.

mdstat.go
```diff
@@ -42,75 +44,79 @@ type MDStat struct {
...
-		if len(lines) <= i+3 {
-			return mdStates, fmt.Errorf(
-				"error parsing %s: too few lines for md device %s",
-				mdStatusFilePath,
-				mdName,
-			)
+		mdStat := MDStat{}
+		deviceLineFields := strings.Fields(line)
+		if len(deviceLineFields) < 3 {
+			return mdStats, fmt.Errorf("mdstat device line does not contain enough fields: %s", line)
```
mdlayher: Since we're refactoring this now, any error return sites should `return nil, err` rather than whatever state has been accumulated thus far.

mdstat.go
```diff
@@ -42,75 +44,82 @@ type MDStat struct {
...
+		if strings.TrimSpace(line) == "" ||
+			strings.HasPrefix(line, "Personalities") ||
+			strings.HasPrefix(line, "unused") {
 			continue
 		}
 
-		mainLine := strings.Split(l, " ")
-		if len(mainLine) < 3 {
-			return mdStates, fmt.Errorf("error parsing mdline: %s", l)
+		mdStat := MDStat{}
```
mdlayher: Since this is referred to many times and the function is about parsing these structures, consider a short name like `m`.

mdstat.go
pgier: I implemented the changes and removed the TODO.

mdstat.go
pgier: Done

mdstat.go
pgier: Done

xfs/xfs.go
```diff
@@ -177,6 +177,12 @@ type FS struct {
 	sys  *fs.FS
 }
 
+// NewDefaultFS returns a new XFS handle using the default proc and sys mountPoints.
+// It will error if either of the mounts point can't be read.
+func NewDefaultFS(procMountPoint string, sysMountPoint string) (FS, error) {
```
mdlayher: Oops, no parameters should be here.

pgier: Fixed.

pgier:
@discordianfish Does this look ok to you?

pgier:
Rabased and cleaned up the commits.

sbramin:
This breaks the client_golang pkg as NewStat has been renamed to Stat
```go/src/github.com/prometheus/client_golang/prometheus/process_collector.go:169:20: p.NewStat undefined (type procfs.Proc has no field or method NewStat)```

unless you are making changes to that at the same time?

peterbourgon:
:sad:

pgier:
@sbramin @peterbourgon  The plan is to update client_golang to match this updated API.  client_golang should still be pointing to the older commit in its module config, are you automatically updating to latest?
If needed, we could tag a release version before this change, or I can revert if this is causing a problem.

sbramin:
Depends on the expected time frame to update client_golang, If its a matter of hours then I can sit on my hands, if its days then perhaps its better to have procfs/golang_client being released at the same time? As I am sure I am not the only cowboy running off of master. Re-adding a deprecated section sounds like more trouble than its worth.

peterbourgon:
Modules are not universally used. My vote would be for

- Revert
- Add new constructors to procfs
- Migrate client_golang to new constructors
- Remove old constructors

peterbourgon:
Something needs to be done immediately, the Prometheus universe is currently broken.

PR #164, opened 2019-05-23
Add support for InfiniBand

bdrung:
The Prometheus node exporter collects metrics for the InfiniBand network protocol including the amount of packets sent and received, the number of times the link has been downed and how many times the link has recovered from an error state.

Reading all those information from sysfs is better placed in the procfs library. Also collect the state, physical state, and the rate.

### DIFF
--- a/internal/util/valueparser.go
+++ b/internal/util/valueparser.go
@@ -52,6 +52,25 @@ func (vp *ValueParser) PInt64() *int64 {
 	return &v
 }
 
+// PUInt64 interprets the underlying value as an uint64 and returns a pointer to
+// that value.
+func (vp *ValueParser) PUInt64() *uint64 {
+	if vp.err != nil {
+		return nil
+	}
+
+	// A base value of zero makes ParseInt infer the correct base using the
+	// string's prefix, if any.
+	const base = 0
+	v, err := strconv.ParseUint(vp.v, base, 64)
+	if err != nil {
+		vp.err = err
+		return nil
+	}
+
+	return &v
+}
+
 // Err returns the last error, if any, encountered by the ValueParser.
 func (vp *ValueParser) Err() error {
 	return vp.err
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -1,0 +1,370 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package sysfs
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+const infinibandClassPath = "class/infiniband"
+
+// InfiniBandCounters contains counter values from files in
+// /sys/class/infiniband/<Name>/ports/<Port>/counters or
+// /sys/class/infiniband/<Name>/ports/<Port>/counters_ext
+// for a single port of one InfiniBand device.
+type InfiniBandCounters struct {
+	LegacyPortMulticastRcvPackets  *uint64 // counters_ext/port_multicast_rcv_packets
+	LegacyPortMulticastXmitPackets *uint64 // counters_ext/port_multicast_xmit_packets
+	LegacyPortRcvData64            *uint64 // counters_ext/port_rcv_data_64
+	LegacyPortRcvPackets64         *uint64 // counters_ext/port_rcv_packets_64
+	LegacyPortUnicastRcvPackets    *uint64 // counters_ext/port_unicast_rcv_packets
+	LegacyPortUnicastXmitPackets   *uint64 // counters_ext/port_unicast_xmit_packets
+	LegacyPortXmitData64           *uint64 // counters_ext/port_xmit_data_64
+	LegacyPortXmitPackets64        *uint64 // counters_ext/port_xmit_packets_64
+
+	LinkDowned               *uint64 // counters/link_downed
+	LinkErrorRecovery        *uint64 // counters/link_error_recovery
+	MulticastRcvPackets      *uint64 // counters/multicast_rcv_packets
+	MulticastXmitPackets     *uint64 // counters/multicast_xmit_packets
+	PortRcvConstraintErrors  *uint64 // counters/port_rcv_constraint_errors
+	PortRcvData              *uint64 // counters/port_rcv_data
+	PortRcvDiscards          *uint64 // counters/port_rcv_discards
+	PortRcvErrors            *uint64 // counters/port_rcv_errors
+	PortRcvPackets           *uint64 // counters/port_rcv_packets
+	PortXmitConstraintErrors *uint64 // counters/port_xmit_constraint_errors
+	PortXmitData             *uint64 // counters/port_xmit_data
+	PortXmitDiscards         *uint64 // counters/port_xmit_discards
+	PortXmitPackets          *uint64 // counters/port_xmit_packets
+	PortXmitWait             *uint64 // counters/port_xmit_wait
+	UnicastRcvPackets        *uint64 // counters/unicast_rcv_packets
+	UnicastXmitPackets       *uint64 // counters/unicast_xmit_packets
+}
+
+// InfiniBandPort contains info from files in
+// /sys/class/infiniband/<Name>/ports/<Port>
+// for a single port of one InfiniBand device.
+type InfiniBandPort struct {
+	Name        string
+	Port        uint
+	State       string // String representation from /sys/class/infiniband/<Name>/ports/<Port>/state
+	StateID     uint   // ID from /sys/class/infiniband/<Name>/ports/<Port>/state
+	PhysState   string // String representation from /sys/class/infiniband/<Name>/ports/<Port>/phys_state
+	PhysStateID uint   // String representation from /sys/class/infiniband/<Name>/ports/<Port>/phys_state
+	Rate        uint64 // in bytes/second from /sys/class/infiniband/<Name>/ports/<Port>/rate
+	Counters    InfiniBandCounters
+}
+
+// InfiniBandDevice contains info from files in /sys/class/infiniband for a
+// single InfiniBand device.
+type InfiniBandDevice struct {
+	Name            string
+	BoardID         string // /sys/class/infiniband/<Name>/board_id
+	FirmwareVersion string // /sys/class/infiniband/<Name>/fw_ver
+	HCAType         string // /sys/class/infiniband/<Name>/hca_type
+	Ports           map[uint]InfiniBandPort
+}
+
+// InfiniBandClass is a collection of every InfiniBand device in
+// /sys/class/infiniband.
+//
+// The map keys are the names of the InfiniBand devices.
+type InfiniBandClass map[string]InfiniBandDevice
+
+// InfiniBandClass returns info for all InfiniBand devices read from
+// /sys/class/infiniband.
+func (fs FS) InfiniBandClass() (InfiniBandClass, error) {
+	path := fs.sys.Path(infinibandClassPath)
+
+	dirs, err := ioutil.ReadDir(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list InfiniBand devices at %q: %v", path, err)
+	}
+
+	ibc := make(InfiniBandClass, len(dirs))
+	for _, d := range dirs {
+		device, err := fs.parseInfiniBandDevice(d.Name())
+		if err != nil {
+			return nil, err
+		}
+
+		ibc[device.Name] = *device
+	}
+
+	return ibc, nil
+}
+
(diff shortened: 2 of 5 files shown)


bdrung:
Thanks for the review. I will address those points. @mdlayher, do you really prefer to have a switch statement for the 16 different values? This list might grow. If the counter fields are regular values, what should I do if the system does not have this counter?

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,239 @@
...
+		res = append(res, deviceDir.Name())
+	}
+
+	return res, nil
+}
+
+// InfiniBandDevicePorts scans /sys/class/infiniband/<device> for ports and returns them as a list of integers.
+func (fs FS) InfiniBandDevicePorts(device string) ([]uint, error) {
+	var res []uint
+	path := fs.sys.Path(filepath.Join(infinibandclassPath, device, "ports"))
```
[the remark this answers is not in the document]
bdrung: Done.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,239 @@
...
+	infiniBandClass := InfiniBandClass{}
+	for _, device := range devices {
+		ports, err := fs.InfiniBandDevicePorts(device)
+		if err != nil {
+			return nil, err
+		}
+
+		infiniBandClass[device] = map[uint]InfiniBandPort{}
+		for _, port := range ports {
+			path := fs.sys.Path(filepath.Join(infinibandclassPath, device, "ports", fmt.Sprint(port)))
```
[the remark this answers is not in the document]
bdrung: Done.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,239 @@
...
+		if err != nil {
+			return res, fmt.Errorf("failed to convert %s into uint", portDir.Name())
+		}
+		res = append(res, uint(port))
+	}
+
+	return res, nil
+}
+
+// NewNetClass returns info for all ports (port) of all InfiniBand devices (device) read from /sys/class/infiniband/<device>/ports/<port>.
```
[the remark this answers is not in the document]
bdrung: Done.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,239 @@
...
+func (fs FS) NewInfiniBandClass() (InfiniBandClass, error) {
```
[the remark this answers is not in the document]
bdrung: Okay. Renamed to `InfiniBandClass`.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return rate, fmt.Errorf("failed to convert %s into uint", strings.TrimSpace(s[0]))
+	}
+	rate = value * 125000000
+	return rate, nil
+}
+
+// parseInfiniBandPort scans predefined files in /sys/class/infiniband/<device>/ports/<port>
+// directory and gets their contents.
+func (_ InfiniBandClass) parseInfiniBandPort(portPath string) (*InfiniBandPort, error) {
```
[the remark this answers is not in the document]
bdrung: Done.

bdrung:
I looked at https://github.com/prometheus/node_exporter/blob/master/collector/infiniband_linux.go and borrowed the workarounds and the legacy counters.

bdrung:
I changed the two parse functions to take a string as parameter (instead of a path to the file).

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,272 @@
...
+		return id, name, fmt.Errorf("failed to convert %s into uint", strings.TrimSpace(parts[0]))
+	}
+	id = uint(value)
+	return id, name, nil
+}
+
+// Parse rate (example: "100 Gb/sec (4X EDR)") and return it as bytes/second
+func parseRate(s string) (uint64, error) {
+	var rate uint64
+	parts := strings.Split(s, "Gb/sec")
```
discordianfish: Is this the only unit possible?

discordianfish:
Tend to agree with @mdlayher. What about using a temporary map to iterate over instead of using reflect to iterate over the map?

sysfs/class_infiniband.go
dswarbrick: Yes - see `rate_show` function in https://elixir.bootlin.com/linux/v5.1.7/source/drivers/infiniband/core/sysfs.c:
```c
	return sprintf(buf, "%d%s Gb/sec (%dX%s)\n",
		       rate / 10, rate % 10 ? ".5" : "",
		       ib_width_enum_to_int(attr.active_width), speed);
```

bdrung: I changed to code to also correctly parse `2.5 Gb/sec` and added a test case for it.

bdrung:
`sysfs/class_power_supply.go` and `sysfs/net_class.go` uses reflect. Can you point me to a component that uses a temporary map to iterate over?

bdrung:
@mdlayher I saw your recent commits on master and adapted this pull request to the logic used for PowerSupply. Therefore `reflect` is not used any more. Additional changes:
* Introduce `InfiniBandDevice` type
* also collect `board_id`, `fw_ver`, and `hca_type` (in InfiniBandDevice)

All remarks are addressed and this pull request is ready for an in depth review.

internal/util/valueparser.go
```diff
@@ -52,6 +52,25 @@ func (vp *ValueParser) PInt64() *int64 {
 	return &v
 }
 
+// PUInt64 interprets the underlying value as an uint64 and returns a pointer to
+// that value.
+func (vp *ValueParser) PUInt64() *uint64 {
```
mdlayher: Mind adding a test for this new method as well? You can probably just duplicate the PInt64 cases.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,372 @@
...
+
+	return ibc, nil
+}
+
+// Parse one InfiniBand device.
+func (fs FS) parseInfiniBandDevice(name string) (*InfiniBandDevice, error) {
+	path := fs.sys.Path(infinibandClassPath, name)
+	device := InfiniBandDevice{Name: name}
+
+	for _, f := range [3]string{"board_id", "fw_ver", "hca_type"} {
```
mdlayher: Nit: how about `[...]` so the size will be inferred automatically? I'd usually just go with a slice but an array seems reasonable since this will probably never change.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,372 @@
...
+
+		device.Ports[port.Port] = *port
+	}
+
+	return &device, nil
+}
+
+// Parse InfiniBand state. Expected format: "<id>: <string-representation>"
+func parseState(s string) (uint, string, error) {
+	var id uint
```
mdlayher: IMO remove these and just use the usual `:=` syntax below.

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,372 @@
...
+	var rate uint64
+	parts := strings.Split(s, "Gb/sec")
+	if len(parts) != 2 {
+		return rate, fmt.Errorf("failed to split '%s' by 'Gb/sec'", s)
+	}
+	value, err := strconv.ParseFloat(strings.TrimSpace(parts[0]), 32)
+	if err != nil {
+		return rate, fmt.Errorf("failed to convert %s into uint", strings.TrimSpace(parts[0]))
+	}
+	rate = uint64(value * 125000000)
```
mdlayher: Same, `:=`.

Can you leave a comment about the meaning of this constant?

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+// parseInfiniBandPort scans predefined files in /sys/class/infiniband/<device>/ports/<port>
+// directory and gets their contents.
+func (fs FS) parseInfiniBandPort(name string, port string) (*InfiniBandPort, error) {
+	portNumber, err := strconv.ParseUint(port, 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert %s into uint", port)
+	}
+	infiniBandPort := InfiniBandPort{Name: name, Port: uint(portNumber)}
```
mdlayher: Nit: you can go with a shorter name since this is confined to a single function. Maybe `ibp`?

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,372 @@
...
+		name := filepath.Join(path, f.Name())
+		value, err := util.SysReadFile(name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read file %q: %v", name, err)
+		}
+
+		// According to Mellanox, the metrics port_rcv_data, port_xmit_data,
+		// port_rcv_data_64, and port_xmit_data_64 "are divided by 4 unconditionally"
+		// as they represent the amount of data being transmitted and received per lane.
+		// Mellanox cards have 4 lanes per port, so all values must be multiplied by 4
```
mdlayher: Does anyone other than Mellanox make infiniband gear?

sysfs/class_infiniband.go
```diff
@@ -0,0 +1,372 @@
...
+)
+
+const infinibandClassPath = "class/infiniband"
+
+// InfiniBandCounters contains counter values from files in
+// /sys/class/infiniband/<Name>/ports/<Port>/counters or
+// /sys/class/infiniband/<Name>/ports/<Port>/counters_ext
+// for a single port of one InfiniBand device.
+type InfiniBandCounters struct {
+	LegacyPortMulticastRcvPackets  *uint64 // counters_ext/port_multicast_rcv_packets
```
mdlayher: Do we have to have the legacy versions as well? Would it be possible to look for both new/legacy and only use legacy if the new version is absent?

sysfs/class_infiniband_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Fatal(err)
+	}
+
+	got, err := fs.InfiniBandClass()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var (
+		port1LinkDowned               uint64 = 0
```
mdlayher: Nit: omit any `= 0`; it's inferred automatically.

sysfs/class_infiniband_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+package sysfs
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestParseSlowRate(t *testing.T) {
```
mdlayher: Nit: combine this and the below test into a table-driven test.

sysfs/class_infiniband.go
dswarbrick: Mellanox (recently acquired by Nvidia) dominate the IB market by far, but previously Qlogic (tech acquired by Intel) and Voltaire (merged with Mellanox) produced IB hardware. Oracle produces their own IB chipsets.

sysfs/class_infiniband.go
dswarbrick: We run a mixture of kernel versions where I work, and I've had to do some elaborate metric relabelling to make our IB metric names conform to a common standard, so that they're supported by a one-size-fits-all Grafana dashboard. Brace yourself for a verbose reason why...

Kernels 4.5 and later merge the `counters_ext` IB counters in sysfs into the `counters` directory. This effectively replaces the 32-bit counters that were previously exposed in the `counters` directory with the 64-bit counters from the `counters_ext` directory.

Unfortunately, InfiniBand counters do not automatically roll over, so when the 32-bit counters reach 4 billion, they simply latch at that value (which is obviously bad for `rate` / `irate` queries).

Node_exporter doesn't really handle this in a sane way, so on pre-4.5 kernels it exposes 32-bit counters as e.g. `node_infiniband_port_packets_transmitted_total` and 64-bit counters as
e.g. `node_infiniband_legacy_packets_transmitted_total`. This makes it difficult to have unified dashboards / alerting rules which can handle both pre- and post-4.5 kernels.

To cut a long story short, I ended up configuring the following relabelling rules, which rename the legacy metrics if exposed, and overwrite the "non-legacy" metrics (which will be 32-bit, and thus useless, if legacy metrics are present):
```yaml
  # Rename node_infiniband_legacy_* (pre-MLNX_OFED 4.0) to unify PromQL queries
  - source_labels: [__name__]
    target_label: __name__
    regex: node_infiniband_legacy_((?:data|packets)_.+)
    replacement: node_infiniband_port_$1
  - source_labels: [__name__]
    target_label: __name__
    regex: node_infiniband_legacy_((?:multicast|unicast)_.+)
    replacement: node_infiniband_$1
```
In short - it would be nice if node_exporter simply exposed the widest counters available, regardless of kernel / OFED version. However, that is something that needs to happen in node_exporter, not procfs, and it may be desirable to still expose both legacy (32-bit) and modern (64-bit) counters, at least in procfs.

internal/util/valueparser.go
bdrung: Done.

sysfs/class_infiniband.go
bdrung: Done.

sysfs/class_infiniband.go
bdrung: Done. Added: "Convert Gb/s into bytes/s"

sysfs/class_infiniband_test.go
bdrung: Done, but I like to prefer the explicit definition.

sysfs/class_infiniband_test.go
bdrung: Done.

sysfs/class_infiniband.go
bdrung: Done.

sysfs/class_infiniband.go
bdrung: Thanks for the answer @dswarbrick. Therefore I mark this conversation as resolved without changing the code.

pgier:
Merged, thanks!

PR #165, opened 2019-05-28
Fix overflow in mounstats

SuperQ:
Don't use time.Duration for mountstats timing data. These values can be
large uint64 milliseconds values and easily overflow time.Duration
(in64 nanoseconds). Return the raw time in milliseconds.


mountstats.go
```diff
@@ -204,7 +204,7 @@ type NFSTransportStats struct {
 	// spent waiting for connections to the server to be established.
 	ConnectIdleTime uint64
 	// Duration since the NFS mount last saw any RPC traffic.
-	IdleTime time.Duration
+	IdleTime uint64
```
mdlayher: This needs a unit.

SuperQ: Fixed

SuperQ:
Yes, the example data came from a RHEL 7.3 system, 3.10 kernel.

PR #168, opened 2019-05-30
re-add functions required by client_golang

pgier:
Adding back some functions that were removed during refactoring.
These are needed by client_golang for backwards compatibility.
They are now marked deprecated and can be removed once there
are no more dependencies on them.

Related to PR #162

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ point is initialized, and then the stat information is read.
 
 ```go
 fs, err := procfs.NewFS("/proc")
-stats, err := fs.NewStat()
+stats, err := fs.Stat()
 ```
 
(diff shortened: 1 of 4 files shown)


pgier:
CC @peterbourgon

sbramin:
@pgier Thanks for the quick turn around!

discordianfish:
I don't get the urgency. Things change, depending code breaks. You need to account for that either way. There are solutions to handle that.. That being said, sounds reasonable to re-add.

beorn7:
@pgier also thanks for tagging v0.0.1. For the transition to the non-deprecated types, it would be handy to also have a version tag _after_ this PR (where both new and old are present). Essentially, just tagging current head with v0.0.2 would do.

PR #171, opened 2019-06-01
Rebase and resubmit PR for the lio iscsi procfs module 

AvengerMoJo:
replace the old PR https://github.com/prometheus/procfs/pull/88

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -1806,3 +1806,287 @@ Lines: 1
 extent_alloc 2 0 0 0
 Mode: 644
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/enable
+Lines: 1
+1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0/statistics
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0/statistics/scsi_tgt_port
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0/statistics/scsi_tgt_port/in_cmds
+Lines: 1
+104950
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0/statistics/scsi_tgt_port/read_mbytes
+Lines: 1
+20095
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0/statistics/scsi_tgt_port/write_mbytes
+Lines: 1
+71235
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/core
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/core/iblock_0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/core/iblock_0/block_lio_rbd1
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.abcd1abcd2ab/tpgt_1/lun/lun_0/795b7c7026
+SymlinkTo: ../../../../../../target/core/iblock_0/block_lio_rbd1
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/core/iblock_0/block_lio_rbd1/enable
+Lines: 1
+1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/core/iblock_0/block_lio_rbd1/udev_path
+Lines: 1
+/dev/rbd1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1/enable
+Lines: 1
+1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1/lun
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1/lun/lun_0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1/lun/lun_0/statistics
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1/lun/lun_0/statistics/scsi_tgt_port
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/kernel/config/target/iscsi/iqn.2003-01.org.linux-iscsi.osd1.x8664:sn.8888bbbbddd0/tpgt_1/lun/lun_0/statistics/scsi_tgt_port/in_cmds
+Lines: 1
+204950
+Mode: 644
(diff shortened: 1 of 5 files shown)


AvengerMoJo:
@pgier hi Paul, I try to rebase and merge with too much conflict. So I just create a new PR which is faster with all the remark update you send me from the #88 Thanks. 
I don't think I can add you as the reviewer.

iscsi/get.go
```diff
@@ -0,0 +1,314 @@
...
+	istats.Name = filepath.Base(iqnPath)
+	istats.RootPath = filepath.Dir(iqnPath)
+
+	matches, err := filepath.Glob(filepath.Join(iqnPath, "tpgt*"))
+	if err != nil {
+		return nil, fmt.Errorf("iscsi: GetStats: get TPGT path error %v", err)
+	}
+	istats.Tpgt = make([]TPGT, len(matches))
+
+	for ii, tpgtPath := range matches {
```
pgier: Rename the index to `i` instead of `ii` to better follow conventions.

iscsi/get.go
```diff
@@ -0,0 +1,314 @@
...
+		istats.Tpgt[ii].Name = filepath.Base(tpgtPath)
+		istats.Tpgt[ii].TpgtPath = tpgtPath
+		istats.Tpgt[ii].IsEnable, _ = isPathEnable(tpgtPath)
+		if istats.Tpgt[ii].IsEnable {
+			matchesLunsPath, _ := getLun(tpgtPath)
+			istats.Tpgt[ii].Luns = make([]LUN, len(matchesLunsPath))
+
+			for ll, lunPath := range matchesLunsPath {
+				backstore, objectName, typeNumber, err := getLunLinkTarget(lunPath)
```
pgier: I think it would be better here to return a `(LUN, err)` from `getLunLinkTarget()` and then do something like `istats.Tpgt[ii].Luns[ll] = lun`

iscsi/get.go
```diff
@@ -0,0 +1,314 @@
...
+
+	for ii, tpgtPath := range matches {
+		istats.Tpgt[ii].Name = filepath.Base(tpgtPath)
+		istats.Tpgt[ii].TpgtPath = tpgtPath
+		istats.Tpgt[ii].IsEnable, _ = isPathEnable(tpgtPath)
+		if istats.Tpgt[ii].IsEnable {
+			matchesLunsPath, _ := getLun(tpgtPath)
+			istats.Tpgt[ii].Luns = make([]LUN, len(matchesLunsPath))
+
+			for ll, lunPath := range matchesLunsPath {
```
pgier: I'd probably call the index here `j` instead of `ll`.

iscsi/iscsi.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// ConfigFsPath is static configfs mount point
+const ConfigFsPath = "/kernel/config"
+
+// TargetPath is static sys path for iscsi target
+const TargetPath = "/target/iscsi"
+
+// TargetCore static sys path for backstore
+const TargetCore = "/target/core"
```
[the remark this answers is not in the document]
AvengerMoJo: This module was originally split from https://github.com/AvengerMoJo/node_exporter/tree/lio_rework
Which also need to rework now since I have not touch it for a year as well.

iscsi/iscsi.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/fs"
+)
+
+// ConfigFsPath is static configfs mount point
+const ConfigFsPath = "/kernel/config"
```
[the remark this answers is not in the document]
AvengerMoJo: Sure, done just pushed.

iscsi/get.go
AvengerMoJo: Do you mean, I should create another function to call on top? Because I still need do this dynamic assignment inside that function to return the LUN list and passing the Tgpt object inside that function to return LUN object into it.

iscsi/iscsi.go
```diff
@@ -0,0 +1,125 @@
...
+
+// TargetPath is static sys path for iscsi target
+const TargetPath = "/target/iscsi"
```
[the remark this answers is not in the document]
AvengerMoJo: haha, I don't remember why I did it this way. But I'm pretty sure at some point I started the path from the root target path /target/iscsi and build the rest from there. But sure I can update this.

iscsi/iscsi.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// NewFS returns a new configfs mounted under the given mountPoint. It will error
+// if the mount point can't be read.
+func NewFS(configfsMountPoint string) (FS, error) {
+	if strings.TrimSpace(configfsMountPoint) == "" {
+		// default path should be /sys/kernel/config
+		configfsMountPoint = fs.DefaultSysMountPoint + ConfigFsPath
+	}
+	Configfs, err := fs.NewFS(configfsMountPoint)
```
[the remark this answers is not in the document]
AvengerMoJo: Thanks, done, pushed.

iscsi/iscsi.go
```diff
@@ -0,0 +1,125 @@
...
+// TargetPath is static sys path for iscsi target
+const TargetPath = "/target/iscsi"
+
+// TargetCore static sys path for backstore
+const TargetCore = "/target/core"
+
+// FS represents the pseudo-filesystem configfs, which provides an interface to
+// iscsi kernel data structures.
+type FS struct {
+	Configfs *fs.FS
```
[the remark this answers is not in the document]
AvengerMoJo: I really don't remember why I need to export it. Updated and pushed.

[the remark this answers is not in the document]
AvengerMoJo: oh ... I know why now, it is the test case debug log ... I was trying to show what path is being set and printing it out as log. I will remove the log.

iscsi/get.go
pgier: What I mean is just change the return type of `getLunLinkTarget()` to something like `(*LUN, error)`, so instead of receiving multiple return values and then assigning them individually.  You can just assign the whole struct to the array.  And if there is an error, you can return `nil` for the lun struct.

pgier: The contents of the inner loop becomes something like this:
```golang
	for j, lunPath := range matchesLunsPath {
		lun, err := getLunLinkTarget(lunPath)
		if err != nil {
			continue
		}
		istats.Tpgt[i].Luns[j] = *lun
	}
```

pgier: Or could maybe simplify a bit?
```golang
	for _, lunPath := range matchesLunsPath {
		lun, err := getLunLinkTarget(lunPath)
		if err == nil {
			istats.Tpgt[i].Luns = append(istats.Tpgt[i].Luns, lun)
		}
	}
```

AvengerMoJo: Lovely I can do this for sure. I will update it.

iscsi/iscsi.go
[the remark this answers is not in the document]
pgier: That's fine, if you expect it to be used from outside the package, you can just leave it.  But maybe remove the first slash to make it clear that it's a relative path and not absolute.

AvengerMoJo:
@pgier do you I should rebase squash all my commit?

iscsi/get.go
```diff
@@ -0,0 +1,311 @@
...
+
+// ReadWriteOPS read and return the stat of read and write in megabytes,
+// and total commands that send to the target
+func ReadWriteOPS(iqnPath string, tpgt string, lun string) (readmb uint64,
+	writemb uint64, iops uint64, err error) {
+
+	readmbPath := filepath.Join(iqnPath, tpgt, "lun", lun,
+		"statistics/scsi_tgt_port/read_mbytes")
+
+	if _, err := os.Stat(readmbPath); os.IsNotExist(err) {
```
pgier: This check seems a bit unnecessary since readUintFromFile will provide a similar error if the file doesn't exist.  Maybe this can be removed?  Or is there a reason to check specifically for file existence before trying to read the file?

internal/fs/fs.go
```diff
@@ -25,6 +25,9 @@ const (
 
 	// DefaultSysMountPoint is the common mount point of the sys filesystem.
 	DefaultSysMountPoint = "/sys"
+
+	// ConfigfsPath is the commont mount point of the configfs
```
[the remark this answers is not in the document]
AvengerMoJo: Thanks for baby sitting me. Fix and pushed.

iscsi/iscsi.go
```diff
@@ -0,0 +1,124 @@
...
+const TargetCore = "target/core"
+
+// FS represents the pseudo-filesystem configfs, which provides an interface to
+// iscsi kernel data structures in /sys/kernel/config.
+type FS struct {
+	configfs *fs.FS
+}
+
+// NewFS returns a new configfs mounted under the given mountPoint. It will error
+// if the mount point can't be read.
```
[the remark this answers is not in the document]
AvengerMoJo: Sure, done and pushed.

iscsi/get.go
```diff
@@ -0,0 +1,311 @@
...
+		return nil, fmt.Errorf("iscsi: GetStats: get TPGT path error %v", err)
+	}
+	istats.Tpgt = make([]TPGT, len(matches))
+
+	for i, tpgtPath := range matches {
+		istats.Tpgt[i].Name = filepath.Base(tpgtPath)
+		istats.Tpgt[i].TpgtPath = tpgtPath
+		istats.Tpgt[i].IsEnable, _ = isPathEnable(tpgtPath)
+		if istats.Tpgt[i].IsEnable {
+			matchesLunsPath, _ := getLun(tpgtPath)
```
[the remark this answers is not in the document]
AvengerMoJo: understand, I break is out from the lio node_exporter module which used to printing out debug error message but they suggest not to have error message print in a lib. Done, removed, Pushed.

iscsi/get.go
```diff
@@ -0,0 +1,311 @@
...
+
+func getLunLinkTarget(lunPath string) (lunObject LUN, err error) {
+	lunObject.Name = filepath.Base(lunPath)
+	lunObject.LunPath = lunPath
+	files, err := ioutil.ReadDir(lunPath)
+	if err != nil {
+		return lunObject, fmt.Errorf("getLunLinkTarget: ReadDir path %s error %v", lunPath, err)
+	}
+	for _, file := range files {
+		fileInfo, _ := os.Lstat(lunPath + "/" + file.Name())
```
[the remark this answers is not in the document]
AvengerMoJo: Sorry for the in-consistency of my using string cat to filepath here. Fixed and pushed. Thanks

iscsi/get.go
AvengerMoJo: I think I'm just trying to clear out the error type, before the iqn setup and enable all those status may not exist. Compare to if the read out is something unreadable that should be a different reason to debug from the user that using the lio module.

pgier: The error message returned from ioutil.ReadFile if the file is missing will say something like "`open <filename>: No such file or directory`", which is a fairly common error message, so I think it would be clear enough to the user.

iscsi/iscsi.go
```diff
@@ -0,0 +1,126 @@
...
+const TargetCore = "target/core"
+
+// FS represents the pseudo-filesystem configfs, which provides an interface to
+// iscsi kernel data structures in /sys/kernel/config.
+type FS struct {
+	configfs *fs.FS
+}
+
+// NewFS returns a new configfs mounted under the given mount point. It will
+// error and return empty FS. If the mount point can't be read. For the ease of
```
pgier: Looks like there is a trailing space at the end of this line which is causing a CI failure on the go fmt check.

internal/fs/fs.go
```diff
@@ -25,6 +25,9 @@ const (
 
 	// DefaultSysMountPoint is the common mount point of the sys filesystem.
 	DefaultSysMountPoint = "/sys"
+
+	// DefaultConfigfsMountPath is the commont mount point of the configfs
```
pgier: The comment should be updated to match the variable name.

iscsi/iscsi.go
```diff
@@ -0,0 +1,126 @@
...
+// FS represents the pseudo-filesystem configfs, which provides an interface to
+// iscsi kernel data structures in /sys/kernel/config.
+type FS struct {
+	configfs *fs.FS
+}
+
+// NewFS returns a new configfs mounted under the given mount point. It will
+// error and return empty FS. If the mount point can't be read. For the ease of 
+// use, an empty string parameter will call internal fs for the /sys/kernel/config
+// DefaultConfigfsMountPoint
```
pgier: Same with this line, there is a trailing space being picked up by the gofmt check.

iscsi/get.go
AvengerMoJo: Thanks for explaining it. I removed it and pushed.

internal/fs/fs.go
AvengerMoJo: Get it, done, pushed. Thanks

iscsi/iscsi.go
AvengerMoJo: sorry about that, forget to run gofmt before submitting and I was working on adding the test case and rewrite the getLunLinkTarget, and it should be all fix now with the new test case. Thanks for looking into it.

iscsi/iscsi.go
AvengerMoJo: Done, pushed.

iscsi/get.go
```diff
@@ -0,0 +1,283 @@
...
+		fileInfo, _ := os.Lstat(filepath.Join(lunPath, file.Name()))
+		if fileInfo.Mode()&os.ModeSymlink != 0 {
+			target, err := os.Readlink(filepath.Join(lunPath, fileInfo.Name()))
+			if err != nil {
+				return lunObject, fmt.Errorf("getLunLinkTarget: Readlink err %v", err)
+			}
+			targetPath, objectName := filepath.Split(target)
+			_, typeWithNumber := filepath.Split(filepath.Clean(targetPath))
+
+			splitTypeWithNumber := strings.Split(typeWithNumber, "_")
```
pgier: I think this could be a little nicer using `strings.LastIndex`, and then using the index to split the string.
```golang
underscore := strings.LastIndex(typeWithNumber, "_")
lunObject.Backstore = typeWithNumber[:underscore]
lunObject.ObjectName = typeWithNumber[underscore+1:]
```
Might also need to check for a filename without an underscore.

iscsi/get.go
```diff
@@ -0,0 +1,283 @@
...
+// isPathEnable is a utility function
+// check if the file "enable" contain enable message
+func isPathEnable(path string) (bool, error) {
+	var isEnable bool
+	isEnable = false
+	enableReadout, err := ioutil.ReadFile(filepath.Join(path, "enable"))
+	if err != nil {
+		return false, fmt.Errorf("iscsi: isPathEnable error %v", err)
+	}
+	enableReadoutNum, err := strconv.Atoi(strings.TrimSpace(string(enableReadout)))
```
pgier: Instead of converting to an integer and then interpreting the integer, I think you can use `strings.ParseBool` which will interpret `0` as false and `1` as true.

iscsi/get.go
```diff
@@ -0,0 +1,283 @@
...
+			continue
+		}
+		bSystemImage, err := ioutil.ReadFile(systemImagePath)
+		if err != nil {
+			continue
+		} else {
+			systemImage = strings.TrimSpace(string(bSystemImage))
+		}
+
+		if matchRBD(fmt.Sprintf("%d", systemRbdNumber), rbd.Rnumber) &&
```
pgier: I think `strconv.FormatInt(systemRbdNumber)` would be better here instead of `fmt.Sprintf`.

fixtures.ttar
```diff
@@ -1806,3 +1806,284 @@ Lines: 1
 extent_alloc 2 0 0 0
 Mode: 644
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/kernel
```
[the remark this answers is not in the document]
AvengerMoJo: Done, pushed. Updated the test accordingly.

fixtures.ttar
```diff
@@ -1806,3 +1806,284 @@ Lines: 1
...
+Lines: 1
+1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/kernel/config/target/core/rbd_0/iscsi-images-demo/udev_path
+Lines: 1
+/dev/rbd/iscsi-images/demo
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/devices
```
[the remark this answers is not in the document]
AvengerMoJo: Done, pushed. Updated the test accordingly.

iscsi/get.go
AvengerMoJo: There are 2 kind of of backstore name format, 
backstore_number or back_store_number 
I guess we can error out when no underscore is being found.

iscsi/get.go
AvengerMoJo: Thanks, yeah replace all the un-needed variable and test with strconv.ParseBool

iscsi/get.go
AvengerMoJo: Thanks, remove matchRBD all together and replace it with FormatInt(int64(systemRbdNumber) and inline string compare to reduce complexity.

AvengerMoJo:
@pgier thank you so much for all the comments, I'm looking forward to fix the lio module in node_exporter as well. Let me know if there is anything I should clean up.

pgier:
Also, would probably be good to squash this down to one or just a few commits since the commit history is becoming pretty large.

iscsi/get.go
```diff
@@ -0,0 +1,267 @@
+// Copyright 2017 The Prometheus Authors
```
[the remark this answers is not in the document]
AvengerMoJo: lol this reminded me how long this patch has been around, done, pushed.

iscsi/get.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("iscsi: GetRDMCPPath: error %v", err)
+	}
+	if isEnable {
+		return &rdmcp, nil
+	}
+	return nil, nil
+}
+
+func readUintFromFile(path string) (uint64, error) {
```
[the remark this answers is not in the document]
AvengerMoJo: Thanks, removed and replace with the internal util.

iscsi/get.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return 0, err
+	}
+	return value, nil
+}
+
+func matchPoolImage(pool string, image string, matchPoolImage string) (isEqual bool) {
+	isEqual = false
+	var poolImage = fmt.Sprintf("%s-%s", pool, image)
+	if strings.Compare(poolImage, matchPoolImage) == 0 {
```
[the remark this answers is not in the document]
AvengerMoJo: Sure, reduce it down. Done, pushed.

AvengerMoJo:
@pgier I squash it down, should I change the comments? I'm worry it maybe too much as well. Maybe I can merge back to upstream master when the review is done?

pgier:
Yes, would be good to clean up the commit comment as well.

mdlayher:
Apologies, I'm busy for a couple of weeks and probably won't have time to review.

PR #173, opened 2019-06-05
Parsing Mountinfo to get information not in mountstats

dipack95:
As @SuperQ said [here](https://github.com/prometheus/node_exporter/pull/994#issuecomment-499133986), I've added parsing for `mountinfo` to `procfs`, along with test cases for it.

### DIFF
--- a/mountinfo.go
+++ b/mountinfo.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 )
 
+var validOptionalFields = map[string]bool{"shared": true, "master": true, "propagate_from": true, "unbindable": true}
+
 // Fields description for mountinfo is here:
 // http://man7.org/linux/man-pages/man5/proc.5.html
 
@@ -50,23 +52,11 @@ type MountInfo struct {
 }
 
 // Returns part of the mountinfo line, if it exists, else an empty string
-func getMountInfoPartString(parts []string, idx int) string {
+func getStringSliceElement(parts []string, idx int, defaultValue string) string {
 	if idx >= len(parts) {
-		return ""
+		return defaultValue
 	}
 	return parts[idx]
-}
-
-// Checks if the 'keys' in the optional fields in the mountinfo line are acceptable
-// Allowed 'keys' are shared, master, propagate_from, unbindable
-func isValidOptionalField(field string) bool {
-	acceptableFields := []string{"shared", "master", "propagate_from", "unbindable"}
-	for _, acceptable := range acceptableFields {
-		if field == acceptable {
-			return true
-		}
-	}
-	return false
 }
 
 // Reads each line of the mountinfo file, and returns a list of formatted MountInfo structs
@@ -82,11 +72,8 @@ func parseMountInfo(r io.Reader) ([]*MountInfo, error) {
 		mounts = append(mounts, parsedMounts)
 	}
 
-	if err := scanner.Err(); err != nil {
-		return mounts, err
-	}
-
-	return mounts, nil
+	err := scanner.Err()
+	return mounts, err
 }
 
 // Parses a mountinfo file line, and converts it to a MountInfo struct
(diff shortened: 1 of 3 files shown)


dipack95:
@pgier Can you take a look at this?

mknapphrt:
This looks good to me but it's up to @pgier or @SuperQ to give a more authoritative review

mountinfo.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse parent ID")
+	}
+	// Has optional fields, which is a space separated list of values
+	// Example: shared:2 master:7
+	if len(beforeFields) > 6 {
+		mount.OptionalFields = make(map[string]string)
+		optionalFields := beforeFields[6:]
+		for _, field := range optionalFields {
+			optionSplit := strings.Split(field, ":")
```
pgier: Since Split/SplitN should always return at least length 1, I think you can simplify this section a bit to something like this.
```golang
optionSplit := strings.SplitN(field, ":", 2)
value = ""
if len(optionSplit) == 2 {
	value = optionSplit[1]
}
if isValidOptionalField(optionSplit[0]) {
	mount.OptionalFields[optionSplit[0]] = value
}
```

dipack95:
@pgier Thanks for the feedback! I've modified the code to bring them in line with your suggestions.

mountinfo.go
```diff
@@ -0,0 +1,171 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ: New files should get the current year. Please update this to 2019.

mountinfo.go
```diff
@@ -0,0 +1,171 @@
...
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+)
+
+var validOptionalFields = map[string]bool{"shared": true, "master": true, "propagate_from": true, "unbindable": true}
```
SuperQ: Please break up this long line.

```
var (
  validOptionalFields = map[string]bool{
    "shared": true,
...
```

mountinfo.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Fields description for mountinfo is here:
+// http://man7.org/linux/man-pages/man5/proc.5.html
+
+type MountInfo struct {
+	// Unique Id for the mount
```
SuperQ: We try and have comments be full sentences. Starting with a capitol letter and ending a `.`.

dipack95:
@SuperQ I've made the changes as you requested! Thanks for the feedback.

mountinfo_test.go
```diff
@@ -0,0 +1,135 @@
+// Copyright 2018 The Prometheus Authors
```
SuperQ: Missed another date change.

dipack95: Yup, my bad. Just fixed it.

dipack95:
@SuperQ Do you think the PR is in an acceptable state right now?

dipack95:
@SuperQ Gentle reminder! :)

proc.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Prometheus Authors
+// Copyright 2019 The Prometheus Authors
```
pgier: I think this copyright date should remain 2018 since the file already existed before this change.

dipack95: Okay, I'll change it now

dipack95:
@pgier I am currently using the changes in this PR in my own fork of `node_exporter` (by replacing the `procfs` module in the `vendor/` folder), and would like to create a PR for that too. 

I was hoping that could create a new release of `procfs` with these changes included, so that there wouldn't be an issue with the modules in `node_exporter`. Are you planning to do so?

pgier:
@dipack95 I'd like to get a couple more PRs merged and then I'll tag a new release probably some time next week.

Issue #157, opened 2019-05-14
Support for /proc/sys/vm


stevenjm:
Hi,

I raised the suggestion of having the node exporter expose metrics from `/proc/sys/vm` [on the mailing list](https://groups.google.com/forum/?fromgroups#!topic/prometheus-developers/iXzsjy0NFAE) and was suggested to raise an issue here.

For my specific use case, exposing `/proc/sys/vm/max_map_count` is sufficient. There are also other files in `/proc/sys/vm` that we might consider exposing.

I'm happy to work on coding this once there's consensus on the approach.

Regards,
Steven

pgier:
@stevenjm Feel free to submit a PR for this.  This can probably be implemented similarly to some of the functions in the `sysfs` package, which also reads a lot of small files with one value each.

TheMeier:
I'd like to take that one.

PR #176, opened 2019-06-10
procfs: add parsing of sys/vm

TheMeier:
Read data from /proc/sys/vm entries
Closes: #157

Theres still work to be done for `lowmem_reserve_ratio`, actually i don't understand the format yet fully especially when dealing with NUMA zones.

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -619,6 +619,227 @@ Path: fixtures/proc/symlinktargets/xyz
 Lines: 0
 Mode: 644
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/proc/sys
+Mode: 775
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/proc/sys/vm
+Mode: 775
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/admin_reserve_kbytes
+Lines: 1
+8192
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/block_dump
+Lines: 1
+0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/compact_unevictable_allowed
+Lines: 1
+1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirty_background_bytes
+Lines: 1
+0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirty_background_ratio
+Lines: 1
+10
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirty_bytes
+Lines: 1
+0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirty_expire_centisecs
+Lines: 1
+3000
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirty_ratio
+Lines: 1
+20
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirty_writeback_centisecs
+Lines: 1
+500
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/dirtytime_expire_seconds
+Lines: 1
+43200
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/drop_caches
+Lines: 1
+0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/extfrag_threshold
+Lines: 1
+500
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/hugetlb_shm_group
+Lines: 1
+0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/laptop_mode
+Lines: 1
+5
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/legacy_va_layout
+Lines: 1
+0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/lowmem_reserve_ratio
+Lines: 1
+256	256	32	0	0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/sys/vm/max_map_count
+Lines: 1
+65530
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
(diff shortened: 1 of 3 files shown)


vm.go
```diff
@@ -0,0 +1,131 @@
...
+	interfaceClass := Vm{}
+	interfaceElem := reflect.ValueOf(&interfaceClass).Elem()
+	interfaceType := reflect.TypeOf(interfaceClass)
+
+	//start from 1 - skip the Name field
+	for i := 0; i < interfaceElem.NumField(); i++ {
+		fieldType := interfaceType.Field(i)
+		fieldValue := interfaceElem.Field(i)
+
+		value, err := util.SysReadFile(devicePath + "/" + fieldType.Tag.Get("fileName"))
```
pgier: Use filepath.Join instead of direct string concatenation for the path.

vm.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// parseVm scans predefined files in /proc/sys/vm
+// directory and gets their contents.
+func parseVm(devicePath string) (Vm, error) {
+	interfaceClass := Vm{}
+	interfaceElem := reflect.ValueOf(&interfaceClass).Elem()
+	interfaceType := reflect.TypeOf(interfaceClass)
+
+	//start from 1 - skip the Name field
```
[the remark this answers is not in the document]
TheMeier: copy&paste artefact, removed

vm.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// The VM interface is described at
+//   https://www.kernel.org/doc/Documentation/sysctl/vm.txt
+// Each setting is exposed as a single file.
+// Each file contains one line with a single numerical value, except lowmem_reserve_ratio which holds an array
+// and numa_zonelist_order (deprecated) which is a string
+type Vm struct {
```
mdlayher: *VM

Same throughout. Go style is that acronyms/initialisms are all caps.

vm.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	return parseVm(path)
+}
+
+// parseVm scans predefined files in /proc/sys/vm
+// directory and gets their contents.
+func parseVm(devicePath string) (Vm, error) {
+	interfaceClass := Vm{}
+	interfaceElem := reflect.ValueOf(&interfaceClass).Elem()
```
mdlayher: I would really prefer to not see reflection here either. The data isn't truly dynamic, and you've already had to type out all the file names above.

IMHO this really should be a switch statement.

vm.go
```diff
@@ -0,0 +1,131 @@
...
+	interfaceType := reflect.TypeOf(interfaceClass)
+
+	for i := 0; i < interfaceElem.NumField(); i++ {
+		fieldType := interfaceType.Field(i)
+		fieldValue := interfaceElem.Field(i)
+
+		value, err := util.SysReadFile(filepath.Join(devicePath, fieldType.Tag.Get("fileName")))
+
+		if err != nil {
+			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
```
mdlayher: What's going on here with these other error conditions?

vm_test.go
```diff
@@ -0,0 +1,121 @@
...
+	fs, err := NewFS(procTestFixtures)
+	if err != nil {
+		t.Fatal(err)
+	}
+	stats, err := fs.Vm()
+	if err != nil {
+		t.Fatal(err)
+	}
+	var (
+		adminReserveKbytes        int64 = 8192
```
mdlayher: Remove this var block and move all of these non-zero values into the struct declaration below to cut a lot of redundancy. You can omit any zero values from the struct declaration as they are inferred as the zero value of int64 (0) anyway.

vm.go
```diff
@@ -0,0 +1,131 @@
...
+	DirtyExpireCentisecs      int64 `fileName:"dirty_expire_centisecs"`
+	DirtyRatio                int64 `fileName:"dirty_ratio"`
+	DirtytimeExpireSeconds    int64 `fileName:"dirtytime_expire_seconds"`
+	DirtyWritebackCentisecs   int64 `fileName:"dirty_writeback_centisecs"`
+	DropCaches                int64 `fileName:"drop_caches"`
+	ExtfragThreshold          int64 `fileName:"extfrag_threshold"`
+	HugetlbShmGroup           int64 `fileName:"hugetlb_shm_group"`
+	LaptopMode                int64 `fileName:"laptop_mode"`
+	LegacyVaLayout            int64 `fileName:"legacy_va_layout"`
+	// LowmemReserveRatio        [?]int64 `fileName:"lowmem_reserve_ratio"`
```
mdlayher: Looks like this can be parsed into a Go slice.

vm.go
TheMeier: Ok, will rewrite this. The motivation was to follow the style used elswhere, as it was pointed out in the issue  

> similarly to some of the functions in the sysfs package

vm_test.go
TheMeier: right, also a copy&paste artefact

vm.go
TheMeier: as I wrote already I am not quite sure what possible formats this can have. What I have seen so far are 3 or 5 values, but it seems depending on numa zones this could be much larger?

vm.go
mdlayher: I see, I'd like to get rid of reflect in this repository and have filed: https://github.com/prometheus/procfs/issues/177.

vm.go
mdlayher: Seems reasonable to me to parse whatever's in here as `[]int64`.

vm.go
TheMeier: was following the style in class_power_supply.go and net_class.go

vm.go
TheMeier: Hm after looking around a bit apart from reflect the only solution to iterate over a struct seems to be something like

``` for _, d := range []interface{}{instance.field1, instance.field2, instance.field3} {...```

That doesn't look very sexy either, I just wait and see what you come up with in #177 ;)

File: fixtures.ttar [hunk 8]
```diff
@@ -1806,3 +2027,221 @@ Lines: 1
 extent_alloc 2 0 0 0
 Mode: 644
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/vm
```
pgier: This seems to be a duplicate of what's in `/proc/sys/vm`, is this separate directory structure needed?

vm.go
```diff
@@ -0,0 +1,150 @@
+// Copyright 2018 The Prometheus Authors
```
pgier: Should be 2019

vm_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2018 The Prometheus Authors
```
pgier: 2019

fixtures.ttar
TheMeier: fixed

TheMeier:
@pgier i changed the MR using the same pattern. I used your util ValueParser.Pint64 although I don't use Pointers here, I wonder if that should be changed also.

pgier:
The fixtures file ordering should be fixed by #185

vm.go
```diff
@@ -0,0 +1,210 @@
...
+	Swappiness                int64   // /proc/sys/vm/swappiness
+	UserReserveKbytes         int64   // /proc/sys/vm/user_reserve_kbytes
+	VfsCachePressure          int64   // /proc/sys/vm/vfs_cache_pressure
+	WatermarkBoostFactor      int64   // /proc/sys/vm/watermark_boost_factor
+	WatermarkScaleFactor      int64   // /proc/sys/vm/watermark_scale_factor
+	ZoneReclaimMode           int64   // /proc/sys/vm/zone_reclaim_mode
+}
+
+// VM reads the VM statistics from the specified `proc` filesystem.
+func (fs FS) VM() (VM, error) {
```
pgier: I think it would be better to use a `*VM` for the return type so that the struct doesn't need to be copied.

TheMeier: Hm everything else returns a struct and not a pointer to struct. Why would VM be different?

pgier: If you look at `class_power_supply.go` for example, `parsePowerSupply` returns a struct pointer, and `PowerSupplyClass()` returns a map which is also a reference type.  However, there seems to be a mix of both pointer and non-pointer return types for different functions.  In general, if a struct has more than a few fields then I'd tend to use a pointer, but maybe that's a premature optimization, so I'm ok with it if you prefer to leave as a struct.

TheMeier: changed it to return `*VM`

TheMeier:
@pgier hm fair question the documentation at https://www.kernel.org/doc/Documentation/sysctl/vm.txt doesn't really say anything about data format. 
Looking around a bit in the kernel source i found a bunch of the settings defined as `CTL_INT` https://github.com/torvalds/linux/blob/f1c2f8857c5aa6c92aa903bc06437503422e5925/kernel/sysctl_binary.c#L147-L174 which is a define to bin_intvec which returns ssize_t so those should be fine.
Some others i randomly checked are defined as int ot u64.
My intuition is it should be quite safe, but to really answer the question someone (me?) should check the definition of each setting in the kernel sources.

TheMeier:
> LGTM. Although after thinking about possible nil pointers similar to #187, I'm wondering if we should be using pointer types for the int64 values. @TheMeier what are the chances of some of the sys/vm files containing non-integer values and causing nil pointer dereference? @mdlayher wdyt?

Actually I just realized I should have used pointers here. In case the valueparser comes back with nil the current implementation will lead to:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

Will open an issue and patch...

PR #180, opened 2019-06-14
change VSize from int to uint64

codeskyblue:
VSize might > MAX(int) on 32bit platform.


pgier:
Did you see an overflow on a particular platform?  I would think the existing `uint` and `int` would adjust appropriately for 32-bit vs 64-bit systems.

Also each commit needs a signoff (git commit --signoff)

codeskyblue:
There is uint overflow on Android system

codeskyblue:
Change to another merge request. https://github.com/prometheus/procfs/pull/182

PR #181, opened 2019-06-14
Fixed edge cases, added more tests, and extracted failed/spare device information from mdstat

You-NeverKnow:
This pull request aims to add functionality as required in [this issue](https://github.com/prometheus/node_exporter/issues/261) in [this comment](https://github.com/prometheus/node_exporter/issues/261#issuecomment-500433010)

This PR modifies the MdStat struct to export failed device count and spare device count, so that node_exporter could expose them. It also adds several edge test cases.  

- Added code to extract Failed and Spare device counts from `/proc/mdstat`
- Refactored some variable names to be more descriptive.
- Added code to extract "recovering" or "resyncing" state from `/proc/mdstat` 
- Added more test cases from `https://github.com/prometheus/node_exporter/mdadm_linux_test.go`
- Added a test case to check for a `read-only raid0 device`: md101 in fixtures.ttar file. 

@discordianfish @pgier

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -3,6 +3,9 @@ Directory: fixtures
 Mode: 775
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 Directory: fixtures/proc
+Mode: 775
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/proc/invalid
 Mode: 755
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 Directory: fixtures/proc/26231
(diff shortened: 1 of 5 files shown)


mdstat.go
```diff
@@ -59,50 +63,77 @@ func (fs FS) MDStat() ([]MDStat, error) {
...
+			strings.HasPrefix(line, "Personalities") ||
+			strings.HasPrefix(line, "unused") {
 			continue
 		}
 
-		deviceFields := strings.Fields(l)
+		deviceFields := strings.Split(line, " ")
 		if len(deviceFields) < 3 {
-			return nil, fmt.Errorf("not enough fields in mdline (expected at least 3): %s", l)
+			return nil, fmt.Errorf("error parsing mdline: %s", line)
```
pgier: This seems to make the error message less informative than it was previously.

mdstat.go
```diff
@@ -59,50 +63,77 @@ func (fs FS) MDStat() ([]MDStat, error) {
...
+
+	for i, line := range lines {
+		if strings.TrimSpace(line) == "" || line[0] == ' ' ||
+			strings.HasPrefix(line, "Personalities") ||
+			strings.HasPrefix(line, "unused") {
 			continue
 		}
 
-		deviceFields := strings.Fields(l)
+		deviceFields := strings.Split(line, " ")
```
pgier: What's the reason to use `Split` instead of `Fields`?  More efficient?

mdstat.go
```diff
@@ -112,39 +143,52 @@ func parseMDStat(mdstatData []byte) ([]MDStat, error) {
 	return mdStats, nil
 }
 
-func evalStatusLine(statusline string) (active, total, size int64, err error) {
-	matches := statuslineRE.FindStringSubmatch(statusline)
-	if len(matches) != 4 {
-		return 0, 0, 0, fmt.Errorf("unexpected statusline: %s", statusline)
-	}
+func evalStatusLine(lines []string, i int) (active, total, size int64, err error) {
```
pgier: I think this would be better if it took two string parameters like `deviceLine` and `statusLine` to make it more readable and avoid any potential index out of bounds.

fixtures.ttar
```diff
@@ -355,33 +358,72 @@ xpc 399724544 92823103 86219234
 debug 0
 Mode: 644
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/invalid/mdstat
```
pgier: I'd prefer to keep the contents of `fixtures/proc` as close to a real system as possible.  Can you move this test example in-line with the test sources as a multi-line string and then call `parseMDStat` directly?

mdstat_test.go
```diff
@@ -13,24 +13,39 @@
...
 
-import (
-	"testing"
-)
+import "testing"
+
+func TestMdadm(t *testing.T) {
+	mountPoint := "fixtures/proc"
+	fs, errFs := NewFS(mountPoint)
+
```
pgier: Why not just use the existing `getProcFixtures(t)` here instead?  If you want a better error message, feel free to change `getProcFixtures(t)` directly.

mdstat.go
You-NeverKnow: Same reason as before: started with node_exporter/mdamd_linux.go as base, and it had Split(), so stuck with it.

mdstat.go
You-NeverKnow: You're right! I didn't give it much thought. I started editing node_exporter/mdadm_linux.go as my base, and so didn't pay much attention to lack of verbosity of error message. I'll fix it.

mdstat.go
You-NeverKnow: Good point

fixtures.ttar
You-NeverKnow: Sure! Onto it

mdstat_test.go
You-NeverKnow: I somehow forgot it existed. I'll use that instead, and change the error message there.

You-NeverKnow:
@pgier How about now?

fs_test.go
```diff
@@ -33,7 +33,7 @@ func TestNewFS(t *testing.T) {
 func getProcFixtures(t *testing.T) FS {
 	fs, err := NewFS(procTestFixtures)
 	if err != nil {
-		t.Fatal(err)
+		t.Errorf("Creating pseudo fs from getProcFixtures failed at fixtures/proc")
```
pgier: Should probably include the original error in the message string to make it easier to tell what is the root cause.

You-NeverKnow: Got it.

You-NeverKnow:
@pgier Hey I was wondering when v0.0.3 would come out for this project. I would like to use these changes to make another PR in node_exporter.

PR #182, opened 2019-06-17
change vss and rss from int to uint64

codeskyblue:
Signed-off-by: codeskyblue <codeskyblue@gmail.com>

vss will overflow on Android system


pgier:
This could be caused by a difference between the compile target (32-bit) and the runtime platorm (64-bit).  Are you compiling your app using `GOARCH=arm64`?

PR #183, opened 2019-06-19
Supported for /proc/<pid>/environ

takumakume:
Supported for `/proc/<pid>/environ`

```sh
# cat -e /proc/1/environ
PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin^@HOSTNAME=33cfa6747903^@TERM=xterm^@GOLANG_VERSION=1.12.5^@GOPATH=/go^@HOME=/root^@
```

- Split `\000` .
- Remove last empty array element.
- About the format of the return value is the slice `[]string` : It uses the same interface as [os.Environ](https://golang.org/pkg/os/#Environ).

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -21,6 +21,11 @@ Mode: 644
 Path: fixtures/proc/26231/cwd
 SymlinkTo: /usr/bin
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/26231/environ
+Lines: 1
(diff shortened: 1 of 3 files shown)


takumakume:
@pgier

I am trying to create a procfs dumper.
It is used, for example, to collect information on processes at the time of an incident occurrence.

I want to use this library for the software I am developing.
In the process, I think that I will send another PR when the information that needs to be collected increases.

Thank you.

PR #184, opened 2019-06-20
Add sysfs handle and remove path from Get backstore function 

AvengerMoJo:
I do need to export configfs path in lio collector module, so instead of exporting the configfs all together I export the Path functions to help building configfs path. 

Signed-off-by: Alex Lau (AvengerMoJo) <alau@suse.com>

### DIFF
--- a/iscsi/get.go
+++ b/iscsi/get.go
@@ -137,14 +137,13 @@ func ReadWriteOPS(iqnPath string, tpgt string, lun string) (readmb uint64,
 
 // GetFileioUdev is getting the actual info to build up
 // the FILEIO data and match with the enable target
-func (fileio FILEIO) GetFileioUdev(targetCorePath string, fileioNumber string,
-	objectName string) (fio *FILEIO, err error) {
-
-	fileio.Name = "fileio_" + fileioNumber
-	fileio.Fnumber = fileioNumber
-	fileio.ObjectName = objectName
-
-	udevPath := filepath.Join(targetCorePath, fileio.Name, fileio.ObjectName, "udev_path")
+func (fs FS) GetFileioUdev(fileioNumber string, objectName string) (*FILEIO, error) {
+	fileio := FILEIO{
+		Name:       "fileio_" + fileioNumber,
+		Fnumber:    fileioNumber,
+		ObjectName: objectName,
+	}
+	udevPath := fs.configfs.Path(targetCore, fileio.Name, fileio.ObjectName, "udev_path")
 
 	if _, err := os.Stat(udevPath); os.IsNotExist(err) {
 		return nil, fmt.Errorf("iscsi: GetFileioUdev: fileio_%s is missing file name", fileio.Fnumber)
@@ -160,14 +159,13 @@ func (fileio FILEIO) GetFileioUdev(targetCorePath string, fileioNumber string,
 
 // GetIblockUdev is getting the actual info to build up
 // the IBLOCK data and match with the enable target
-func (iblock IBLOCK) GetIblockUdev(targetCorePath string, iblockNumber string,
-	objectName string) (ib *IBLOCK, err error) {
-
-	iblock.Name = "iblock_" + iblockNumber
-	iblock.Bnumber = iblockNumber
-	iblock.ObjectName = objectName
-
-	udevPath := filepath.Join(targetCorePath, iblock.Name, iblock.ObjectName, "udev_path")
+func (fs FS) GetIblockUdev(iblockNumber string, objectName string) (*IBLOCK, error) {
+	iblock := IBLOCK{
+		Name:       "iblock_" + iblockNumber,
+		Bnumber:    iblockNumber,
+		ObjectName: objectName,
+	}
+	udevPath := fs.configfs.Path(targetCore, iblock.Name, iblock.ObjectName, "udev_path")
 
 	if _, err := os.Stat(udevPath); os.IsNotExist(err) {
 		return nil, fmt.Errorf("iscsi: GetIBlockUdev: iblock_%s is missing file name", iblock.Bnumber)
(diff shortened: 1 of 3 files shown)


AvengerMoJo:
@pgier sorry to trouble you again for since my node_exporter module do need the path from the configfs. 
https://github.com/prometheus/node_exporter/pull/1386 
So I need to add this Path function or exporter configfs as Configfs as before. But I believe we don't want to do that

pgier:
I'm guessing this is related to using `GetFileioUdev` and `GetIblockUdev`?  I'm wondering if these could be re-factored in a different way.  I didn't notice before that these methods basically use an empty struct and then return a pointer to the same struct.  I'm thinking a better way for these functions to work is to make them methods of FS instead of IBLOCK, FILEIO, etc, so you can use the configfs mount point from the FS struct and build the targetcore path inside the function instead of passing it as a parameter.  Does that make sense?

AvengerMoJo:
I believe I understand what you mean but I don't know I have the solution. 
If I remember correctly I was using static path in the collector module before then we need procfs module to pull out the path. Then the test case require to setup non-standard system path. So I need to start this configfs path create it at run time or else I can't finish the fixture test case. 
Therefore ... If I don't pass it in parameter, I can't do fixture test. Or the test case mount point get to be the same as system and fix.

AvengerMoJo:
Oh I see what you did different ... you put the function assign to the fs. Instead ... so the maybe I can move all the function away. Let me try.

pgier:
I don't think it matters too much whether the functions are in `iscsi.go` or `get.go`, you can organize them in whatever way makes the most sense to you.

iscsi/get.go
```diff
@@ -137,13 +137,14 @@ func ReadWriteOPS(iqnPath string, tpgt string, lun string) (readmb uint64,
...
-
+// func (fileio FILEIO) GetFileioUdev(fileioNumber string, objectName string) (fio *FILEIO, err error) {
+func (fs FS) GetFileioUdev(fileioNumber string, objectName string) (*FILEIO, error) {
+	targetCorePath := fs.configfs.Path(TargetCore)
+	fileio := FILEIO{
+		Name:       "fileio_" + fileioNumber,
+		Fnumber:    fileioNumber,
+		ObjectName: objectName,
+	}
 	udevPath := filepath.Join(targetCorePath, fileio.Name, fileio.ObjectName, "udev_path")
```
pgier: I think this can be simplified a little bit and remove the need for `targetCorePath`.
```golang
	udevPath := fs.configfs.Path(TargetCore, fileio.Name, fileio.ObjectName, "udev_path")
```

iscsi/iscsi.go
```diff
@@ -30,6 +30,10 @@ const IqnGlob = "target/iscsi/iqn*"
 // reading runtime status
 const TargetCore = "target/core"
 
+// DevicePath static path /sys/devices/rbd/[0-9]* for rbd devices to
+// read at runtime status
+const DevicePath = "devices/rbd"
```
pgier: Are `DevicePath` and `TargetCore` still needed outside this package?  Or can they be changed to `devicePath` and `targetCore`?

iscsi/get.go
```diff
@@ -183,12 +184,17 @@ func (iblock IBLOCK) GetIblockUdev(targetCorePath string, iblockNumber string,
...
+	sysDevicePath := fs.configfs.Path(DevicePath)
+	rbd := RBD{
+		Name:    "rbd_" + rbdNumber,
+		Rnumber: rbdNumber,
+		Pool:    "",
+		Image:   "",
+	}
 
-	systemRbds, err := filepath.Glob(filepath.Join(sysDevicePath, "/devices/rbd/[0-9]*"))
+	systemRbds, err := filepath.Glob(filepath.Join(sysDevicePath, "[0-9]*"))
```
pgier: This can be 
```golang
systemRbds, err := filepath.Glob(fs.configfs.Path(DevicePath, "[0-9]*"))
```

iscsi/get.go
AvengerMoJo: You are right, I'm rewriting that right now.

iscsi/iscsi.go
AvengerMoJo: yup, I update all the static path and pushed.

iscsi/get.go
AvengerMoJo: Thanks , done.

iscsi/iscsi.go
```diff
@@ -23,12 +23,16 @@ import (
...
 
-// TargetCore static path /sys/kernel/config/target/core for node_exporter
+// targetCore static path /sys/kernel/config/target/core for node_exporter
 // reading runtime status
-const TargetCore = "target/core"
+const targetCore = "target/core"
+
+// devicePath static path /sys/devices/rbd/[0-9]* for rbd devices to
+// read at runtime status
+const devicePath = "../../devices/rbd"
```
pgier: Should this one be relative to sysfs (`/sys`) instead of configfs (`/sys/kernel/config`)?  A user could potentially have sysfs mounted to a different location, and that would break this relative path.

AvengerMoJo: Yes I agree. However the hard part is how to "remember" the sys path and configfs path for fixture test.  Because we only setup path Once in the NewFS, so I either use relative path for /sys or relative path for "/sys/kernel/config". Or I can update NewFS construct with two different path /sys and kernel/config with two variables.

iscsi/get.go
```diff
@@ -183,12 +181,14 @@ func (iblock IBLOCK) GetIblockUdev(targetCorePath string, iblockNumber string,
...
-	rbd.Name = "rbd_" + rbdNumber
-	rbd.Rnumber = rbdNumber
-
-	systemRbds, err := filepath.Glob(filepath.Join(sysDevicePath, "/devices/rbd/[0-9]*"))
+func (fs FS) GetRBDMatch(rbdNumber string, poolImage string) (*RBD, error) {
+	rbd := RBD{
+		Name:    "rbd_" + rbdNumber,
+		Rnumber: rbdNumber,
+		Pool:    "",
+		Image:   "",
```
pgier: Since `Pool` and `Image` are empty strings, you can leave them out of the struct initialization.

pgier:
Looks good, would you mind squashing the commits and updating the log message?

iscsi/get.go
AvengerMoJo: Sure, I will do that and squashing it down for the PR. Thanks

pgier:
Ah, one more change, the commit message still says `Add helper function for configfs export Path`, which no longer describes this commit :)

pgier:
@AvengerMoJo I mean the commit itself still has the old title.  I think you need to do a `git commit --amend` and then another force push.

AvengerMoJo:
You are too fast lol :) I was in the middle of doing the amend 👍

pgier:
LOL, I guess I noticed the PR title change at just the wrong time :)

PR #186, opened 2019-06-25
Parse task queue stats from /proc/schedstat

bitglue:
These are useful as an indication of CPU contention and scheduler
latency.


bitglue:
@discordianfish @pgier ready for review.

I'll update https://github.com/prometheus/node_exporter/pull/1389 to use this after review.

schedstat.go
```diff
@@ -0,0 +1,91 @@
...
+type Schedstat struct {
+	CPUs []*SchedstatCPU
+}
+
+// SchedstatCPU contains the values from one "cpu<N>" line
+type SchedstatCPU struct {
+	CPUNum string
+
+	RunningJiffies uint64
+	RunningSeconds float64
```
pgier: What do you think about adding a method to dynamically provide the seconds instead of storing both jiffies and seconds in the struct?

```golang
func (s SchedstatCPU) RunningSeconds() float64 {
	return float64(s.RunningJiffies) / userHZ
}
```

schedstat.go
```diff
@@ -0,0 +1,91 @@
...
+func (fs FS) Schedstat() (*Schedstat, error) {
+	file, err := os.Open(fs.proc.Path("schedstat"))
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	stats := &Schedstat{}
+	scanner := bufio.NewScanner(file)
+
```
pgier: Would it be helpful to check the version number at the top of the file, or is the format consistent enough between versions that it's not necessary?

bitglue: I think there would be little practical value in doing so. Version 15 is current, and according to the documentation:

> Version 15 of schedstats dropped counters for some sched_yield:
yld_exp_empty, yld_act_empty and yld_both_empty. Otherwise, it is
identical to version 14.

Version 14 seems to cover kernels at least 10 years back, and is compatible. So to support any older versions would be a bit of an archaeology project.

Newer versions may not remain compatible, but probably more likely than not they will be compatible enough for the subset of metrics extracted here.

pgier: Ok, sounds good.

bitglue:
@pgier Everything addressed I think, please re-review

bitglue:
@discordianfish @pgier bump, review please

schedstat.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return stats, nil
+}
+
+func parseProcSchedstat(filename string) (ProcSchedstat, error) {
+	stats := ProcSchedstat{}
+
+	file, err := os.Open(filename)
```
pgier: I think it would be better to use `ioutil.ReadFile`, then you don't need the loop and scanner.

bitglue: good idea, one minute...

bitglue:
@pgier ok, sorry to change things just after asking for review. It occured to me reading the diff with fresh eyes and a cup of coffee that there were untested corner cases. I've added some more tests and rearranged some things. Should be good now, for reals.

pgier:
Merged, thanks!

bitglue:
What's the process for getting a released tagged so I can complete https://github.com/prometheus/node_exporter/pull/1389 ?

bitglue:
Actually now that I've learned a bit more about go modules I've figured out how to update it without a release. Seems I'm not the first one to pin node_exporter to a non-released version of procfs.

PR #187, opened 2019-06-26
Fix crash if port_rcv_data contains "N/A (no PMA)"

bdrung:
In case `port_rcv_data` contains `N/A (no PMA)`, `vp.PUInt64` returns `nil` which is assigned to `counters.PortRcvData`. Then the content of `counters.PortRcvData` cannot be multiplied by four. The program will fail with:

```
```

This issue was found by the tests in https://github.com/prometheus/node_exporter/pull/1396


discordianfish:
Wondering if we could handle that better in the value parser but possibly not, so LGTM either way.

bdrung:
Alternatively the value parser could accept a function that it applies on the value before returning it.

PR #191, opened 2019-07-13
Fix seconds reported by schedstat

bitglue:
As explained further in the code comments, the previous implementation
was based on incorrect documentation and reported values 10000000 times
too large for the values in seconds.

Signed-off-by: Phil Frost <phil@postmates.com>

### DIFF
--- a/schedstat.go
+++ b/schedstat.go
@@ -31,6 +31,12 @@ var (
 // See
 // https://www.kernel.org/doc/Documentation/scheduler/sched-stats.txt
 // for a detailed description of what these numbers mean.
+//
+// Note the current kernel documentation claims some of the time units are in
+// jiffies when they are actually in nanoseconds since 2.6.23 with the
+// introduction of CFS. A fix to the documentation is pending. See
+// https://lore.kernel.org/patchwork/project/lkml/list/?series=403473
+
 type Schedstat struct {
 	CPUs []*SchedstatCPU
 }
@@ -39,16 +45,16 @@ type Schedstat struct {
 type SchedstatCPU struct {
 	CPUNum string
 
-	RunningJiffies uint64
-	WaitingJiffies uint64
-	RunTimeslices  uint64
+	RunningNanoseconds uint64
+	WaitingNanoseconds uint64
+	RunTimeslices      uint64
 }
 
 // ProcSchedstat contains the values from /proc/<pid>/schedstat
 type ProcSchedstat struct {
-	RunningJiffies uint64
-	WaitingJiffies uint64
-	RunTimeslices  uint64
+	RunningNanoseconds uint64
+	WaitingNanoseconds uint64
+	RunTimeslices      uint64
 }
 
 func (fs FS) Schedstat() (*Schedstat, error) {
@@ -67,12 +73,12 @@ func (fs FS) Schedstat() (*Schedstat, error) {
 			cpu := &SchedstatCPU{}
 			cpu.CPUNum = match[1]
 
-			cpu.RunningJiffies, err = strconv.ParseUint(match[8], 10, 64)
+			cpu.RunningNanoseconds, err = strconv.ParseUint(match[8], 10, 64)
 			if err != nil {
 				continue
 			}
 
-			cpu.WaitingJiffies, err = strconv.ParseUint(match[9], 10, 64)
+			cpu.WaitingNanoseconds, err = strconv.ParseUint(match[9], 10, 64)
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


bitglue:
@discordianfish @pgier simple bugfix of recently added functionality, ready for review.

bitglue:
@discordianfish @pgier bump

bitglue:
@discordianfish @pgier rebump, easy bug fix, needs review.

bitglue:
@discordianfish @pgier bump bump bump bump bump bump bump bump bump bump bump

SuperQ:
Hrm, I really think we need some additional testing here to verify what's correct, and when exactly things changed in the kernel. We still support very old kernels, like RHEL 6 contains 2.6.32.

SuperQ:
I ran a quick test, booting up a CentOS 6 box. It looks like it's been in NS for a long time.

```console
[bjk@bjk-centos-6 ~]$ cat /proc/schedstat 
version 15
timestamp 4294786116
cpu0 0 0 65279 8841 37792 37792 80015632879 18335798709 56073
[bjk@bjk-centos-6 ~]$ uptime
 07:47:20 up 2 min,  1 user,  load average: 0.72, 0.38, 0.14
[bjk@bjk-centos-6 ~]$ uname -a
Linux bjk-centos-6 2.6.32-754.15.3.el6.x86_64 #1 SMP Tue Jun 18 16:25:32 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

Field 7 indicates ~80.015 seconds of CPU time used with NS, and an absurd number if it were in Jiffies.

schedstat.go
```diff
@@ -112,17 +125,17 @@ func parseProcSchedstat(contents string) (stats ProcSchedstat, err error) {
 }
 
 func (stat *SchedstatCPU) RunningSeconds() float64 {
-	return float64(stat.RunningJiffies) / userHZ
+	return float64(stat.RunningNanosec) / 1000000000
```
SuperQ: Minor Go style nit. I would prefer to define `1000000000` as `const nanoSeconds = 1000000000`.

bitglue: even better, I could just use `time.Duration`. It's defined as `int64`, whereas the kernel defines the counters as `unsigned long long`. So after about 290 years of uptime, there will be an overflow problem. But that seems practically unlikely. Well before that, there will be issues with float precision, so I guess anyone wanting precise metrics with prometheus will need to reboot their servers every 100 years or so.

SuperQ: No, we've run into issues using `time.Duration` with this library previously[0].

Actually, want I'd like to do is follow the same pattern. Have the function return `RunningNanoseconds` as the raw uint64, and leave it up to the user (node_exporter) to convert from uint64 nanoseconds into float64 seconds.

[0]: https://github.com/prometheus/procfs/pull/165

schedstat.go
```diff
@@ -39,16 +53,16 @@ type Schedstat struct {
 type SchedstatCPU struct {
 	CPUNum string
 
-	RunningJiffies uint64
-	WaitingJiffies uint64
-	RunTimeslices  uint64
+	Running       time.Duration
```
SuperQ: Thinking about this more, let's keep this as `RunningNanoseconds uint64`.

bitglue:
The stats went from jiffies to ns in https://github.com/torvalds/linux/commit/425e0968a25fa3f111f9919964cac079738140b5

The log for that commit says "no code changes are caused by this patch." but I don't think that's true. Look at the implementation of `sched_info_arrive()`; the old implementation has

    unsigned long now = jiffies, delta_jiffies = 0;

and the new implementation

    unsigned long long now = sched_clock(), delta = 0;

[`sched_clock()` returns time in nanoseconds](https://github.com/torvalds/linux/blob/425e0968a25fa3f111f9919964cac079738140b5/arch/i386/kernel/tsc.c#L100-L119).

This patch is part of the CFS implementation, which was first released in 2.6.23-rc1. Prior to that, stats were in jiffies. After that, stats were in ns.

Unfortunately, `SCHEDSTAT_VERSION` is set to 14 in either case. Between this, the commit message above ("no code changes"), and the documentation that was never updated, I guess this is a backwards-incompatible change which was introduced unintentionally.

So we don't have a straightforward way to know how to parse schedstats. But we could perhaps implement some heuristic, for example compare the counters to the `timestamp` field (which in all cases, does actually seem to be in jiffies) and if the counters are absurdly larger than `timestamp`, then assume they are in ns. If they are all less than `timestamp`, jiffies.

Or we can just forget about it, and people running kernels older than 2.6.23 can just have wrong metrics.

schedstat.go
bitglue: guh....why?

SuperQ: Because the job of this library is to provide access to the data, not interpret it.

SuperQ:
Well, good news, Go doesn't support older than 2.6.23, so we're safe to assume the nanosecond behavior. 😁

schedstat.go
bitglue: Should the variable names not contain "nanoseconds" then, since they might be jiffies? What if the C compiler used to build the kernel has an `unsigned long long` that's longer than 64 bits?

SuperQ: Hopefully Go will support longer than uint64 by then.

PR #196, opened 2019-07-15
Adds /sys/class/thermal/cooling_device* support

davemcphee:
Adds support for `/sys/class/thermal/cooling_device[0-9]+` devices, returning name, type, cur_state and max_state. 

The latter 2 are int64, which differs from most values in this library which appear to be uint64 - this is due to the type `intel_powerclamp` returning a -1 as it's `cur_value` when idle injection is disabled.

### DIFF
--- a/sysfs/class_cooling_device.go
+++ b/sysfs/class_cooling_device.go
@@ -1,0 +1,91 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


sysfs/class_cooling_device.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		coolingDeviceStats.Name = cdName
+		stats[i] = coolingDeviceStats
+	}
+	return stats, nil
+}
+
+func parseCoolingDeviceStats(cd string) (ClassCoolingDeviceStats, error) {
+	// get cd type
```
SuperQ: Please make comments complete sentences.

This comment seems a little unnecessary.

davemcphee: whoops, overlooked that one. Removed.

PR #197, opened 2019-07-27
Collect additional XFS statistics

skreuzer:
Include the xstrat, push_ail, abtb2, bmbt2, ibt2, qm, and debug
runtime statistics when parsing the XFS stats file.


discordianfish:
This broke the node-exporter for people with different kernel configs because of the field length checks introduced: https://github.com/prometheus/node_exporter/issues/1553

In general we should probably get (back) to having two people review PRs. Or did I miss when we changed this? /cc @SuperQ

PR #199, opened 2019-08-03
rebase Initial support for parsing /proc/net/arp #105 from @skottler

TheMeier:
Rebase and slightly refactor abandoned PR https://github.com/prometheus/procfs/pull/105

Signed-off-by: Christoph Maser <christoph.maser+github@gmail.com>

/cc prometheus/node_exporter#540 where this was originally discussed
/cc prometheus/node_exporter#535

### DIFF
--- a/arp.go
+++ b/arp.go
@@ -1,0 +1,84 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


arp.go
```diff
@@ -0,0 +1,84 @@
...
+	lines := strings.Split(string(data), "\n")
+	entries := make([]ARPEntry, 0)
+	var err error
+
+	for _, line := range lines {
+		columns := strings.Fields(line)
+		width := len(columns)
+		expectedDataWidth := 6
+		expectedHeaderWidth := 9
+
```
pgier: I think `expectedDataWidth` and `expectedHeaderWidth` should be moved to a `const` block outside of the for loop.

TheMeier: Right, changed it.

PR #203, opened 2019-08-18
Add support for /proc/net/softnet_stat

kormat:
softnet_stat contains important information for network performance
tuning. See these resources for details:
- https://blog.packagecloud.io/eng/2016/06/22/monitoring-tuning-linux-networking-stack-receiving-data/#monitoring-network-data-processing
- https://access.redhat.com/sites/default/files/attachments/20150325_network_performance_tuning.pdf
- https://github.com/leandromoreira/linux-network-performance-parameters#interrupt-coalescing-soft-irq-and-ingress-qdisc


net_softnet.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2019 Stephen Shirley <kormat@gmail.com>
```
pgier: Should be "The Prometheus Authors"

net_softnet_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019 Stephen Shirley <kormat@gmail.com>
```
pgier: Should be "The Prometheus Authors"

net_softnet.go
kormat: Done.

net_softnet_test.go
kormat: Done.

PR #204, opened 2019-08-30
Add parsing of `Tgid:` from `/proc/<pid>/status`

k1LoW:
Hi @discordianfish @pgier ! Thank you for GREAT library !

I added parsing of `Tgid:` from `/proc/<pid>/status`

This pull request contains 2 fixes.

- Fix fixtures.ttar ( /proc/26231/status `Pid:` should be 26231 )
- Add parsing of `Tgid:` from `/proc/<pid>/status`

## Use case

- I want to detect if thread reader process ( using `p.PID == p.TGID` )

Thank you !


k1LoW:
@pgier Thank you!!!

PR #209, opened 2019-09-03
add parsing of /proc/cpuinfo

pgier:


pgier:
@SuperQ @discordianfish you guys ok with this?

SuperQ:
Yes, this is totally in line with this code.

pgier:
@SuperQ I initially implemented the fields of the `CPUInfo` struct using uints and floats where it seemed appropriate, however, I'm kind of wondering if it would be better to just make all the struct fields strings.  Since most or all of these would likely be used as labels and not as metric values, it seems kind of pointless to convert to a numeric type if we're just going to convert back to a string later anyway.  This would also mean we need less error checking and we can just return whatever value is read even if it's not a valid number.
WDYT?  Should all the struct fields be strings here?

pgier:
Another possibility is to just use a `map[string]string` instead of interpreting the fields at all

SuperQ:
I think keeping things numeric is fine, as this library can be used for other things besides node_exporter labels. :smile:

pgier:
My concern is that we'll run into some case where cpuinfo contains a non-numeric value for one of these fields, and then it becomes a bug in procfs.  If we just return the string, then it's up to the client code to determine how to correctly interpret the data, and they can add some handling for specific weird cases if needed.

SuperQ:
`#proc-life`

pgier:
I'm just going to change the ID fields to strings since I think math operations on those fields normally doesn't make sense, and also change the CPU MHz to a float64 since that should always be numeric hopefully.

PR #210, opened 2019-09-05
add cpu topology information

pgier:
Parse /sys/devices/system/cpu/cpuN/topology and /sys/devices/system/cpu/cpuN/thermal_throttle


pgier:
@discordianfish @SuperQ  PTAL

PR #213, opened 2019-09-05
ignore symlinks in sysfs when iterating directories

pgier:
Prevents error when attempting to read a symlink as a regular file.

Related to https://github.com/prometheus/node_exporter/issues/1465


pgier:
Same problem as #206

PR #215, opened 2019-09-11
additional error handling when reading files in /sys/class

pgier:
Some regular files in /sys/class are not readable on certain
systems, so just ignore them.

Related to #213 which checks that directory entries are regular files and not symlinks.  However, there is the additional problem that some files in /sys/class/ (e.g. `/sys/class/net/enp0s31f6/phys_port_id`) are not readable even though they are regular files.  Before the recent refactoring, these read errors were ignored.  This change makes these errors ignored again.


pgier:
See also https://github.com/prometheus/node_exporter/issues/1465

PR #219, opened 2019-10-02
introduce new function ReadFileNoStat

pgier:
This function should be a slight improvement over
ioutil.ReadFile because it skips the call to os.Stat which commonly
returns the wrong file size for files in /proc and /sys.

Usage of this function also makes the code a bit more concise for
parsing of several of the files in /proc.

### DIFF
--- a/internal/util/sysreadfile.go
+++ b/internal/util/sysreadfile.go
@@ -17,6 +17,7 @@ package util
 
 import (
 	"bytes"
+	"io"
 	"io/ioutil"
 	"os"
 	"syscall"
@@ -51,12 +52,17 @@ func SysReadFile(file string) (string, error) {
 // ReadFileNoStat uses ioutil.ReadAll to read contents of entire file.
 // This is similar to ioutil.ReadFile but without the call to os.Stat, because
 // many files in /proc and /sys report incorrect file sizes (either 0 or 4096).
+// Reads a max file size of 512kB.  For files larger than this, a scanner
+// should be used.
 func ReadFileNoStat(filename string) ([]byte, error) {
+	const maxBufferSize = 1024 * 500
+
 	f, err := os.Open(filename)
 	if err != nil {
 		return nil, err
 	}
 	defer f.Close()
 
-	return ioutil.ReadAll(f)
+	reader := io.LimitReader(f, maxBufferSize)
+	return ioutil.ReadAll(reader)
 }
--- a/internal/util/sysreadfile.go
+++ b/internal/util/sysreadfile.go
@@ -17,12 +17,15 @@ package util
 
 import (
 	"bytes"
+	"io/ioutil"
 	"os"
 	"syscall"
 )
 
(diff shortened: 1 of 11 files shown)


pgier:
To provide a bit more context on this, I started updating procfs functions to follow the new guidelines to read an entire file at once using `ioutil.ReadFile`.  However, I found several places in the root package where we were basically doing the same thing as `ReadFile` but without the call to `os.Stat`.  Since stat does not return valid file sizes for most of the files in /proc and /sys it seemed to make sense to skip that part of `ioutil.ReadFile`.

pgier:
Before the changes we also weren't making calls to stat in these cases, so I was a bit worried that using `ioutil.ReadFile` could introduce an issue if any of the stat calls were to fail.  Separately, I'll make some additional changes to remove usage of `ioutil.ReadFile` where it's currently being used.

internal/util/sysreadfile.go
```diff
@@ -35,11 +38,25 @@ func SysReadFile(file string) (string, error) {
...
+// This is similar to ioutil.ReadFile but without the call to os.Stat, because
+// many files in /proc and /sys report incorrect file sizes (either 0 or 4096).
+func ReadFileNoStat(filename string) ([]byte, error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	return ioutil.ReadAll(f)
```
mdlayher: It would probably be a good idea to put a limit on this of like 32/64/128 MiB or something via io.LimitReader, just as a sanity check.

pgier: Seems like a good idea.  I think we could make it even smaller like 500kB?  Any larger and we should be using a scanner instead of reading the whole thing at once.

internal/util/sysreadfile.go
```diff
@@ -35,11 +39,30 @@ func SysReadFile(file string) (string, error) {
...
+
+	f, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	reader := io.LimitReader(f, maxBufferSize)
+	return ioutil.ReadAll(reader)
+}
```
dmitshur: This .go file has the following build constraint on line 14:

```
// +build linux,!appengine
```

If you're adding a new function here, you should also add it to [internal/util/sysreadfile_compat.go](https://github.com/prometheus/procfs/blob/c0c2a8be4d30a2e2cb95ea371a6f32a506d3e45e/internal/util/sysreadfile_compat.go) file. Otherwise, this function will only be available on non-appengine linux.

dmitshur: Unlike `SysReadFile`, `ReadFileNoStat` doesn't actually use syscalls, so it should probably be added to another .go file without the build constraint. That way you wouldn't have to have 2 identical implementations of it in `sysreadfile.go` and `sysreadfile_compat.go`.

PR #224, opened 2019-10-11
RAPL reader implementation

uniemimu:
This adds support for reading the Running Average Power Limit values from sysfs.

https://www.kernel.org/doc/Documentation/power/powercap/powercap.txt

@discordianfish
@pgier

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -2776,6 +2776,134 @@ SymlinkTo: ../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:00/PNP0C09:00
 Path: fixtures/sys/class/power_supply/BAT0
 SymlinkTo: ../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:00/PNP0C09:00/PNP0C0A:00/power_supply/BAT0
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/powercap
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/powercap/intel-rapl
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/powercap/intel-rapl/enabled
+Lines: 1
+1
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/powercap/intel-rapl/uevent
+Lines: 0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/powercap/intel-rapl:0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/powercap/intel-rapl:0/constraint_0_max_power_uw
+Lines: 1
+95000000
+Mode: 444
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/powercap/intel-rapl:0/constraint_0_name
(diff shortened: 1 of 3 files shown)


sysfs/rapl.go
```diff
@@ -0,0 +1,139 @@
...
+
+// NewRaplValues returns a slice of RaplValues read from the sysfs
+// https://www.kernel.org/doc/Documentation/power/powercap/powercap.txt
+func (reader RaplReader) NewRaplValues() ([]RaplValue, error) {
+	length := len(reader.raplValues)
+	retValues := make([]RaplValue, length, length)
+	var err error
+	for i := 0; i < length; i++ {
+		reader.raplValues[i].Microjoules, err =
+			util.ReadUintFromFile(reader.raplValues[i].sysfsFilename)
```
pgier: This error is not being checked each time through the loop, so if you get an error on first iteration and then a `nil` on the second, the error will be ignored, and only the `nil` will be returned.

sysfs/rapl.go
```diff
@@ -0,0 +1,139 @@
...
+			reader.raplValues = append(reader.raplValues, privRv)
+
+			// Store into map how many times this name has been used. There can
+			// be e.g. multiple "dram" instances without any index postfix. The
+			// count is then used for indexing
+			countNameUsages[modifiedPromName] = index + 1
+		}
+	}
+
+	return &reader, nil
```
pgier: Why do you return a reader here instead of just reading and setting the Microjoules value right away?

sysfs/rapl.go
```diff
@@ -0,0 +1,139 @@
+// Copyright 2019 The Prometheus Authors
```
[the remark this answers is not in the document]
uniemimu: Ok I'll rename

sysfs/rapl.go
```diff
@@ -0,0 +1,139 @@
...
+	// loop through directory files searching for file "name" from subdirs
+	for _, f := range files {
+		nameFile := filepath.Join(raplDir, f.Name(), "/name")
+		nameBytes, err := ioutil.ReadFile(nameFile)
+		if err == nil {
+			// add new rapl value since name file was found
+			name := strings.TrimSpace(string(nameBytes))
+
+			// make RAPL name prometheus-compatible
+			modifiedPromName := strings.Replace(name, "-", "_", -1)
```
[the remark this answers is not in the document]
uniemimu: Sure

sysfs/rapl.go
uniemimu: It is more efficient to resolve the duplicate name issue and possible name postfix parsing just once, instead of every time the values are read. This is what I intended for the API usage:

1) user creates a RaplReader, which at that point goes through the files and figures out which names are duplicates and which need their postfix (_0, _1) removed and delivered as an index instead (that value can be used as a label then).
2) user calls NewRaplValues which only goes through the slice of known values for which all duplicate-issues and indexing has been resolved earlier. The function only reads the values from the known paths, as fast as possible.

Now, the added API-complexity may not be justified by the amount of saved CPU-cycles. If you want, I can combine the functions. Your call.

sysfs/rapl.go
uniemimu: true, I'll fix that

sysfs/rapl.go
pgier: Ok, makes sense, you are avoiding re-reading some of the static information such as the directory structure and max energy since those doesn't normally change between reads.  Maybe some slightly different naming of can make this more clear.  What do you think about instead of having `RaplValue` and `raplValue` structs, there could be a `RaplZone` struct which contains the static info such as the name, max energy, and the path?  You would have a function like `GetRaplZones` which returns a slice of the RaplZones, then you can have a method of RaplZone which reads the current state.

```go
func (z *RaplZone) GetEnergyMicroJoules() (uint64, err) {
    return util.ReadUintFromFile(filepath.Join(z.Path, "energy_uj")
}
```
In this case the caller would be responsible for looping through the zones, which I think would be ok, since they would need to loop through the results in either case. WDYT?

uniemimu: That sounds fine by me. I was attempting to make an effort at hiding any procfs-details from the API user, such as the paths, but having the path in the struct does make it simpler and your proposed API is very clearly named. I'll upload a version with these changes next.

uniemimu: See 
2192ec2

PR #226, opened 2019-10-22
fix: RxPackets is doubled in package procfs

beautytiger:


discordianfish:
Uhh good catch, thanks!

Issue #225, opened 2019-10-17
Parsing hypen in mountinfo incorrect


meriororen:
Hello,

Sometimes other field (like mount points) can have hyphen in them, causing mountinfo parser to put wrong information in wrong field. 

I do think this line
https://github.com/prometheus/procfs/blob/master/mountinfo.go#L93

can be improved. 

In my case, my usb drive mount point is, /media/isa/300A-6F09, this could cause incorrect optional field separation.

SuperQ:
Yup, there are other cases where this is broken. We also want to eliminate the use of buffio scanners, so I think a bit of a re-write is in order. It might be worth making this a regexp parser.

PR #228, opened 2019-10-25
Update mountinfo parsing

SuperQ:
* Use a regexp to parse the mountinfo lines.
* Use util.ReadFileNoStat() to instead of a buffered reader.
* Add a test for hypens in the mountpoint.

Fixes: https://github.com/prometheus/procfs/issues/225

### DIFF
--- a/mountinfo.go
+++ b/mountinfo.go
@@ -15,19 +15,13 @@ package procfs
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
-	"io"
-	"os"
 	"strconv"
 	"strings"
-)
 
-var validOptionalFields = map[string]bool{
-	"shared":         true,
-	"master":         true,
-	"propagate_from": true,
-	"unbindable":     true,
-}
+	"github.com/prometheus/procfs/internal/util"
+)
 
 // A MountInfo is a type that describes the details, options
 // for each mount, parsed from /proc/self/mountinfo.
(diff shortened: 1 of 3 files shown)


discordianfish:
Hrm not sure that using a regex is the best way. How do the usual linux tools parse this? I'd assume there is some regex-free way to do that. If not, this is fine.

SuperQ:
The only code I've found so far was some random Perl code. It splits on spaces, which might be a reasonable way to go. But I thought the regexp would be a little better for validating the contents of the various sections.

discordianfish:
I'd argue we don't want to validate this at all.

See: http://man7.org/linux/man-pages/man5/proc.5.html
I think the right way to parse this is:
- split by space
- store elements 1-6 individually
- start looping over element 7+ until you find a -
- store remaining elements individually

SuperQ:
Ok, I'll re-work this.

SuperQ:
@discordianfish Ok, now it's a bunch of much simpler offsets and loops. Thanks for the suggestion.

SuperQ:
I cleaned up a couple more things, I feel like the optional field parsing is now easier to understand, and less buggy.

mountinfo.go
```diff
@@ -89,57 +75,76 @@ func parseMountInfo(r io.Reader) ([]*MountInfo, error) {
...
+func mountOptionsParseOptionalFields(o []string) (map[string]string, error) {
+	optionalFields := make(map[string]string)
+	for _, field := range o {
+		optionSplit := strings.SplitN(field, ":", 2)
+		value := ""
+		if len(optionSplit) == 2 {
+			value = optionSplit[1]
+		}
+		if !mountOptionsIsValidField(optionSplit[0]) {
+			return optionalFields, fmt.Errorf("invalid optional key: %s", optionSplit[0])
```
pgier: I'm wondering if we should just ignore invalid fields instead of returning an error since the procfs docs say: `Parsers should ignore all unrecognised optional fields.`

discordianfish: @SuperQ Whats the reasoning for validating them at all? I feel like we should just include whatever is in there without any validation.

discordianfish:
Beside that, LGTM

mountinfo.go
SuperQ: Hmm, good point, I will update this part.

SuperQ: @discordianfish I feel like we should follow the procfs docs and ignore them.

SuperQ:
Ok, I returned the optional field check to ignore not fail.

PR #232, opened 2019-11-06
feat: Add support for meminfo

### LINKED ISSUE #230: Add support for parsing meminfo
Would be great to have support for /proc/meminfo. 
I see https://github.com/prometheus/procfs/pull/18 from a few years ago. Is this a good enough base to start with?

bradbeam:
fixes #230 

Signed-off-by: Brad Beam <brad.beam@talos-systems.com>


bradbeam:
I was looking at node exporter and noticed that the values are normalized as [bytes](https://github.com/prometheus/node_exporter/blob/master/collector/meminfo_linux.go#L55-L62), should I do the same here or leave as the default units?

pgier:
Hmm, good question.  Since these fields seem to be hard coded to kB (https://elixir.bootlin.com/linux/latest/source/fs/proc/meminfo.c) except for the HugePage fields, I think it's ok to just leave them as is and allow the client to do the unit conversion as needed.  We should just make it clear in the godoc that the fields are all kB except for HugePages fields which are a count.
@SuperQ wdyt?

SuperQ:
For the library, we tend to keep things in the original form, and leave it up to the user (node_exporter) to do the conversion.

PR #236, opened 2019-11-21
procfs: clean up /proc/crypto parser and tests

mdlayher:
I think this code could use more unit tests, but I've tidied it up a bit to make it easier to follow.

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -14,10 +14,10 @@
 package procfs
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
-	"io/ioutil"
-	"strconv"
+	"io"
 	"strings"
 
 	"github.com/prometheus/procfs/internal/util"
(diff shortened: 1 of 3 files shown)


crypto.go
```diff
@@ -52,80 +53,99 @@ type Crypto struct {
...
+		if len(kv) != 2 {
+			return nil, fmt.Errorf("malformed crypto line: %q", text)
 		}
-		crypto = append(crypto, newCryptoElem)
+
+		k := strings.TrimSpace(kv[0])
+		v := strings.TrimSpace(kv[1])
+
+		// Parse the key/value pair into the currently focused element.
+		parseCryptoKV(k, v, &out[len(out)-1])
```
pgier: Should parseCryptoKV maybe be a method of Crypto?

mdlayher: Yeah it certainly could be.

mdlayher:
@pgier updated and fixed a bug. Mind doing another review?

pgier:
Looks good, thanks!

PR #237, opened 2019-11-23
procfs: implement parsing of NetSockstat{,6}

### LINKED ISSUE #229: Add parsing of sockstat
We currently parse `/proc/net/sockstat` in the node_exporter, it would be good to extract this functionality.

* [ ] Parse `/proc/net/sockstat`
* [ ] Parse `/proc/net/sockstat6`

Sadly, it looks like these files are not documented well in the kernel docs.

mdlayher:

Fixes #229.

### DIFF
--- a/net_sockstat.go
+++ b/net_sockstat.go
@@ -1,0 +1,158 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procfs
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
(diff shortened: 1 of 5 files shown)


net_sockstat.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// A NetSockstat contains the output of /proc/net/sockstat{,6} for IPv4 or IPv6,
+// respectively.
+type NetSockstat struct {
+	// Used is non-nil for IPv4 sockstat results, but nil for IPv6.
+	Used      *int
+	Protocols []NetSockstatProtocol
```
mdlayher: The existing sockstat code in node_exporter uses a map for this, but I honestly really dislike maps in APIs. Slices and structures are almost certainly easier to work with than a map of strings.

SuperQ:
Hrmm, there's no update to the end to end test.

mdlayher:
We don't have those in procfs right?

mdlayher:
Thoughts on the implementation here?

SuperQ:
Sorry, I was confused, thought I was looking at the node_exporter repo. :facepalm:

net_sockstat.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// NetSockstat6 retrieves IPv6 socket statistics.
+func (fs FS) NetSockstat6() (*NetSockstat, error) {
+	return readSockstat(fs.proc.Path("net", "sockstat6"))
+}
+
+// readSockstat opens and parses a NetSockstat from the input file.
+func readSockstat(name string) (*NetSockstat, error) {
+	f, err := os.Open(name)
```
SuperQ: We should comment this to document that this shouldn't be converted to `util.ReadFileNoStat()` due to the possible large file size.

mdlayher: @SuperQ actually this file in particular is quite small, so I should be using the util function. Updated now.

PR #240, opened 2019-11-23
procfs: simplify meminfo parser

mdlayher:

This function should probably return `*Meminfo` per the usual Go conventions. Are we okay to make that change too?

### DIFF
--- a/meminfo.go
+++ b/meminfo.go
@@ -16,6 +16,8 @@ package procfs
 import (
 	"bufio"
 	"bytes"
+	"fmt"
+	"io"
 	"strconv"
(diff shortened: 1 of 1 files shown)


SuperQ:
Good question, we need to standardize a bit on how we return stuff from procfs. I'd like to be able to tell the difference between "no data" and "error" for https://github.com/prometheus/node_exporter/issues/1323.

mdlayher:
Indeed. But any thoughts on the contents of this PR as is?

SuperQ:
Yea, fine for now.

PR #242, opened 2019-11-25
procfs: tidy up NetUnix APIs and internals

mdlayher:

- remove unnecessary exported APIs; `FS.NetUnix` is sufficient and anything else is added API surface for little gain
- remove unecessary constants which are only used once (twice for error messages)
- remove duplication in tests and switch to go-cmp
- more idiomatic error messages

### DIFF
--- a/net_unix.go
+++ b/net_unix.go
@@ -15,7 +15,6 @@ package procfs
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


net_unix.go
```diff
@@ -81,101 +68,103 @@ type NetUnix struct {
...
-// NewNetUnix returns data read from /proc/net/unix.
-func NewNetUnix() (*NetUnix, error) {
-	fs, err := NewFS(DefaultMountPoint)
-	if err != nil {
-		return nil, err
-	}
-
-	return fs.NewNetUnix()
+// NetUnix returns data read from /proc/net/unix.
+func (fs FS) NetUnix() (*NetUnix, error) {
```
mdlayher: Since I'd like to drop the "New" prefix since methods are not constructors, is this also a suitable time to rename this function and types to `NetUNIX`?

pgier: Yes, I think that's fine since the names are changing anyway.

discordianfish:
@mdlayher: So many nice contributions! 🔥

PR #243, opened 2019-12-02
Add UIDs to ProcStatus struct

keitwb:
These are left as a string since the built-in function os.LookupId
accepts UID as a string.


keitwb:
@mdlayher Does this seem alright to you?

PR #245, opened 2019-12-11
less restrictive checking for xfs quota stats

pgier:
It seems that the xfs quota stats provided in /proc/fs/xfs/stats can
provide more than 8 values, so instead of erroring, just ignore the
extra fields.


mdlayher:
Test coverage? Otherwise there's no guarantee this won't break.

Issue #244, opened 2019-12-11
Add parsing of /proc/swaps


stevenjm:
I currently have a WIP implementation at https://github.com/usabilla/prometheus-procfs/tree/swap (with a corresponding WIP change to the node exporter at https://github.com/usabilla/prometheus-node_exporter/tree/swap), but both of these lack tests. Creating an issue to claim this task; I should be ready to raise a PR soon.

discordianfish:
Great, thanks!

pgier:
Fixed by #246

PR #246, opened 2019-12-12
Add support for /proc/swaps

stevenjm:
This code (both implementation and tests) is largely based on the existing mountinfo implementation.

Implementation of #244.


stevenjm:
@mdlayher Does this look reasonable to you?

PR #249, opened 2019-12-14
Expose xs_qm_dquot_unused XFS counter

skreuzer:
Add support for exposing the xs_qm_dquot_unused XFS statistic which appears in
Linux 4.20 and later.

Signed-off-by: Steven Kreuzer <skreuzer@FreeBSD.org>

### DIFF
--- a/xfs/parse.go
+++ b/xfs/parse.go
@@ -381,11 +381,15 @@ func extendedPrecisionStats(us []uint64) (ExtendedPrecisionStats, error) {
 }
 
 func quotaManagerStats(us []uint32) (QuotaManagerStats, error) {
-	if l := len(us); l < 8 {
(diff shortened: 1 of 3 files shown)


xfs/parse.go
```diff
@@ -394,7 +398,14 @@ func quotaManagerStats(us []uint32) (QuotaManagerStats, error) {
 		Wants:         us[5],
 		ShakeReclaims: us[6],
 		InactReclaims: us[7],
-	}, nil
+	}
+
+	if l == 8 {
+		return s, nil
+	}
```
pgier: I think here you could make this slightly simpler with this:
```golang
if l > 8 {
    s.Unused = us[8]
}
```
And then just `return s, nil`.  Saves one line of code anyway.

skreuzer: Yeah, good point. I just updated the pull request with that style change

pgier:
@SuperQ @discordianfish Ok to merge this?

PR #250, opened 2019-12-19
Fix the overflow while parsing ipv6 addr inside net_udp.go

peterbueschel:
Hey @pgier, Hey @discordianfish, Hey @SuperQ 

Currently inside the net_udp.go while parsing the local_addr and remote_addr fields based on ipv6, uint64 is not enough. So, I decided to use []uint8 instead.

Example output running the node_exporter using this lib:

```
ERROR: udp_queues collector failed after 0.001350s: couldn't get upd6 queued bytes:
cannot parse local_address value in udp socket line: strconv.ParseUint: parsing "000080FE00000000FFADE15609667CFE": value out of range  source="collector.go:132"
```

Signed-off-by: Peter Bueschel <peter.bueschel@logmein.com>


discordianfish:
Oh good catch!

peterbueschel:
Hi @discordianfish, @pgier, @SuperQ can someone please hit the merge button?

net_udp.go
```diff
@@ -54,9 +55,9 @@ type (
 	// For the proc file format details, see https://linux.die.net/man/5/proc.
 	netUDPLine struct {
 		Sl        uint64
-		LocalAddr uint64
+		LocalAddr []uint8
```
SuperQ: Maybe we should use Go [net.IP](https://golang.org/pkg/net/#IP) here?

peterbueschel: @SuperQ Done.

net_udp.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
```
SuperQ: @pgier We don't update copyright dates in Prometheus repos. Only the initial date is required. We don't need to revert this, but we should avoid updating these in the future.

pgier: Ah sorry, didn't notice that change when I reviewed this.  I agree, make sense not to change these in the future.

PR #251, opened 2019-12-20
Add support for loadavg

### LINKED ISSUE #239: Implement parsing of /proc/loadavg
Looks like "/proc/loadavg" file isn't parsed at now. 
It's useful from this to get current system load.

tukeJonny:
Signed-off-by: tukeJonny <ne250143@yahoo.co.jp>.

This is my first contribution to this project.
This PR adds support for parsing /proc/loadavg.

fixes #239

### DIFF
--- a/loadavg.go
+++ b/loadavg.go
@@ -1,0 +1,62 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procfs
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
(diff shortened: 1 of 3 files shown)


tukeJonny:
@mdlayher 
Could you review this PR when you're free? 🙇

loadavg.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return &LoadAvg{
+		Load1:  loads[0],
+		Load5:  loads[1],
+		Load15: loads[2],
+	}, nil
+}
+
+// Parse /proc loadavg and return 1m, 5m and 15m.
+func parseLoad(loadavgString string) (loads []float64, err error) {
```
pgier: The pattern that we've been using for other functions is that this should take a `[]byte` and return the struct.  That way the logic is all in one place for determining which parsed fields go into each struct field.

tukeJonny:
@pgier 
Thank you for your review!
I fix up parseLoad, And squash commits (Exactly, there are extra commits... Thanks for your advice!).

Please re-review this PR 🙇

loadavg.go
```diff
@@ -0,0 +1,66 @@
...
+
+// LoadAvg returns loadavg from /proc.
+func (fs FS) LoadAvg() (*LoadAvg, error) {
+	path := fs.proc.Path("loadavg")
+
+	data, err := util.ReadFileNoStat(path)
+	if err != nil {
+		return nil, err
+	}
+	loads, err := parseLoad(data)
```
pgier: I think here you can just `return parseLoad(data)` and skip the error check since it the result should be the same either way.

tukeJonny: @pgier 
Thanks for the feedback!
I had to check carefullly.

I fix it.

PR #252, opened 2020-01-07
Added the ability to parse /proc/[pid]/maps

timothy-boners:
Hello Prometheus Team,

My name is Timothy and I am new at the Git hub so please bare with me as to also english is not the first of languages of which I have learned. I am a very big fan of your time series! It is probably my favorite of all the series, so I thank you for that project first.

We are a small but simple team of developers here, we have mostly all quit smoking, but all writing tools in which debug memories in the linux system. Unfortunately this is in Golang and we are using procfs, maybe you would like support for the things in which we have added? 

Thank you so much for your continued support of the time series! Please call or tiktok me if you need any updates. I am sorry for any problems! I made sure all additions met requirements. Thank you again!

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -1,4 +1,4 @@
-# Archive created by ttar -c -f fixtures.ttar fixtures/
+# Archive created by ttar -cvf fixtures.ttar fixtures
 Directory: fixtures
 Mode: 775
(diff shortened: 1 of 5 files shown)


timothy-boners:
I fix test. Sorry.

PR #253, opened 2020-01-07
/proc/[pid]/maps parser

timothy-boners:
Hello Prometheus Team,

My name is Timothy and I am new at the Git hub so please bare with me as to also english is not the first of languages of which I have learned. I am a very big fan of your time series! It is probably my favorite of all the series, so I thank you for that project first.

We are a small but simple team of developers here, we have mostly all quit smoking, but all writing tools in which debug memories in the linux system. Unfortunately this is in Golang and we are using procfs, maybe you would like support for the things in which we have added?

Thank you so much for your continued support of the time series! Please call or tiktok me if you need any updates. I am sorry for any problems! I made sure all additions met requirements. Thank you again!

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -1,4 +1,4 @@
-# Archive created by ttar -c -f fixtures.ttar fixtures/
+# Archive created by ttar -cvf fixtures.ttar fixtures
 Directory: fixtures
 Mode: 775
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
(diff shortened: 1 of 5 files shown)


fixtures.ttar
```diff
@@ -1,4 +1,4 @@
-# Archive created by ttar -c -f fixtures.ttar fixtures/
+# Archive created by ttar -cvf fixtures.ttar fixtures
```
pgier: If you use `make update_fixtures` to generate this, then you should get the original line.  I just don't want this one to change if it's going to change back on the next PR.

proc_maps.go
```diff
@@ -0,0 +1,209 @@
...
+	pmap.Inode = inode
+	pmap.Pathname = fname
+
+	return &pmap, nil
+}
+
+// ProcMaps reads from /proc/[pid]/maps to get the memory-mappings of the
+// process.
+func (p Proc) ProcMaps() ([]*ProcMap, error) {
+	file, err := os.Open(p.path("maps"))
```
pgier: The pattern we've been using for reading most `/proc` files is to use the utility function `util.ReadFileNoStat` which returns a byte slice which can then be used in the bufio scanner.  This way the file is not kept open during parsing.

proc_maps.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var fname string
+	if len(fields) >= 5 {
+		fname = strings.Join(fields[5:], " ")
+	}
+
+	pmap.StartAddr = saddr
```
pgier: I think here you can either allocate pmap immediately before this line, or just assign the return values directly to the fields of pmap and then you don't need these assignments.

proc_maps.go
```diff
@@ -0,0 +1,209 @@
...
+	Private bool
+}
+
+// ProcMap contains the process memory-mappings of the process,
+// read from /proc/[pid]/maps
+type ProcMap struct {
+	// The start address of current mapping.
+	StartAddr uintptr
+	// The end address of the current mapping
+	EndAddr uintptr
```
pgier: Do you have a use case for actually converting these into `uintptr`?  I'm kind of wondering if they should just be strings simplicity and the user can convert them as necessary.

timothy-boners:
Thank you very much for the comments! I will update the next morning!

proc_maps.go
timothy-boners: the idea was to keep these as close to the "native" types as possible. I expect `uintptr` to always be of the right type and size for the data contained within this file. Good for address calculation and fighting bear!

fixtures.ttar
timothy-boners: yes, you make point. I will regenterate for next push.

proc_maps.go
SuperQ: This might not be a good idea for maps, some of them could be quite large. On my laptop I found one that's 616KiB, which is larger than `util.ReadFileNoStat` supports.

proc_maps.go
SuperQ: If the kernel produces `uintptr`, then this is fine to use.

timothy-boners: I decided to leave as it stands, but also added a unit test for 32b compatibility checking.

proc_maps.go
timothy-boners: file can be big yes, but also very easy to parse, I worry not of file being open during parse, there is nothing wrong with this, no race condition, this is an adhoc seek hook in kernel.

Issue #217, opened 2019-09-17
Add parsing of net/stat/nf_conntrack


pgier:
See https://github.com/prometheus/node_exporter/pull/1155 and https://github.com/prometheus/node_exporter/issues/1233

kozl:
Have been done in https://github.com/prometheus/procfs/pull/254

PR #254, opened 2020-01-07
Added parsing netfilter conntrack statistics from net/stat/nf_conntrack

kozl:
Added parsing of conntrack statistics from procfs for this PR in node_exporter project: https://github.com/prometheus/node_exporter/pull/1155

@discordianfish please review

### DIFF
--- a/net_conntrackstat.go
+++ b/net_conntrackstat.go
@@ -1,0 +1,153 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procfs
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// A ConntrackStatEntry represents one line from net/stat/nf_conntrack
+// and contains netfilter conntrack statistics at one CPU core
+type ConntrackStatEntry struct {
+	Entries       uint64
+	Found         uint64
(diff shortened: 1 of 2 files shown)


net_conntrackstat.go
```diff
@@ -0,0 +1,154 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// A ConntrackStatEntry represents one line from net/stat/nf_conntrack
+// and contains netfilter conntrack statistics at one CPU core
+type ConntrackStatEntry struct {
+	CPU           int
```
pgier: Is CPU necessary here?  Seems like this can be determined by the location in the slice.

net_conntrackstat.go
```diff
@@ -0,0 +1,154 @@
+// Copyright 2019 The Prometheus Authors
```
pgier: Should be 2020

net_conntrackstat_test.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2018 The Prometheus Authors
```
pgier: Should be 2020

pgier:
Fixes #217

net_conntrackstat.go
kozl: Fixed

net_conntrackstat_test.go
kozl: Fixed

net_conntrackstat.go
kozl: Yeah, probably this is excessive, removed it.

net_conntrackstat.go
```diff
@@ -0,0 +1,150 @@
...
+			return nil, err
+		}
+		entries = append(entries, *conntrackEntry)
+	}
+
+	return entries, nil
+}
+
+// Parses a ConntrackStatEntry from given array of fields
+func parseConntrackStatEntry(fields []string) (*ConntrackStatEntry, error) {
```
discordianfish: Check for length of fields to avoid a runtime panic.

pgier:
Thanks for the update, looks like the new commit doesn't have a signoff, you can either squash it into the first commit, or do a `git commit --amend --signoff`

kozl:
> Thanks for the update, looks like the new commit doesn't have a signoff, you can either squash it into the first commit, or do a `git commit --amend --signoff`

:ok_hand:

net_conntrackstat.go
kozl: Added

PR #257, opened 2020-01-20
multiplatform support for cpuinfo

### LINKED ISSUE #256: /proc/cpuinfo parsing crashes on s390x
The current logic for parsing /proc/cpuinfo is written for x86 platforms, but doesn't really handle other formats currently.  The worst case is probably s390x which causes an array index out of bounds panic.

```
goroutine 39 [running]:
github.com/prometheus/procfs.parseCPUInfo(0xc000203000, 0x4c3, 0xe00, 0x4c3, 0xe00, 0x0, 0x0, 0x1)
	/go/src/github.com/prometheus/node_exporter/vendor/github.com/prometheus/procfs/cpuinfo.go:85 +0x1d32
github.com/prometheus/procfs.FS.CPUInfo(0x3ffcbd7f550, 0xa, 0x24000000000f4684, 0x1d394, 0xc0001eb6c0, 0x70, 0x70)
	/go/src/github.com/prometheus/node_exporter/vendor/github.com/prometheus/procfs/cpuinfo.go:61 +0x180
github.com/prometheus/node_exporter/collector.(*cpuCollector).updateInfo(0xc00005d180, 0xc0000de180, 0x32ea661f00000000, 0x325e45f98388)
	/go/src/github.com/prometheus/node_exporter/collector/cpu_linux.go:96 +0x3c
github.com/prometheus/node_exporter/collector.(*cpuCollector).Update(0xc00005d180, 0xc0000de180, 0xc311a0, 0x0)
	/go/src/github.com/prometheus/node_exporter/collector/cpu_linux.go:81 +0xdc
github.com/prometheus/node_exporter/collector.execute(0x6e60b2, 0x3, 0x7cb140, 0xc00005d180, 0xc0000de180)
	/go/src/github.com/prometheus/node_exporter/collector/collector.go:127 +0x68
github.com/prometheus/node_exporter/collector.NodeCollector.Collect.func1(0xc0000de180, 0xc0000d82f0, 0x6e60b2, 0x3, 0x7cb140, 0xc00005d180)
	/go/src/github.com/prometheus/node_exporter/collector/collector.go:118 +0x48
created by github.com/prometheus/node_exporter/collector.NodeCollector.Collect
	/go/src/github.com/prometheus/node_exporter/collector/collector.go:117 +0xe2
```

This is due to format of `/proc/cpuinfo` being significantly different from the x86 version.  Here's an example which causes the panic:
```
# cat /proc/cpuinfo
vendor_id       : IBM/S390
# processors    : 2
bogomips per cpu: 11061.00
features	: esan3 zarch stfle msa ldisp eimm dfp etf3eh highgprs 
processor 0: version = FF,  identification = 32C5C2,  machine = 2097
processor 1: version = FF,  identification = 32C5C2,  machine = 2097
```

pgier:
This adds basic handling of reading /proc/cpuinfo for ARM, PPC, and
s390x.
Fixes #256

### DIFF
--- a/cpuinfo.go
+++ b/cpuinfo.go
@@ -19,7 +19,6 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
-	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
@@ -57,40 +56,10 @@ type CPUInfo struct {
 	PowerManagement string
 }
 
-const (
-	platformX86     = "x86"
-	platformARM     = "arm"
-	platformS390X   = "s390x"
-	platformPPC     = "ppc"
-	platformUnknown = "unknown"
-)
-
 var (
-	cpuinfoX86Regexp   = regexp.MustCompile(`(?m)^\s*processor\s*:\s*\d+\s*vendor`)
-	cpuinfoARMRegexp   = regexp.MustCompile(`^\s*Processor\s*:\s*ARM`)
-	cpuinfoS390XRegexp = regexp.MustCompile(`^\s*vendor_id\s*:\s*IBM/S390`)
-	cpuinfoPPCRegexp   = regexp.MustCompile(`(?m)^\s*processor\s*:\s*\d+\s+cpu\s+:\s+POWER`)
-
 	cpuinfoClockRegexp          = regexp.MustCompile(`([\d.]+)`)
 	cpuinfoS390XProcessorRegexp = regexp.MustCompile(`^processor\s+(\d+):.*`)
 )
-
-// cpuinfoDetectFormat attempts to determine the format used by the cpuinfo.
-// This format corresponds to the platform generating the /proc/cpuinfo file.
-// Returns "unknown"
-func cpuinfoDetectFormat(info []byte) string {
-	switch {
-	case cpuinfoX86Regexp.Match(info):
-		return platformX86
-	case cpuinfoARMRegexp.Match(info):
-		return platformARM
-	case cpuinfoPPCRegexp.Match(info):
-		return platformPPC
-	case cpuinfoS390XRegexp.Match(info):
-		return platformS390X
-	}
-	return platformUnknown
-}
 
 // CPUInfo returns information about current system CPUs.
 // See https://www.kernel.org/doc/Documentation/filesystems/proc.txt
@@ -98,10 +67,6 @@ func (fs FS) CPUInfo() ([]CPUInfo, error) {
 	data, err := util.ReadFileNoStat(fs.proc.Path("cpuinfo"))
 	if err != nil {
 		return nil, err
-	}
-	if format := cpuinfoDetectFormat(data); format != expectedPlatform {
-		return nil, fmt.Errorf("unable to parse CPU info, expected format %v, found: %v",
-			expectedPlatform, format)
 	}
 	return parseCPUInfo(data)
 }
--- a/cpuinfo.go
+++ b/cpuinfo.go
@@ -11,12 +11,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build linux
+
 package procfs
 
 import (
 	"bufio"
 	"bytes"
 	"errors"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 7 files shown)


cpuinfo.go
```diff
@@ -163,5 +225,148 @@ func parseCPUInfo(info []byte) ([]CPUInfo, error) {
...
+	i := 0
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
+		field := strings.SplitN(line, ": ", 2)
+		switch strings.TrimSpace(field[0]) {
+		case "processor":
```
Prashanth684: don't we want to parse the "clock" field in the cpu info for cpu MHz like for x86 ?

cpuinfo.go
```diff
@@ -163,5 +225,148 @@ func parseCPUInfo(info []byte) ([]CPUInfo, error) {
...
+	commonCPUInfo := CPUInfo{VendorID: field[1]}
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
+		field := strings.SplitN(line, ": ", 2)
+		switch strings.TrimSpace(field[0]) {
+		case "bogomips per cpu":
```
Prashanth684: don't we want to parse the "cpu MHz static" for the cpu MHz info ?

pgier: Yeah, these were just due to me trying to get the minimal stuff working at first, and it seems that the 'cpu MHz' fields might not always be available, based on an old BZ (https://bugzilla.redhat.com/show_bug.cgi?id=746288).  I'm thinking we want to use "cpu MHz dynamic" instead of static since dynamic appears to be the frequency at which the cpu is currently running vs. static is frequency at which it's supposed to run.
https://www.spinics.net/lists/util-linux-ng/msg12976.html

pgier:
@Prashanth684 thanks for the feedback!  I've added support for CPUMHz on both PPC and s390x.

cpuinfo.go
Prashanth684: ok..sounds good.. yes we can use the dynamic cpu ..

Prashanth684:
> @Prashanth684 thanks for the feedback! I've added support for CPUMHz on both PPC and s390x.

no problem. thanks for this!

Prashanth684:
lgtm

discordianfish:
Looks good to me in general but I'm wondering if we should rather use build flags to determine the expected format. Or is there a use case where a x64 system would read a arm cpuinfo?

pgier:
@discordianfish I did it this way originally because I was thinking that might be able to handle platforms that weren't accounted for, and it seemed easier to test this way.  But I think you are right that build flags would be better.

discordianfish:
@pgier kk, awaiting your changes :)

pgier:
@discordianfish I experimented a bit with separate files and build flags for the different arches, but I don't have an easy way to test the various platforms.  I'm wondering if we should just leave this as is since none of the code is really platform dependent, it only depends on the different text file formats.  WDYT?

discordianfish:
@pgier Hrm I don't like this implicit runtime detection. I'd say we should either use build flags or add functions for each platform and leave it to the caller to invoke the right one for the platform.

But what was the problem with separate files and build flags? Just the tests? Maybe we could keep the functions but use platform specific files with build flags that implement the correct function for the platform, e.g:

**cpuinfo_amd64.go:**
```
var parseCPUInfo = parseCPUInfoX86
```

pgier:
@discordianfish Thanks for the suggestions, sorry it took so long to get this updated!  Please take another look.

discordianfish:
Not sure we need and want to maintain the detection code. All this would work without it, right? We just need to run parseCPUInfo which is set in the platform specific files.
I'm not strictly against it, but I don't really see where this would be useful. I can't think of a case where I would like to parse a foreign cpuinfo.

pgier:
@discordianfish ok, I have removed the format detection code.

mfraioli:
Hey guys, I'm trying to build some code which apparently pulls in this procfs project as a dependency, and it started failing yesterday on ppc and s390 with this error:

```
+ go get github.com/uber/jaeger-lib/metrics/prometheus
golib/src/github.com/prometheus/procfs/cpuinfo.go:71:9: undefined: parseCPUInfo
```
The actual `go build` command fails in the same way.

I assume it's related to this change.  I see that parseCPUInfo is a function variable that is set in a platform-specific file to point to the platform-specific function, but I"m not clear on how the appropriate file is supposed to be included. Somehow it needs to know to use `cpuinfo_ppc64.go` on ppc, etc. How is this meant to be specified? Thanks!

mfraioli:
On 390 there is a slightly different error:

```
+ go get github.com/prometheus/client_golang/prometheus/promhttp
golib/src/github.com/prometheus/procfs/cpuinfo.go:71:2: not enough arguments to return
golib/src/github.com/prometheus/procfs/cpuinfo_s390x.go:19:20: undefined: parseCPUInfoS390x
```

This seems to be a simple typo in `cpuinfo_s390x.go`. It specifies the function name as `parseCPUInfoS390x` but it should be `parseCPUInfoS390X` (upper case X at the end).

For the ppc problem, the issue may be with the build constraints in `cpuinfo_ppc64.go`:

```
// +build linux
// +build ppc64,ppc64le
```

The Go doc on build constraints says:

> A build constraint is evaluated as the OR of space-separated options. Each option evaluates as the AND of its comma-separated terms.

So that means this will only be included if both ppc64 AND ppc64le are set, but I believe only one or the other will ever be true.

mfraioli:
With
```
// +build ppc64,ppc64le
```
corrected to:
```
// +build ppc64 ppc64le
```
it still doesn't work, because `go build` filters by the filename as well. Since I am using ppc64le, the file `cpuinfo_ppc64.go` gets left out. If I rename it to `cpuinfo_ppc64le.go` then it works. So probably it needs to be two separate files, one for each platform.

pgier:
@mfraioli thanks so much for reporting this!  I created #288 which hopefully fixes the issues.  Can you give that one a quick look?

PR #258, opened 2020-01-29
An inode entry does not always contain a mask part

lespea:
Additionally, there was no check to make sure the regex even matched so this was panicing at runtime

### DIFF
--- a/proc_fdinfo.go
+++ b/proc_fdinfo.go
@@ -16,6 +16,7 @@ package procfs
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"regexp"
 
 	"github.com/prometheus/procfs/internal/util"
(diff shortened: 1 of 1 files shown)


proc_fdinfo.go
```diff
@@ -96,15 +98,22 @@ type InotifyInfo struct {
...
+		}
+		i := &InotifyInfo{
+			WD:   m[1],
+			Ino:  m[2],
+			Sdev: m[3],
+			Mask: mask,
+		}
+		return i, nil
+	} else {
+		return nil, errors.New("invalid inode entry: " + line)
```
pgier: I think you can remove the `else` block around this return.

lespea: Okay I can remove it.  Do you need me to do a squash/force push or are you okay with another commit?

pgier: Either way is fine, but you'll have to add the signoff to each commit.

lespea:
Sorry I took so long to get this signed...

PR #260, opened 2020-02-06
Parse /proc/<pid>/cgroup file

### LINKED ISSUE #201: Add support for parsing cgroup
We've encountered a case where we needed to know wether a particular process was running inside a container or not. One way to achieve this is to check for cgroups inheritance from the container runtime via `/proc/<pid>/cgroup`.

It would be great to have visibility on the hierarchical dependencies per cgroup resource controller.

hamiltont:
This is a quick patch to support parsing the per-process cgroup file. Should work OK on both cgroups-v1 and cgroups-v2. 

There is one TODO in the code I could use some feedback on :-) 

I've intentionally not gone too crazy in adding this since it's my first contrib. Future work would be to locate the full path to the cgroup folder per each `Cgroup` struct (this could be accomplished by reading the mount point for that controller hierarchy and then prefixing this to the parsed Path), and also reading the actual subsystem-specific files. I do have some code to handle these later issues but I wanted to get one small change through review before biting off more 

fixes #201

### DIFF
--- a/proc_cgroup.go
+++ b/proc_cgroup.go
@@ -1,4 +1,4 @@
-// Copyright 2019 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/proc_cgroup.go
+++ b/proc_cgroup.go
@@ -70,7 +70,7 @@ func parseCgroupString(cgroupStr string) (*Cgroup, error) {
 
 // parseCgroups reads each line of the /proc/[pid]/cgroup file
 func parseCgroups(data []byte) ([]*Cgroup, error) {
-	cgroups := []*Cgroup{}
+	var cgroups []*Cgroup
 	scanner := bufio.NewScanner(bytes.NewReader(data))
 	for scanner.Scan() {
 		mountString := scanner.Text()
@@ -85,9 +85,11 @@ func parseCgroups(data []byte) ([]*Cgroup, error) {
 	return cgroups, err
 }
 
-// GetCgroups returns a Cgroup struct for all process control hierarchies running on this system
-func GetCgroups(pid int) ([]*Cgroup, error) {
-	data, err := util.ReadFileNoStat(fmt.Sprintf("/proc/%d/cgroup", pid))
+// Cgroups reads from /proc/<pid>/cgroups and returns a []*Cgroup struct locating this PID in each process
+// control hierarchy running on this system. On every system (v1 and v2), all hierarchies contain all processes,
+// so the len of the returned struct is equal to the number of active hierarchies on this system
+func (p Proc) Cgroups() ([]*Cgroup, error) {
+	data, err := util.ReadFileNoStat(fmt.Sprintf("/proc/%d/cgroup", p.PID))
 	if err != nil {
 		return nil, err
 	}
--- a/proc_cgroup.go
+++ b/proc_cgroup.go
@@ -26,21 +26,25 @@ import (
 // specific control hierarchy. The kernel has two cgroup APIs, v1 and v2. v1 has one hierarchy per available resource
 // controller, while v2 has one unified hierarchy shared by all controllers. Regardless of v1 or v2, all hierarchies
 // contain all running processes, so the question answerable with a Cgroup struct is 'where is this process in
-// this hierarchy' (where==what path). By prefixing this path with the mount point of this hierarchy, you can locate
-// the relevant pseudo-files needed to read/set the data for this PID in this hierarchy
+// this hierarchy' (where==what path on the specific cgroupfs). By prefixing this path with the mount point of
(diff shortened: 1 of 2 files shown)


hamiltont:
Update: forgot to tag maintainers 

/cc @discordianfish @pgier

hamiltont:
Thanks @discordianfish

proc_cgroup.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	cgroup := &Cgroup{
+		Path: fields[2],
+	}
+	cgroup.HierarchyId, err = strconv.Atoi(fields[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse hierarchy ID")
+	}
+	// fields[1] may be ""
```
pgier: I'm thinking it might be better to initialize cgroup.Controllers to an empty slice, and then have a check here for empty string.  Then do the split and append the ssNames only if `fields[1]` is not empty.  And then you can remove the `TODO` in the comments above.

hamiltont:
@pgier updated, thanks. One question - should this not return a pointer, to stay consistent with other APIs? I see `Limits()` and `IO()` and `Namespaces()` return base structs

pgier:
@hamiltont For `parseCgroupString` I think it's fine to return a pointer since it's not exported, and I don't think there is a reason to return by copy.  For `Cgroups()` it might be better to return a slice of structs instead of a slice of pointers to structs, since when you return a slice you are only copying the slice metadata anyway and not copying the entire set of data.  And with the non-pointer type the caller doesn't have to worry about checking for nils.

proc_cgroup.go
```diff
@@ -0,0 +1,97 @@
...
+	Path string
+}
+
+// parseCgroupString parses each line of the /proc/[pid]/cgroup file
+// Line format is hierarchyID:[controller1,controller2]:path
+func parseCgroupString(cgroupStr string) (*Cgroup, error) {
+	var err error
+
+	fields := strings.Split(cgroupStr, ":")
+	if len(fields) != 3 {
```
discordianfish: Sometimes we check for the exact number of fields, sometime that there are at least x. We should probably write down what we want to do in the docs somewhere. I'd say in these cases we should always check for 'at least x fields' since it seems common for linux kernel changes to append fields. While unlikely here, I'd still change it to `>= 3`.

discordianfish: @pgier @SuperQ wdyt?

discordianfish:
Beside that looks good, great actually (yay comments!) :)

proc_cgroup.go
pgier: Yeah, I agree it's usually better to check for a minimum number of fields instead of an exact match to handle future cases where some fields may be added.

proc_cgroup.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if fields[1] != "" {
+		ssNames := strings.Split(fields[1], ",")
+		cgroup.Controllers = append(cgroup.Controllers, ssNames...)
+	}
+	return cgroup, nil
+}
+
+// parseCgroups reads each line of the /proc/[pid]/cgroup file
+func parseCgroups(data []byte) ([]*Cgroup, error) {
```
pgier: Can you change this from `[]*Cgroup` to `[]Cgroup`?  I don't think there is a need for pointers here.

proc_cgroup.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	err := scanner.Err()
+	return cgroups, err
+}
+
+// Cgroups reads from /proc/<pid>/cgroups and returns a []*Cgroup struct locating this PID in each process
+// control hierarchy running on this system. On every system (v1 and v2), all hierarchies contain all processes,
+// so the len of the returned struct is equal to the number of active hierarchies on this system
+func (p Proc) Cgroups() ([]*Cgroup, error) {
```
pgier: Change `[]*Cgroup` to `[]Cgroup`.

proc_cgroup.go
pgier: Change this from ` != 3` to ` >= 3`.

hamiltont:
> @hamiltont Would you mind making a couple more minor changes? And then I'll merge this.

Yup. Had to reformat my main computer, but will get back to this shortly :-)

hamiltont:
@pgier done (ps - does github notify you when I click re-request review? e.g. is me commenting actually pinging folks twice? I find the review UI a bit confusing...)

proc_cgroup.go
SuperQ: Yes, the common kernel pattern is to add new fields, and optionally deprecate old ones. The kernel "Don't break userspace" philosophy.

pgier:
@hamiltont Looks like the latest commit is missing a signoff.  Would you mind also squashing to a single commit?

hamiltont:
> @hamiltont Looks like the latest commit is missing a signoff. Would you mind also squashing to a single commit?

Fixed :-)

pgier:
Merged, thanks!

PR #264, opened 2020-02-14
Unwrap infiband error

SuperQ:
Don't wrap the error returned by the infiniband parser so that external
systems can parse the error with `os.IsNotExist(err)`.


SuperQ:
This is needed to help fix error detecting in the node_exporter.

PR #265, opened 2020-02-20
bcache: add priorityStats flag

AlexZzz:
bcache: add priorityStats flag

This commit adds priorityStats flag,
which is used when we need priority stats info.

The reason for this flag to be false is that
collecting of priority stats might be expensive
as described here:
https://github.com/prometheus/node_exporter/issues/1593

Signed-off-by: Aleksei Zakharov <zaharov@selectel.ru>

### DIFF
--- a/bcache/get.go
+++ b/bcache/get.go
@@ -51,21 +51,21 @@ func NewFS(mountPoint string) (FS, error) {
 	return FS{&fs}, nil
 }
 
-// Stats is a wrapper around _Stats
+// Stats is a wrapper around stats()
 // It returns full available statistics
 func (fs FS) Stats() ([]*Stats, error) {
-	return fs._Stats(true)
+	return fs.stats(true)
 }
 
-// StatsWithoutPriority is a wrapper around _Stats.
+// StatsWithoutPriority is a wrapper around stats().
 // It ignores priority_stats file, because it is expensive to read.
 func (fs FS) StatsWithoutPriority() ([]*Stats, error) {
-	return fs._Stats(false)
+	return fs.stats(false)
 }
 
-// _Stats retrieves bcache runtime statistics for each bcache.
+// stats() retrieves bcache runtime statistics for each bcache.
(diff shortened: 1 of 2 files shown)


SuperQ:
Yes, I was thinking something similar, splitting up the methods so you can call them more granularly, rather than establish pattern of a lot of params across this library.

AlexZzz:
@pgier , thanks!
How about make them wrappers?

bcache/get.go
```diff
@@ -51,8 +51,21 @@ func NewFS(mountPoint string) (FS, error) {
...
+
+// StatsWithoutPriority is a wrapper around _Stats.
+// It ignores priority_stats file, because it is expensive to read.
+func (fs FS) StatsWithoutPriority() ([]*Stats, error) {
+	return fs._Stats(false)
+}
+
+// _Stats retrieves bcache runtime statistics for each bcache.
+// priorityStats flag controls if we need to read priority_stats.
+func (fs FS) _Stats(priorityStats bool) ([]*Stats, error) {
```
pgier: Go convention is to use mixed case instead of underscores for variable and function names, so this should be `stats` or `getStats` or something similar as long as it starts with a lower case letter.

bcache/get.go
```diff
@@ -251,7 +264,7 @@ func (p *parser) getPriorityStats() PriorityStats {
 }
 
 // GetStats collects from sysfs files data tied to one bcache ID.
-func GetStats(uuidPath string) (*Stats, error) {
+func GetStats(uuidPath string, priorityStats bool) (*Stats, error) {
```
pgier: This one is also an exported function, so it's possible that this change would break some users.  However, it looks like node_exporter doesn't use this directly, so I think this change should be ok.

AlexZzz: We can do wrappers here too, but for me it looks like adding complexity to the internal api. Also, not sure it will be used outside of this library.
Maybe, it is better to change it to getStats.

PR #273, opened 2020-03-12
Ignore infiniband files that report 'permission denied'

### LINKED ISSUE #272: netclass collector fails with permission errors on InfiniBand interfaces
Collector fails with errors such as:
```
prometheus-node-exporter[2255]: time="2020-03-02T17:32:12Z" level=error msg="ERROR: netclass collector failed after 0.011505s: could not get net class info: error obtaining net class info: failed to read file \"/sys/class/net/ib0/create_child\": open /sys/class/net/ib0/create_child: permission denied" source="collector.go:132"
```
As far as I can tell, this is because the collector ranges over all files in the directory, and tries to read their contents if they are a regular file. In the case of InfiniBand interfaces, there are two files which are not world-readable (and in fact not even intended to be read by root):
```
--w------- 1 root root 4096 Mai  7  2019 create_child
--w------- 1 root root 4096 Mär  2 17:41 delete_child
```
For those unfamiliar with InfiniBand, the `create_child` entry is used to create an IB subinterface, with specific P_key (aka partition - similar to an ethernet VLAN). The `delete_child` entry is for removing a subinterface.

The collector would need to perform an additional check besides the existing `f.Mode().IsRegular()` check, or more gracefully handle "permission denied" errors.

treydock:
Example files at `/sys/class/net/ib0/`:

```
--w-------  1 root root 4096 Mar 12 11:34 create_child
--w-------  1 root root 4096 Mar 12 11:34 delete_child
```

This would resolve #272 and https://github.com/prometheus/node_exporter/issues/1634

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -251,7 +251,7 @@ func parseInfiniBandCounters(portPath string) (*InfiniBandCounters, error) {
 		name := filepath.Join(path, f.Name())
 		value, err := util.SysReadFile(name)
 		if err != nil {
-			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
+			if os.IsNotExist(err) || os.IsPermission(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
 				continue
 			}
 			return nil, fmt.Errorf("failed to read file %q: %v", name, err)
@@ -334,7 +334,7 @@ func parseInfiniBandCounters(portPath string) (*InfiniBandCounters, error) {
 		name := filepath.Join(path, f.Name())
 		value, err := util.SysReadFile(name)
 		if err != nil {
-			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
+			if os.IsNotExist(err) || os.IsPermission(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
 				continue
 			}
 			return nil, fmt.Errorf("failed to read file %q: %v", name, err)
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -119,7 +119,7 @@ func (nc NetClass) parseNetClassIface(devicePath string) (*NetClassIface, error)
 		name := filepath.Join(devicePath, f.Name())
 		value, err := util.SysReadFile(name)
 		if err != nil {
-			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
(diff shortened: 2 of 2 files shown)


sysfs/class_infiniband.go
```diff
@@ -251,7 +251,7 @@ func parseInfiniBandCounters(portPath string) (*InfiniBandCounters, error) {
 		name := filepath.Join(path, f.Name())
 		value, err := util.SysReadFile(name)
 		if err != nil {
-			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
+			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" || err.Error() == "permission denied" {
```
dswarbrick:
```suggestion
			if os.IsNotExist(err) || os.IsPermission(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
```

sysfs/class_infiniband.go
```diff
@@ -334,7 +334,7 @@ func parseInfiniBandCounters(portPath string) (*InfiniBandCounters, error) {
...
```
dswarbrick:
```suggestion
			if os.IsNotExist(err) || os.IsPermission(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
```

sysfs/net_class.go
```diff
@@ -119,7 +119,7 @@ func (nc NetClass) parseNetClassIface(devicePath string) (*NetClassIface, error)
 		name := filepath.Join(devicePath, f.Name())
 		value, err := util.SysReadFile(name)
 		if err != nil {
-			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
+			if os.IsNotExist(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" || err.Error() == "permission denied" {
```
dswarbrick:
```suggestion
			if os.IsNotExist(err) || os.IsPermission(err) || err.Error() == "operation not supported" || err.Error() == "invalid argument" {
```

sysfs/class_infiniband.go
SuperQ: Yes, let's use the `os.IsPermission(err)` function.

treydock: Updated to use `os.IsPermission(err)`

sysfs/class_infiniband.go
treydock: os.IsPermission(err)

sysfs/net_class.go
treydock: Updated to use `os.IsPermission(err)`

treydock:
Now using `os.IsPermission(err)` rather than previously checking error text.

PR #277, opened 2020-03-31
Bcache: add bdev/writeback_rate_debug stats

AlexZzz:
Hi! This PR adds support for bache writeback_rate_debug stats.
It's called debug, but it has useful information. It shows current writeback rate, target size of dirty data and PI-controller values. It is very usefull for debugging and helps to make decision to tweak bcache's PI-controller.
@pgier @discordianfish , could you approve this?


discordianfish:
@pgier @SuperQ

PR #281, opened 2020-04-07
Add ProcSMaps from smaps file

PierreF:
The smaps file contains information not available though existing source.

The one that interest me here is the "Pss" and "SwapPss" which is the process proportional usage of resident memory and swap respectively (which is I think the pages size divided by the number of process that share those pages).

I need this to expose a more significant memory usage for a group of process for prometheus process-exporter (https://github.com/ncabatoff/process-exporter/issues/140)

### DIFF
--- a/proc_smaps.go
+++ b/proc_smaps.go
@@ -1,0 +1,246 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package procfs
+
+import (
+	"bufio"
+	"os"
+	"strconv"
+	"strings"
+)
+
+type ProcSMaps []*ProcSMap
+
+type ProcSMap struct {
+	*ProcMap
+	// Size of the mapping
+	Size uint64
+	// Amount of the mapping that is currently resident in RAM
+	Rss uint64
+	// Process's proportional share of this mapping
+	Pss uint64
+	// Size in bytes of clean shared pages
+	SharedClean uint64
+	// Size in bytes of dirty shared pages
+	SharedDirty uint64
+	// Size in bytes of clean private pages
+	PrivateClean uint64
+	// Size in bytes of dirty private pages
+	PrivateDirty uint64
+	// Amount of memory currently marked as referenced or accessed
+	Referenced uint64
+	// Amount of memory that does not belong to any file
+	Anonymous uint64
+	// Amount would-be-anonymous memory currently on swap
+	Swap uint64
+	// Process's proportional memory on swap
+	SwapPss uint64
+	// Page size used by the kernel to back the virtual memory area
+	KernelPageSize uint64
+	// Page size used by the processor MMU to back the virtual memory area
+	MMUPageSize uint64
+	// Kernel flags associated with the virtual memory area
+	VMFlags string
+}
+
+// SizeSum returns the sum of Pss from all mappings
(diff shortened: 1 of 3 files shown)


proc_smaps.go
```diff
@@ -0,0 +1,246 @@
+// Copyright 2018 The Prometheus Authors
```
pgier, line 1: Should be 2020, maybe someday we'll set up some automation for these notices.

proc_smaps.go
```diff
@@ -0,0 +1,246 @@
...
+// PssSum returns the sum of Pss from all mappings
+func (s ProcSMaps) PssSum() (sum uint64) {
+	for _, x := range s {
+		sum += x.Pss
+	}
+
+	return
+}
+
+// SharedCleanSum returns the sum of Pss from all mappings
```
pgier, line 86: Looks like a copy paste typo on these comments since they all say `Pss`.  Maybe just set the comments to `...sum of all mappings` since it should be clear anyway which field is being summed.

proc_smaps.go
```diff
@@ -0,0 +1,246 @@
...
+	var currentMap *ProcSMap
+
+	smaps := []*ProcSMap{}
+	scan := bufio.NewScanner(file)
+
+	for scan.Scan() {
+		line := scan.Text()
+		// first line of a mapping start with an hexadecimal address in lower-case
+		// All other line start with a capitalized words.
+		if (line[0] >= '0' && line[0] <= '9') || (line[0] >= 'a' && line[0] <= 'f') {
```
pgier, line 176: I think it would be more clear hear to do a simple regex match.

proc_smaps.go
```diff
@@ -0,0 +1,246 @@
...
+		kv := strings.SplitN(line, ":", 2)
+		if len(kv) != 2 {
+			continue
+		}
+
+		k := kv[0]
+		v := strings.TrimSpace(kv[1])
+		v = strings.TrimRight(v, " kB")
+
+		vKBytes, _ := strconv.ParseUint(v, 10, 64)
```
pgier, line 202: It might be better to check this error and fail if it's non-nil unless there is reason to ignore the parse failures.

proc_smaps_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2018 The Prometheus Authors
```
pgier, line 1: Should be 2020.

proc_smaps.go
PierreF: I've kept is as an if condition for performance reason.
I've added a benchmark and pushed https://github.com/PierreF/procfs/commit/ff6dfede807f15a2d7ad605249ad41aa93184389 which use regexp, but it has notable performance impact:
```
$ git co 6481fbb # commit just before using regexp
$ for i in `seq 0 10`; do go test -run=none -bench=.* ./ ;done > bench.6481fbb
$ git co ff6dfed # commit using regexp
$ for i in `seq 0 10`; do go test -run=none -bench=.* ./ ;done > bench.ff6dfed
$ go get golang.org/x/perf/cmd/benchstat
$ benchstat bench.6481fbb bench.ff6dfed 
name         old time/op  new time/op  delta
ProcSmaps-8   200µs ± 8%   270µs ± 5%  +34.52%  (p=0.000 n=11+10)
```

I think this the less clear hand-made condition worse the performance gain, but the decision is yours.

PierreF:
Fix comments. About the regexp, I've kept the if condition for performance reason. If you want to go the the regexp version, I'll update this PR to include the commit (https://github.com/PierreF/procfs/commit/d9c39b84ce211fcf1830aaade168bfcf4f65509e)

SuperQ:
FYI, we might want to use [`smaps_rolloup`](https://www.kernel.org/doc/Documentation/ABI/testing/procfs-smaps_rollup) rather than `smaps`. It's a lighter-weight interface to the PSS data. But requires Linux >= 4.15.x

PierreF:
This looks very interesting... and exactly match what I need for ncabatoff/process-exporter#140.
I'll propose a new PR based on this new file, with fallback on smaps, which will only expose the rollup.

PierreF:
Created #286 which use smaps_rollup. I think #286 is better: it's faster (on the benchmark, and likely even more in real usage since the Kernel don't need to provide detailed information) and should provide enough information for most usage.

discordianfish:
Ok lets close this in favor of #286 for now.

PR #282, opened 2020-04-17
Fix mountinfo parsing to handle entries with 10 fields

treydock:
See https://github.com/prometheus/node_exporter/pull/1673 for the full context of how this is impacting node_exporter.


treydock:
Any possibility of getting this merged so that node_exporter can incorporate this fix?

discordianfish:
@pgier @SuperQ

PR #285, opened 2020-05-01
Bump go versions to 1.13 and 1.14

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>


roidelapluie:
Should we document the go versions that we agree to support?

PR #286, opened 2020-05-01
Add smaps information from /proc/pid/smaps_rollup

PierreF:
Follow-up of #281 

Unlike #281, this one only return the sum of value, either from smaps_rollup or (if file does not exists) from smaps.

The information that interest me here is the "Pss" and "SwapPss" which is the process proportional usage of resident memory and swap respectively (which is I think the pages size divided by the number of process that share those pages).

I need this to expose a more significant memory usage for a group of process for prometheus process-exporter (ncabatoff/process-exporter#140)

Signed-off-by: Pierre Fersing <pierre.fersing@bleemeo.com>

### DIFF
--- a/proc_smaps.go
+++ b/proc_smaps.go
@@ -28,7 +28,8 @@ import (
 )
 
 var (
-	rFirstLine = regexp.MustCompile(`^[a-f0-9].*$`)
+	// match the header line before each mapped zone in /proc/pid/smaps
+	procSMapsHeaderLine = regexp.MustCompile(`^[a-f0-9].*$`)
 )
 
 type ProcSMapsRollup struct {
@@ -100,7 +101,7 @@ func (p Proc) procSMapsRollupManual() (ProcSMapsRollup, error) {
 	for scan.Scan() {
 		line := scan.Text()
 
-		if rFirstLine.MatchString(line) {
+		if procSMapsHeaderLine.MatchString(line) {
 			continue
 		}
 
--- a/proc_smaps.go
+++ b/proc_smaps.go
@@ -121,13 +121,15 @@ func (s *ProcSMapsRollup) parseLine(line string) error {
 	}
 
 	k := kv[0]
+	if k == "VmFlags" {
+		return nil
+	}
+
 	v := strings.TrimSpace(kv[1])
 	v = strings.TrimRight(v, " kB")
 
 	vKBytes, err := strconv.ParseUint(v, 10, 64)
-
-	// VmFlags is the only field which is not a number, ignore parse error for it.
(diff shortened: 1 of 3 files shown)


proc_smaps.go
```diff
@@ -0,0 +1,162 @@
...
+	// Process's proportional memory on swap
+	SwapPss uint64
+}
+
+// ProcSMaps reads from /proc/[pid]/smaps_rollup to get summed memory information of the
+// process.
+//
+// If smaps_rollup does not exists (require kernel >= 4.15), the content of /proc/pid/smaps will
+// we read and summed.
+func (p Proc) ProcSMaps() (ProcSMaps, error) {
```
pgier, line 62: Maybe we should call this function `ProcSMapsRollup` just to make it more clear which file it's normally reading, and also to allow for a possible future function `ProcSMaps` which just returns the data from `/proc/<pid>/smaps` without summing.

proc_smaps.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	defer file.Close()
+
+	smaps := ProcSMapsRollup{}
+	scan := bufio.NewScanner(file)
+
+	for scan.Scan() {
+		line := scan.Text()
+
+		if rFirstLine.MatchString(line) {
```
discordianfish, line 103: If this is always the first line, why does it need a regex?
If it's not always the first line, the variable name is misleading.

Either way, since it's a global variable, it should be eited name procSMapsSomething (or something generic if it's suppose to be used by other functions as well)

proc_smaps.go
```diff
@@ -0,0 +1,162 @@
...
+		return errors.New("invalid net/dev line, missing colon")
+	}
+
+	k := kv[0]
+	v := strings.TrimSpace(kv[1])
+	v = strings.TrimRight(v, " kB")
+
+	vKBytes, err := strconv.ParseUint(v, 10, 64)
+
+	// VmFlags is the only field which is not a number, ignore parse error for it.
```
discordianfish, line 128: Why not just `if k == "VmFlags" { return }` before trying to parse?
Either way, checking this first before calling ParseUint seems cleaner.

discordianfish:
Beside the comments LGTM.

proc_smaps.go
PierreF: This was a left-over from previous code that kept VmFlags (because it kept each mapping in #281).
Now that we only kept summary, it's indeed better to return when k == VmFlags.

proc_smaps.go
PierreF: Improved the name (and added a small comment).
It's the first line of each mapping zones and /proc/pid/smaps contains multiple zones.

PR #288, opened 2020-05-06
cpuinfo: fix platform constraints

pgier:
Fix issues with platform constraints found by @mfraioli

### DIFF
--- a/cpuinfo_default.go
+++ b/cpuinfo_default.go
@@ -12,7 +12,8 @@
 // limitations under the License.
 
 // +build linux
+// +build 386 amd64
 
 package procfs
 
-var parseCPUInfo = parseCPUInfoS390X
+var parseCPUInfo = parseCPUInfoX86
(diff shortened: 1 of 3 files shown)


discordianfish:
Ugh, I wish we had a way to test this.

mfraioli:
I have access to ppc and s390 build systems that can verify it. But since we're picking this up as a dependency of a dependency, it's hard for me to pick up a branch rather than the master. If you go ahead and merge it I can verify quickly. Since it's already broken, this at least shouldn't make it any worse.

mfraioli:
For reference description of the issues is in #257

mfraioli:
Thanks @pgier this built successfully on my ppc64le and s390x systems!

discordianfish:
Yeah manual verification is one thing but maybe it's worth investing in a qemu based setup to test all supported platforms. But after spending some hours just trying to get freebsd/x86 working I kinda gave up on that.

PR #289, opened 2020-05-08
add flush request and time spend flushing diskstats for kernel 5.5+

binjip978:


blockdevice/stats.go
```diff
@@ -153,13 +158,15 @@ func (fs FS) ProcDiskstats() ([]Diskstats, error) {
...
+			&d.TimeSpentFlushing,
 		)
 		// The io.EOF error can be safely ignored because it just means we read fewer than
-		// the full 18 fields.
+		// the full 20 fields.
 		if err != nil && err != io.EOF {
 			return diskstats, err
 		}
-		if d.IoStatsCount == 14 || d.IoStatsCount == 18 {
+		if d.IoStatsCount == 14 || d.IoStatsCount == 18 || d.IoStatsCount == 20 {
```
discordianfish, line 169: I don't like this condition. I'd rather check if it's at least 14 or just get rid of it altogether.

discordianfish: @pgier @SuperQ thoughts?

pgier: Yeah, I think that would be better assuming that sscanf doesn't fail on more than 20 fields.

PR #290, opened 2020-05-08
Add missing cpuinfo support for armv6l

tobiasgies:
Dear maintainer,

PR #257 added basic multiplatform support for cpuinfo.

Due to this change, this project does not compile on 32-bit ARM CPUs, e.g. the armv7l in a Raspberry Pi 3b. I think this PR should solve this problem.

### DIFF
--- a/cpuinfo_armv6l.go
+++ b/cpuinfo_armv6l.go
@@ -1,0 +1,19 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


tobiasgies:
Hi @mfraioli, @discordianfish, @SuperQ – you approved #257 and the follow-up #288. Could you have a look at this one?

pgier:
I'm wondering if we should add a default cpuinfo flag which just returns an error about being an unsupported platform.  That way at least the library would compile on the unknown platforms instead of failing to compile.

tobiasgies:
Good question. I actually prefer a failure to compile to finding out the library won't work at runtime, if I can. In this case, would love it if the error message could be somehow made clearer than just "cpuInfo? Never 'heard of it".

discordianfish:
Yeah not sure what is better. Let's discuss that in an issue or future PR though, this LGTM

PR #292, opened 2020-05-12
cpuinfo: Enable for all 32-bit ARM CPUs

### LINKED ISSUE #291: No CPUInfo for all ARM
Hi,

There has been changes in the CpuInfo part in the recent past, and e.g. https://github.com/prometheus/procfs/pull/290 adds some support for some ARM CPUs. However, that's only for v6.

Could we maybe enable parseCPUInfoARM for all ARM versions?

bluecmd:
Repurpose the ARMv6 cpuinfo for all ARM CPUs. CPU info format likely is
not dependent on instruction set version.

Signed-off-by: Christian Svensson <bluecmd@google.com>

Fixes #291

### DIFF
--- a/cpuinfo_arm.go
+++ b/cpuinfo_arm.go
@@ -12,7 +12,6 @@
 // limitations under the License.
 
 // +build linux
-// +build armv6l
 
 package procfs
 


bluecmd:
This fixes the compilation for ARM 32-bit, but the current parsing code only works for kernels older than v3.8.

I think there is value to merging this while the format question is resolved - if nothing else to allow 32-bit ARM to build.

WDYT @discordianfish ?

SuperQ:
One thing we could do here is to build a set of tests that run on different architectures/kernels. This could be done by adding to the buildkite setup we have for the node_exporter.

With buildkite, we run a few different build agents under KVM/QEMU, so we're not dependent on the limited platforms available on CircleCI.

discordianfish:
Fixing the parsing itself can be done in another PR, so LGTM.

Issue #294, opened 2020-05-15
CPUInfo for ARM and Linux kernels newer than 3.8


bluecmd:
From #291 it was concluded that the current parsing code is limited to kernels [older than 3.8](https://github.com/torvalds/linux/commit/b4b8f770eb10a1bccaf8aa0ec1956e2dd7ed1e0a).

Problematic code is:
https://github.com/prometheus/procfs/blob/7a44272e0d68eea081a6f7e08430150a286eb734/cpuinfo.go#L189-L233

This code should be changed to either drop support for older kernels, or be capable of parsing both new and old kernels alike - probably with reduced functionality for the older ones.

pgier:
I believe this is resolved now with the merge of #296

PR #296, opened 2020-05-18
Added support for CPUInfo for ARM on Linux kernels 3.8+

kadern0:
Fixes issue #294

Hi @bluecmd, please have a look and let me know if you think I should do any changes on the PR. Thanks!

### DIFF
--- a/cpuinfo.go
+++ b/cpuinfo.go
@@ -190,15 +190,26 @@ func parseCPUInfoARM(info []byte) ([]CPUInfo, error) {
 	scanner := bufio.NewScanner(bytes.NewReader(info))
 
 	firstLine := firstNonEmptyLine(scanner)
-	if !strings.HasPrefix(firstLine, "Processor") || !strings.Contains(firstLine, ":") {
+	match, _ := regexp.MatchString("^[Pp]rocessor", firstLine)
+	if !match || !strings.Contains(firstLine, ":") {
 		return nil, errors.New("invalid cpuinfo file: " + firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
-	commonCPUInfo := CPUInfo{VendorID: field[1]}
-
 	cpuinfo := []CPUInfo{}
-	i := -1
 	featuresLine := ""
+	commonCPUInfo := CPUInfo{}
+	i := 0
+	if strings.TrimSpace(field[0]) == "Processor" {
+		commonCPUInfo = CPUInfo{ModelName: field[1]}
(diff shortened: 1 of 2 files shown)


cpuinfo.go
```diff
@@ -190,15 +190,26 @@ func parseCPUInfoARM(info []byte) ([]CPUInfo, error) {
...
 	field := strings.SplitN(firstLine, ": ", 2)
-	commonCPUInfo := CPUInfo{VendorID: field[1]}
-
 	cpuinfo := []CPUInfo{}
-	i := -1
 	featuresLine := ""
+	commonCPUInfo := CPUInfo{}
+	i := 0
+	if field[0] == "Processor " {
+		commonCPUInfo = CPUInfo{VendorID: field[1]}
```
bluecmd, line 203: Vendor ID in the old format seems to become Model ID in the new. Sticking to one seems like a good idea, possibly Model as that's spelled out in the new format. Wdyt?

cpuinfo.go
```diff
@@ -190,15 +190,26 @@ func parseCPUInfoARM(info []byte) ([]CPUInfo, error) {
...
+		i = -1
```
bluecmd, line 204: Sadly the format appears to be a bit more complicated. If you look at the [commit](https://github.com/torvalds/linux/commit/b4b8f770eb10a1bccaf8aa0ec1956e2dd7ed1e0a) you see that the presence of "processor" was conditional on CONFIG_SMP. In other words, for cases where you only have one core it's omitted.

I found this example of an old cpuinfo from a Raspberry Pi:

```
pelicone ~ # cat /proc/cpuinfo
model name : ARMv6-compatible processor rev 7 (v6l)
Features : half thumb fastmult vfp edsp java tls
CPU implementer : 0x41
CPU architecture: 7
CPU variant : 0x0
CPU part : 0xb76
CPU revision : 7

Hardware : BCM2708
Revision : 0010 
```

cpuinfo.go
kadern0: I saw this change but my approach was trying to leave the old code as it was. As you say, I also prefer having one format instead of two. Let's use Model then.

cpuinfo.go
kadern0: Didn't check the 'legacy' format, just focused on the newer one so I totally missed this. I'm going to add this example to the test cases and ensure the code also supports it. Thanks!

kadern0:
@bluecmd I think it is ready for a second round :)

PR #297, opened 2020-05-18
cpuinfo: fixes and enhancements

kolyshkin:
This is a set of commits aiming to improve cpuinfo parsers.

Please review commit by commit. Below is a copy-paste of individual commit descriptions.

###  cpuinfo_test: fix test data
    
    Real /proc/cpuinfo files do not contain empty lines at the start
    or at the end of file.

###  parseCPUInfoARM: fix for kernel 3.7+
    
    Since Linux kernel 3.7 (commit v3.7-rc5-2-gb4b8f770eb10 "ARM: kernel:
    update cpuinfo to print all online CPUs features", Sep 10 2012)
    the kernel does not print "Processor" as the first line of
    /proc/cpuinfo. Instead, it adds a line named "model name"
    with similar contents.
    
    Fix the code for this case, and add a unit test.

### parseCPUInfoARM: only parse Features once
    
    The `Features` line is the same for all CPUs, and yet it is assigned to
    and parsed many times.
    
    Optimize the code so it is only parsed once and assigned in place.

### parseCPUInfo: optimize line scan
    
    Avoid double check of each line for ":" -- since we're splitting the line anyway,
    we can reuse the result.
    
    This also prevents a potential panic of referencing non-existen field[1]
    in case input contains a ":" with no space after.

### cpuinfo: read data in place
    
    Instead of reading the data to the buffer, then making a reader and a
    scanner out of that buffer, let's analyze the data as we read it line by
    line.

### parseCPUInfo*: add error checking
    
    In case the input can't be read, scanner.Scan() stops and the code is
    supposed to check for error using scanner.Err(). Do that.
    
    In other words, instead of returning half-read cpuinfo, return an error.

### cpuinfo: rm GetFirstNonEmptyLine, simplify scan
    
    There is no sense to skip empty lines at the beginning of the file since
    there are none (there were in test data but it is fixed by an earlier
    commit).
    
    Logic is simpler now.


cpuinfo.go
```diff
@@ -207,49 +190,53 @@ func parseCPUInfoARM(info []byte) ([]CPUInfo, error) {
 		}
 		field := strings.SplitN(line, ": ", 2)
 		switch strings.TrimSpace(field[0]) {
+		case "Processor":
+			// this appears at the first line in kernels < 3.7
+			modelName = field[1]
 		case "processor":
-			cpuinfo = append(cpuinfo, commonCPUInfo) // start of the next processor
+			cpuinfo = append(cpuinfo, CPUInfo{VendorID: modelName}) // start of the next processor
```
bluecmd, line 197: See this comment on a similar current PR: https://github.com/prometheus/procfs/pull/296/files#discussion_r427036662

kolyshkin: Thanks, fixed!

bluecmd: The link appears to be not as helpful as I wanted, but the conversation was about that "processor" is only present for CONFIG_SMP (pre v3.8) kernels, so you likely need to figure out how to support kernels without that option.

kolyshkin: Yeah, I think I have figured it out, please see the updated commit 928d8ea1fea240, to which I have added a test case with data from the above comment.

bluecmd: 👍 Cool!

kolyshkin:
@discordianfish @pgier PTAL

cpuinfo.go
kolyshkin: @bluecmd can you review it then? I am aware you're not a maintainer but you wrote some of this code.

cpuinfo.go
```diff
@@ -64,36 +63,26 @@ var (
...
 		return nil, err
 	}
-	return parseCPUInfo(data)
+	defer file.Close()
+	return parseCPUInfo(file)
 }
 
-func parseCPUInfoX86(info []byte) ([]CPUInfo, error) {
-	scanner := bufio.NewScanner(bytes.NewReader(info))
+func parseCPUInfoX86(info io.Reader) ([]CPUInfo, error) {
```
discordianfish, line 74: I think we decided to prefer reading everything in a buffer first. While this is unlikely to change mid reading, I think it's what we decided to to in general. But not sure, @SuperQ wdyt?

cpuinfo.go
bluecmd: I didn't write any of this code, I just filed bugs against it :-)

cpuinfo.go
```diff
@@ -183,22 +172,16 @@ func parseCPUInfoX86(info []byte) ([]CPUInfo, error) {
 			cpuinfo[i].PowerManagement = field[1]
 		}
 	}
-	return cpuinfo, nil
+	return cpuinfo, scanner.Err()
```
pgier, line 175: If `scanner.Err()` is not nil here, I'm wondering if we should return a nil instead of cpuinfo.  At least that's the pattern we've usually followed for other parts of this library.

cpuinfo.go
pgier: We have been doing that previously, although I think this pattern is fine also.  The OS is probably buffering the file content anyway.  And I think `io.Reader` is better to use as a parameter for most `parseXXX` functions because it's a little more generic than a string or byte slice.

If we do want to buffer the entire file first, then it would probably be better to create a reader from the byte slice and then pass the reader to the `parse` function instead of passing the byte slice directly.  The byte slice parameter here might just be left over from me thinking I would need random access instead of line by line access to parse certain file formats.

cpuinfo.go
discordianfish: I think this is fine.

cpuinfo.go
discordianfish: Ok sounds reasonable, then let's keep is a reader.

discordianfish:
Yeah needs rebasing but also LGTM in general.

cpuinfo.go
kolyshkin: > if scanner.Err() is not nil here, I'm wondering if we should return a nil instead of cpuinfo. At least that's the pattern we've usually followed for other parts of this library.

It seems like an unnecessary complication. A user should always check for the error first. If the error is non-nil, there are no guarantees wrt what the result would be.

kolyshkin:
> I think this will have a couple minor conflicts with #296, so this might need one more update after that one is merged.

Alas, this PR also fixes the same issue as #296 so it did not have to be merged first.

Anyway, let me rework this on top of it.

cpuinfo.go
pgier: Ok makes sense, I'm fine with this as-is.

SuperQ:
Any interest in continuing this PR?

Issue #266, opened 2020-02-21
Add parser for /sys/block/<device>/queue


SuperQ:
https://www.kernel.org/doc/Documentation/block/queue-sysfs.txt

There are a number of useful files in this subsystem.

Related to https://github.com/prometheus/node_exporter/issues/1601

binjip978:
Would like to help with this one and have two questions:
It will be ok if we will return queue stats for particular device, like this?
```go
func (fs FS) SysBlockDeviceQueueStats(device string) (BlockQueueStats, error) {}
```
And for optional fields like `throttle_sample_time` it's ok to use pointer in BlockQueueStats structure?

pgier:
I believe this is resolved with the merge of #299

PR #299, opened 2020-05-20
parse sys/block/xxx/queue

binjip978:

### DIFF
--- a/blockdevice/stats.go
+++ b/blockdevice/stats.go
@@ -16,6 +16,7 @@ package blockdevice
 import (
 	"bufio"
 	"fmt"
+	"github.com/prometheus/procfs/internal/util"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


binjip978:
should fix https://github.com/prometheus/procfs/issues/266

blockdevice/stats.go
```diff
@@ -220,3 +310,190 @@ func (fs FS) SysBlockDeviceStat(device string) (IOStats, int, error) {
 	}
 	return stat, count, err
 }
+
+func (fs FS) SysBlockDeviceQueueStats(device string) (BlockQueueStats, error) {
+	stat := BlockQueueStats{}
+	addRandom, err := util.ReadUintFromFile(fs.sys.Path(sysBlockPath, device, sysBlockQueue, "add_random"))
```
discordianfish, line 316: What about making the stat struct members all pointers and then use a loop like:
```
for file, val := range map[string]*uint64 {
  "add_random": stat.AddRandom,
  ...
} {
  val, err := util.ReadUintFromFile(fs.sys.Path(sysBlockPath, device, sysBlockQueue, file)
  ...
```

blockdevice/stats_test.go
```diff
@@ -104,4 +104,116 @@ func TestBlockDevice(t *testing.T) {
 	if device1stats.DiscardTicks != 12 {
 		t.Errorf(failMsgFormat, "Incorrect discard ticks", 12, device1stats.DiscardTicks)
 	}
+	blockQueueStat, err := blockdevice.SysBlockDeviceQueueStats(devices[1])
```
discordianfish, line 107: Ideally we should use a loop here as well or just a expected struct and use DeepEqual or somtehing.

blockdevice/stats.go
binjip978: not all fields are uint64, two of them are int64 and another two are strings and one is a *uint64, but it possible to factor all unit64 in such way and handle others as special cases.

blockdevice/stats_test.go
binjip978: yep, DeepEqual sound like a good idea.

blockdevice/stats.go
```diff
@@ -86,11 +87,100 @@ type Diskstats struct {
 	IoStatsCount int
 }
 
+// BlockQueueStats models the queue files that are located in the sysfs tree for each block device.
```
pgier, line 90: Is there any online documentation we can link to for details about these fields?

blockdevice/stats.go
```diff
@@ -220,3 +310,87 @@ func (fs FS) SysBlockDeviceStat(device string) (IOStats, int, error) {
 	}
 	return stat, count, err
 }
+
```
pgier, line 313: Should probably add a godoc comment here since this is an exported function.

blockdevice/stats.go
```diff
@@ -220,3 +310,87 @@ func (fs FS) SysBlockDeviceStat(device string) (IOStats, int, error) {
...
+		"write_same_max_bytes":   &stat.WriteSameMaxBytes,
+		"nr_zones":               &stat.NRZones,
+		"chunk_sectors":          &stat.ChunkSectors,
+		"fua":                    &stat.FUA,
+		"max_discard_segments":   &stat.MaxDiscardSegments,
+		"write_zeroes_max_bytes": &stat.WriteZeroesMaxBytes,
+	} {
+		val, err := util.ReadUintFromFile(fs.sys.Path(sysBlockPath, device, sysBlockQueue, file))
+		if err != nil {
+			return stat, err
```
pgier, line 350: If there is an error here we should return a `nil` instead of a partially filled `stat`.

pgier: For some reason I was thinking this was returning a pointer to a struct so you could return a literal `nil`.  But since this is returning the struct itself I guess it would have been fine to return it as is.

binjip978: Anyway it will be discarded if error is not nil, so i don't think there is much difference what to return in this case.

PR #305, opened 2020-05-28
parse /proc/sys/kernel/random

binjip978:


binjip978:
For "Expose entropy poolsize" https://github.com/prometheus/node_exporter/issues/1715
not sure about making all fields optional, but at least read_wakeup_threshold is missing on my fedora 5.6.14-300.fc32.x86_64

PR #308, opened 2020-06-13
Remove error wrapper from /sys/class/power_supply

SuperQ:
Don't wrap the error of `ioutil.ReadDir()` so that the error return can
be tested with `os.IsNotExist()`. Fixes node_exporter log noise.

```
caller=collector.go:161 msg="collector failed" name=powersupplyclass duration_seconds=0.00079848 err="could not get power_supply class info: error obtaining power_supply class info: failed to list power supplies at \"/sys/class/power_supply\": open /sys/class/power_supply: no such file or directory"
```

### DIFF
--- a/sysfs/class_power_supply.go
+++ b/sysfs/class_power_supply.go
@@ -110,7 +110,7 @@ func (fs FS) PowerSupplyClass() (PowerSupplyClass, error) {
 
 	dirs, err := ioutil.ReadDir(path)
 	if err != nil {
-		return nil, fmt.Errorf("failed to list power supplies at %q: %v", path, err)
+		return nil, err
 	}
 
 	psc := make(PowerSupplyClass, len(dirs))


SuperQ:
This is similar to https://github.com/prometheus/procfs/pull/302.

roidelapluie:
The node exporter is go1.14. can't we use errors.Is ?

SuperQ:
Oh, I didn't know about that one. Yes, we already require Go 1.14 for https, so that should be fine.

SuperQ:
I tried that out on a test build, it doesn't seem to work.

roidelapluie:
Indeed. It requires the fmt.Errors to use %w instead of %v for the error. Do you think this is worth it? Would make procfs 1.13+

SuperQ:
Hrm, maybe good to start using it upstream, but for now let's un-wrap the error. We want to keep procfs compatible with what client_golang needs for Go versions.

PR #312, opened 2020-06-22
Fix missing parseCPUInfo func on unknown architectures

zhsj:
Add a dummy function, so it doesn't fail to build.

Also group same files into one.

It's tested by:

```
 for i in `go tool dist list|grep linux|cut -d'/' -f2`; do GOARCH=$i go build .; done;
```


SuperQ:
I think we might prefer to fail build rather than return a dummy. It seems like the only thing we're currently missing is a `riscv64` function.

zhsj:
Though I'm not sure if there are real people using this library(or even Go) on other architectures, there is gccgo which supports more than gc. And they maybe don't need to use `CPUInfo` func, they just need to use other functions in this library.

Anyway adding a riscv64 function is also great.

zhsj:
To address the concern for "whether the platform is supported", `parseCPUInfoDummy` can be changed to `return nil, errors.New("not implemented")`

SuperQ:
Yes, returning `errors.New("not implemented")` would be preferred. That sounds fine for the dummy function.

discordianfish:
Thanks!

PR #313, opened 2020-06-22
Fix running test on arm64

zhsj:
On arm64, the CPUInfo in TestCPUInfoX86 calls parseCPUInfoARM,
so it panics.

Log: https://ci.debian.net/data/autopkgtest/testing/arm64/g/golang-procfs/5986092/log.gz

### DIFF
--- a/cpuinfo_test.go
+++ b/cpuinfo_test.go
@@ -207,6 +207,7 @@ machine		: CHRP IBM,8233-E8B
 )
 
 func TestCPUInfoX86(t *testing.T) {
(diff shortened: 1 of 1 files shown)


discordianfish:
I think it would make sense to set this in the package scope (aka move it two lines up). What do you think?

zhsj:
> I think it would make sense to set this in the package scope (aka move it two lines up). What do you think?

There are other `TestCPUInfoParse<CPU>` functions in this file, IMO this line `parseCPUInfo = parseCPUInfoX86` should be in `TestCPUInfoX86`.

PR #316, opened 2020-06-30
Add NetStat to parse /proc/net/stat/...

AlexZzz:
This PR is a step to expose arp/ndisc/route cache stats from `/proc/net/stat/`. It was described in the issue https://github.com/prometheus/node_exporter/issues/1626.

I decided to make it similar to [lnstat](https://github.com/shemminger/iproute2/blob/master/misc/lnstat.c) tool, so it will parse all new files in `/proc/net/stat` in future.

### DIFF
--- a/lnstat.go
+++ b/lnstat.go
@@ -1,0 +1,72 @@
+// Copyright 2020 Aleksei Zakharov
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lnstat
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"strconv"
+	"strings"
+  "path/filepath"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+type Lnstats struct {
+  Filename string
+  Name string
+  Value uint64
+  CPU uint64
+}
+
+func Lnstat() (Lnstats[], error) {
+  statFiles, err := filepath.Glob(fs.proc.Path("net/stat/*"))
+  if err != nil {
+    return err
+  }
+
+  var lnstats Lnstats[]
+
+  for filePath range statsFiles {
+    file, err := os.Open(filePath)
+    if err != nil {
+      return nil, err
+    }
+
+    scanner := bufio.NewScanner(file)
+    scanner.Split(bufio.ScanLines)
+    scanner.Scan()
+    for _, header := range strings.Fields(scanner.Text()) {
+      lnstat := Lnstats {
+        Filename: filepath.Base(filePath),
+        Name: header,
+      }
+      lnstats = append(lnstats. lnstat)
+    }
+
+    var cpu uint32 = 0
+    for scanner.Scan() {
+      for num, counter := range strings.Fields(scanner.Text()) {
+        lnstats[num].CPU = cpu
+        lnstats[unm].Value, err = strconv.ParseUint(counter, 16, 32)
+        if err != nil {
+          return nil, err
+        }
+      }
+      cpu++
+    }
+  }
+  return lnstats, nil
+}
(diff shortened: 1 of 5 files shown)


lnstat.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+			lnstat.Value = make(map[uint64]uint64)
+			lnstatsOnce = append(lnstatsOnce, lnstat)
+		}
+
+		// Other strings represent per-CPU counters
+		var cpu uint64
+		for scanner.Scan() {
+			for num, counter := range strings.Fields(scanner.Text()) {
+				lnstatsOnce[num].Value[cpu], err = strconv.ParseUint(counter, 16, 32)
```
pgier, line 69: Since cpu is always just an index, maybe this can be a slice instead of a map?

lnstat.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		var lnstatsOnce []Lnstats
+		scanner := bufio.NewScanner(file)
+		scanner.Split(bufio.ScanLines)
+		scanner.Scan()
+		// First string is always a header for stats
+		for _, header := range strings.Fields(scanner.Text()) {
+			lnstat := Lnstats{
+				Filename: filepath.Base(filePath),
```
pgier, line 58: The filename will be repeated for each header.  I'm wondering if the struct can be changed to allow a single filename and multiple rows of data.

AlexZzz:
Thank you for the review!
I've changed data structure, also added a test.

AlexZzz:
I thought about the name, but there's `/proc/net/netstat` file already, and it is used by node-exporter `netstat` collector, so `netstats` name will add mess, IMO.

It could be named `netcache`, or something related to network-and-cache. But if there is a new file under `/proc/net/stat` and it is not about cache, it will be confusing again:)

discordianfish:
Oh right, hrm yeah I guess lnstat is fine then - unless somebody else has a better idea. @SuperQ @pgier

AlexZzz:
@pgier @SuperQ , any suggestions?

SuperQ:
We still need to port over the `/proc/net/netstat` (and snmp/snmp6) parsing from the node_exporter. So we don't have a naming conflict to worry about. We do have a bit of a precedent for `/proc/net/sockstat` -> `NetSockstat`. I wish there was some documentation about this outside of the `lnstat` man page. I can't find anything in the usual kernel.org/doc.

Maybe _slightly_ confusing, but how about we follow this pattern:

* `/proc/net/netstat` -> `NetNetstat`
* `/proc/net/stat/*` -> `NetStat`

AlexZzz:
>     * `/proc/net/netstat` -> `NetNetstat`
>     * `/proc/net/stat/*` -> `NetStat`
I think this looks good, because it looks like a naming convention really: name consists of path. I'll rename `Lnstat` to `NetStat` then. And here we don't care about the name of the collector, right? I mean, we're leaving `lnstat` and `netstat` collectors names and they'll work with `NetStat` and `NetNetstat` structs, right?

SuperQ:
If you're talking about the node_exporter collector name, Yes, we could call it `lnstat` to match the normal user utility. That seems right to me. What do you think @discordianfish?

SuperQ:
When digging through some other issues/PRs, I discovered we already have some partial parsing of `/proc/net/stat` in https://github.com/prometheus/procfs/pull/254.

How do we want to handle this? We have https://github.com/prometheus/node_exporter/pull/1155 open to add handling of `nf_conntrack` to the node_exporter, but this would be duplicated by `lnstat`.

@pgier Should we use this generic parser instead of the conntrack parser?

AlexZzz:
@pgier , @SuperQ , how is it going?
I would like to vote for this generic parser:)

discordianfish (approved):
Ok, let's merge this. We can make net_conntrackstat.go use the parser here.

AlexZzz:
@discordianfish , btw, I also have a wip pr with more generic metrics exposer which will expose conntrack stats if they appear at /proc/net/stat: https://github.com/prometheus/node_exporter/pull/1771

discordianfish:
What should we do with this now that https://github.com/prometheus/node_exporter/pull/1155 has been merged?

AlexZzz:
I suggest, https://github.com/prometheus/node_exporter/pull/1155 might be reverted in favor of this one 😂

IMHO, it's better to have generic parser for any metric and any file, instead of hard-code here. Kernel API shouldn't be changed and files format should always be the same here. Any new metric will be exposed with no human interaction with generic parser.

discordianfish:
@AlexZzz I agree, lets rebase this and merge it, then create an issue to track using this to replace the conntrack stats in node-exporter.

AlexZzz:
@discordianfish , rebased.

discordianfish:
cc/ @SuperQ

SuperQ (approved):
LGTM

PR #320, opened 2020-08-06
Handle multi-digit powercap indexes

SuperQ:
Correctly handle powercap name strings by splitting on "-" rather than
assuming the index is a single digit number.

https://github.com/prometheus/node_exporter/issues/1808


SuperQ:
I need to get some sample fixtures to test this out.

discordianfish (approved):

PR #322, opened 2020-08-25
Add GIDs to ProcStatus struct

wolandr:
@discordianfish @pgier Just added a trivial parse of process user GIDs from status like UIDs and same tests. Please check it out.


discordianfish (approved):
Makes sense, thanks!

SuperQ (approved):

PR #328, opened 2020-09-16
Add slabinfo support

hodgesds:
This patch adds support for reading slab info from /proc/slabinfo. This is to potentially add support in `node_exporter` (see [#1799](https://github.com/prometheus/node_exporter/issues/1799)). One thing to note is that reading from `/proc/slabinfo` requires root/`CAP_SYS_ADMIN` to read, see the [slabinfo](https://man7.org/linux/man-pages/man5/slabinfo.5.html) man page and kernel [slub](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/tree/Documentation/vm/slub.rst) documentation.


SuperQ (approved):
We're not likely to include `CAP_SYS_ADMIN` in the node_exporter, but this is fine for this library use.

hodgesds:
Sounds reasonable

discordianfish (approved):
LGTM

PR #329, opened 2020-09-16
MDStat parsing enrichment

trearcul:
This PR enriches data parsing of MD devices from `/proc/mdstat`. All the additional data could be useful for enhancing prometheus exporters (e.g., NodeExporter) as a next step.

Added parsing of:
- RAID level
- Assigned devices
- Synced/checked percent and remaining minutes


dswarbrick:
I strongly suggest that future enhancements to mdraid stats use the sysfs entries, rather than add any more parsing of /proc/mdstat. The sysfs interface is designed to be machine-readable, whereas /proc/mdstat is designed to be human-readable. This results in a lot of regexes, which may not stand the test of time if the format of /proc/mdstat is changed or enhanced (unlikely).

See also prometheus/node_exporter#1085

SuperQ:
This looks like it needs a rebase.

PR #331, opened 2020-09-20
Use pointers in Meminfo struct to handle missing values in /proc/meminfo

### LINKED ISSUE #317: Use pointers in Meminfo struct
In order to handle the variability of /proc/meminfo between kernel versions, the Meminfo handler struct should use pointers to uint64. For example, MemAvailable doesn't exist on older kernels.

binjip978:

resolves #317


SuperQ (approved):
LGTM, Thanks!

PR #332, opened 2020-09-20
support to read from /proc/net/tcp{,6}

hubt:
This provides support to read from /proc/net/tcp and /proc/net/tcp6. Which had originally been done with https://github.com/prometheus/procfs/pull/20, but that was considered out of date, so @SuperQ suggested I open a new PR.

This change is not as extensive as it looks.

Because /proc/net/tcp{,6} is identical to /proc/net/udp{,6} and udp support already exists in net_udp.go, I went and refactored the net_udp.go to create a new set of base classes in net_ip_socket.go that could be used for both net_udp.go and net_tcp.go.

I moved most of the test code from net_udp_test.go and into net_ip_socket_test.go. I replicated the existing net_udp_test.go tests into net_tcp_test.go, and added a new tcp fixture identical to the udp one. I validated that the old UDP tests still work and new identical TCP tests also work.

I'm happy to rewrite this in a different way if people don't think this is a good idea. I had considered doing a copy of the existing net_udp.go code and just changing names to support TCP, but thought this was a bit more reusable if formats change or bugs are uncovered they would likely apply to both tcp and udp.


net_ip_socket.go
```diff
@@ -0,0 +1,205 @@
...
+	// as each line represents a single used socket.
+	// In theory, the number of available sockets is 65535 (2^16 - 1) per IP.
+	// With e.g. 150 Byte per line and the maximum number of 65535,
+	// the reader needs to handle 150 Byte * 65535 =~ 10 MB for a single IP.
+	readLimit = 4294967296 // Byte -> 4 GiB
+)
+
+// this contains generic data structures for both udp and tcp sockets
+type (
+	// NetIPSocket represents the contents of /proc/net/udp{,6} file without the header.
```
discordianfish, line 39: {udp,tcp} I assume?

net_ip_socket.go
```diff
@@ -0,0 +1,205 @@
...
+		TxQueueLength uint64
+		// RxQueueLength shows the total queue length of all parsed rx_queue lengths.
+		RxQueueLength uint64
+		// UsedSockets shows the total number of parsed lines representing the
+		// number of used sockets.
+		UsedSockets uint64
+	}
+
+	// netIPSocketLine represents the fields parsed from a single line
+	// in /proc/net/udp{,6}. Fields which are not used by IPSocket are skipped.
```
discordianfish, line 56: Same

net_ip_socket.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+func newNetIPSocket(file string) (NetIPSocket, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	netIPSocket := NetIPSocket{}
```
discordianfish, line 78: I know it's just moved here, but lets use a var block here.

net_ip_socket.go
```diff
@@ -0,0 +1,205 @@
...
+func newNetIPSocketSummary(file string) (*NetIPSocketSummary, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	netIPSocketSummary := &NetIPSocketSummary{}
+
+	lr := io.LimitReader(f, readLimit)
```
discordianfish, line 107: Same

net_ip_socket.go
hubt: yes. fixed thanks.

net_ip_socket.go
hubt: sure thing. fixed.

discordianfish (approved):

SuperQ:
Can you fix the DCO with `git commit -s --amend`?

SuperQ:
This is causing the node_exporter to crash. :-(

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8f6929]

goroutine 79 [running]:
github.com/prometheus/procfs.FS.NetUDP6Summary(0x7ffc346651b1, 0x17, 0x0, 0xc00040dca0, 0x2)
	/go/pkg/mod/github.com/prometheus/procfs@v0.4.0/net_udp.go:50 +0x169
github.com/prometheus/node_exporter/collector.(*udpQueuesCollector).Update(0xc00036c330, 0xc0003854a0, 0x110ea20, 0x0)
	/home/circleci/project/collector/udp_queues_linux.go:72 +0x26a
github.com/prometheus/node_exporter/collector.execute(0xc2f01f, 0xa, 0xd09900, 0xc00036c330, 0xc0003854a0, 0xd08fa0, 0xc0001d90e0)
	/home/circleci/project/collector/collector.go:153 +0x84
github.com/prometheus/node_exporter/collector.NodeCollector.Collect.func1(0xc0003854a0, 0xc0001d92f0, 0xd08fa0, 0xc0001d90e0, 0xc0001162a0, 0xc2f01f, 0xa, 0xd09900, 0xc00036c330)
	/home/circleci/project/collector/collector.go:144 +0x71
created by github.com/prometheus/node_exporter/collector.NodeCollector.Collect
	/home/circleci/project/collector/collector.go:143 +0x13c
```

hubt:
Sorry. Will take a look today.

hubt:
@SuperQ I built and ran the node_exporter master on amd64 linux with 3.x and 4.x kernels and didn't see any issues.

I noticed that node_exporter end-to-end-test.sh was skipping the test because there was no udp6 fixtures. When I add the fixtures from procfs it also didn't crash.

I did a circleci build and it passed tests and built all the cross compiled binaries fine.
  
What environment did you see the crash in? How can I replicate it?

SuperQ:
We should not return a nil pointer if the fixtures don't exist. That edge case needs to be handled.

Issue #218, opened 2019-09-24
Add parsing of /proc/flashcache


SuperQ:
In addition to bcache, there is [flashcache](https://github.com/facebookarchive/flashcache/blob/master/doc/flashcache-sa-guide.txt). It would be useful to have parsing of this for the node_exporter.

bisakhmondal:
Hii @SuperQ, I'd love to work on this issue. Could you please guide me? Thanks.

SuperQ:
Sure, do you have a device that is using flashcache? For example, QNAP NAS devices?

If you don't, I can post some example data here. If you look at `meminfo.go`, you can see a general style for parsing the kind of files provided.

bisakhmondal:
Thanks @SuperQ for the reply. Sadly I don't have any such devices. Please do share some data.
Now I have got a fairly good idea about flash cache. So ultimately we have to just parse it, right?

SuperQ:
For procfs, it's a reasonably simple "parse the data and map it one-to-one with a struct".

I've added a PR with an example set of fixtures for you to test against.

bisakhmondal:
Thanks for the fixtures. I'll open a PR when it's done.

SuperQ:
Going to close this, the facebook patches are abandoned and never went into mainline. There's really no point in supporting this.

PR #335, opened 2020-10-03
Add fixtures for flashcache

SuperQ:
Add an example set of fixtures for /proc/flashcache handling.

https://github.com/prometheus/procfs/issues/218

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -1896,6 +1896,82 @@ Lines: 52
    8       2 sdc2 13126 71 561749 16802 2830 1589 176404 40620 0 10931 50449 0 0 0 0 0 0
 Mode: 664
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/proc/flashcache
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/proc/flashcache/CG0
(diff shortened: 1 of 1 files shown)


fixtures.ttar
```diff
@@ -1896,6 +1896,82 @@ Lines: 52
...
+verify_errors4=0
+verify_errors5=0
+verify_errors6=0
+verify_errors7=0
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/proc/flashcache/CG0/flashcache_iosize_hist
+Lines: 1
+1:15 2:0 4:52419764 8:4301895 16:601787 32:77257 64:35248 128:29153 256:10846 512:260737 1024:128460 
+Mode: 644
```
bisakhmondal, line 1922: @SuperQ Could you please tell me what is the `iosize_hist` about and what is the a and b in this `a:b` format?

SuperQ: I have no idea, this is just what I found on the system. It doesn't appear to be documented. :-(

Looking over the source, it appears to be a [in size of sectors](https://github.com/facebookarchive/flashcache/blob/437afbfe233e94589948b76743c6489080cdd100/src/flashcache_main.c#L2087-L2093), or 512 bytes.

So we'll have to convert it from non-cumulative sectors, to cumulative bytes.

SuperQ:
Not really planning to work on this.

PR #339, opened 2020-10-30
Fix for mountstats error due to multiline "impl_id" on FreeNAS

JakubWes:
This PR fixes the issue described in [prometheus/node_exporter#1583](https://github.com/prometheus/node_exporter/issues/1583)

When reading from /proc/self/mountstats for FreeNAS NFS mount, the data in "impl_id" tends to be printed in 3 lines. As seen in the above issue, the second and third line are one field long. This results in parser exiting with "not enough information for NFS stats" error.
The PR moves the length check so that node-exporter won't crash when trying to export NFS client metrics connected to the FreeNAS server.

@discordianfish @pgier

### DIFF
--- a/mountstats.go
+++ b/mountstats.go
@@ -338,12 +338,12 @@ func parseMountStatsNFS(s *bufio.Scanner, statVersion string) (*MountStatsNFS, e
 		if len(ss) == 0 {
 			break
 		}
-		if len(ss) < 2 {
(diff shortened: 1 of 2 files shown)


discordianfish:
Can you create a new test case with the (obfuscated) content of /proc/self/mountstats you see?

JakubWes:
Sure.

discordianfish (approved):
Thanks!

SuperQ (approved):
LGTM

Issue #338, opened 2020-10-16
Large Process Limits Fail To Parse


brandonbell:
Large process limits fail to parse due to the Regex being used as well as being type `int64`.  

<details>
<summary>NTP Limits</summary>

```
Limit                     Soft Limit           Hard Limit           Units     
Max cpu time              unlimited            unlimited            seconds   
Max file size             unlimited            unlimited            bytes     
Max data size             unlimited            unlimited            bytes     
Max stack size            204800               unlimited            bytes     
Max core file size        0                    unlimited            bytes     
Max resident set          unlimited            unlimited            bytes     
Max processes             30660                30660                processes 
Max open files            1024                 4096                 files     
Max locked memory         18446744073708503040 18446744073708503040 bytes     
Max address space         unlimited            unlimited            bytes     
Max file locks            unlimited            unlimited            locks     
Max pending signals       30660                30660                signals   
Max msgqueue size         819200               819200               bytes     
Max nice priority         0                    0                    
Max realtime priority     0                    0                    
Max realtime timeout      unlimited            unlimited            us 
```
</details>

The current regex fails to split the fields correctly due to a single space between the Soft Limit, Hard Limit, and Units fields: 
https://github.com/prometheus/procfs/blob/aa47f9d89cea5aaf43204594c92a9a8460544388/proc_limits.go#L77 

Adjusting the Regex to `(Max \w+\s{0,1}?\w*\s{0,1}\w*)\s{2,}(\w+)\s+(\w+)` parses correctly (skipping the header) however it leaves the problem of the limits being `int64`.  Currently `unlimited` is being interpreted as `-1` so `uint64` won't currently work.  

An option would be to interpret `unlimited` as `18446744073709551615` rather then `-1`.  I can create a PR if this is agreed upon otherwise I'm open to suggestion!

TheMeier:
Can this be closed as fixed?

PR #340, opened 2020-10-30
Fix proc limits

brandonbell:
Fixes regex parsing of large limits as well as changes from `int64` to `uint64`.  Also adjust `unlimited` to be 18446744073709551615.  

This was realized while trying to parse the limits of `ntp` process on a Debian based system.  

https://github.com/prometheus/procfs/issues/338


SuperQ (approved):
LGTM, Thanks! I found the same problem and didn't check to see if someone had already opened a bugfix. :facepalm:

PR #343, opened 2020-11-20
Swap read and write stats for XFS

foygl:
According to the table in https://xfs.org/index.php/Runtime_Stats, the first number of `rw` stats is write and the second is read. I noticed Prometheus seemed to be reporting the opposite of what I expected.

@discordianfish @pgier


discordianfish (approved):
Oof, good catch. Thanks!

PR #344, opened 2020-11-28
proc/sys/kernel: adds support for parsing core_pattern

### LINKED ISSUE #342: Parsing of /proc/sys/kernel/core_pattern
I can raise a PR if you think this can be added.

danishprakash:
Adds support for parsing `/proc/sys/kernel/core_pattern`.

Closes #342

### DIFF
--- a/kernel_corepattern.go
+++ b/kernel_corepattern.go
@@ -1,0 +1,33 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package procfs
+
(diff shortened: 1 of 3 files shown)


kernel_corepattern_test.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2019 The Prometheus Authors
```
SuperQ, line 1: New files should get the current year.

```suggestion
// Copyright 2020 The Prometheus Authors
```

kernel_corepattern.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"bytes"
+	"os"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+// KernelCorePattern returns value from /proc/sys/kernel/core_pattern
+func (fs FS) KernelCorePattern() (pattern []byte, err error) {
```
SuperQ, line 26: Don't name these returns.

```suggestion
func (fs FS) KernelCorePattern() ([]byte, error) {
```

SuperQ (commented):
I'm not sure how useful this function is. It seems like too trivial a wrapper to be all that useful.

danishprakash:
@SuperQ It does seem trivial tbh. On the other hand, when I was looking for such a functionality recently, I was certain It would be something I'd find in this project. Writing a wrapper is fine with me and in fact, I was a bit skeptic about adding this myself and opened an issue(#342) beforehand to clear this up. cc: @discordianfish

SuperQ:
Without some more useful parsing, I don't think we need this.

PR #347, opened 2021-01-04
Parse /proc/net/protocols

juancb:
This patch adds support for extracting system socket metrics found only in /proc/net/protocols.  The most interesting metrics found here are per-protocol socket counts along with the total memory in pages used by each. For TCP connections this data can be used to tune/monitor a system against the thresholds set in `net.ipv4.tcp_mem`.  This file also exposes a pressure metric which indicates whether or not a protocol like TCP has entered memory pressure, which again is useful in troubleshooting things like memory pressure induced packet drops on loopback.

Some additional albeit less interesting metrics that come out of this file include the size of each protocol's struct, and the maximum header size the kernel is willing to allocate.

The additional columns enumerate capabilities per-protocol and are encoded as bools.


discordianfish (approved):
LGTM

PR #351, opened 2021-01-16
Fix to add the missing parsing of the ARP mac address

Kenterfie:
- to parse a string representation of a mac address it is required to use the ParseMAC function
- added additional test case with empty mac address (incomplete)


SuperQ:
This appears abandoned.

PR #354, opened 2021-01-26
Allow cgroup paths containing colons

### LINKED ISSUE #353: Cgroup paths containing ":" are not supported
Considering https://github.com/prometheus/procfs/blob/bc932d6aaa40fc66d46db859e9e03d954dc622c1/proc_cgroup.go#L52 it is clear that if the path contains additional ":" characters, the path will be truncated.

Colons may be inserted, for example when running kubernetes with containerd using a `systemd` cgroup-driver.

See:
* https://github.com/containerd/containerd/blob/c35b4cfed520bf4827737f294f9292ffa73bdf8b/pkg/cri/server/helpers_linux.go#L67

And how containerd parses the file: https://github.com/containerd/cgroups/blob/8a68de567b68b30602948417a7a8e99ccf00ee9b/utils.go#L305

Changing:
```
	fields := strings.Split(cgroupStr, ":")
```
to
```
	fields := strings.SplitN(cgroupStr, ":", 3)
```
should address this.

elezar:
This change adds support for cgroup paths containing colons in the /procs/<pid>/cgroup file.

This is prompted by the following:
* The spec specifies that the *three* fields are separated by colons and does not forbid them in the path
* Tools such as containerd joins path components with colons when using systemd as a cgroup provider

This closes #353

### DIFF
--- a/proc_cgroup.go
+++ b/proc_cgroup.go
@@ -49,7 +49,7 @@ type Cgroup struct {
 func parseCgroupString(cgroupStr string) (*Cgroup, error) {
 	var err error
 
-	fields := strings.Split(cgroupStr, ":")
+	fields := strings.SplitN(cgroupStr, ":", 3)
 	if len(fields) < 3 {
 		return nil, fmt.Errorf("at least 3 fields required, found %d fields in cgroup string: %s", len(fields), cgroupStr)
 	}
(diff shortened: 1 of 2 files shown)


discordianfish (approved):
Great, thanks!

SuperQ (approved):
LGTM

PR #355, opened 2021-02-07
Fix null pointer return in net_udp

hubt:
Signed-off-by: Hubert Chen <hubert@branchmetrics.io>

### DIFF
--- a/net_udp.go
+++ b/net_udp.go
@@ -60,5 +60,5 @@ func newNetUDPSummary(file string) (*NetUDPSummary, error) {
 		return nil, err
 	}
 	n1 := NetUDPSummary(*n)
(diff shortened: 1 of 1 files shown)


net_udp.go
```diff
@@ -64,5 +60,5 @@ func newNetUDPSummary(file string) (*NetUDPSummary, error) {
 		return nil, err
 	}
 	n1 := NetUDPSummary(*n)
-	return &n1, err
+	return &n1, nil
```
SuperQ, line 63: I think this should probably stay `return &n1, err`, in case `newNetIPSocketSummary()` returns a non-nil result and an error.

SuperQ (approved):
LGTM

Issue #310, opened 2020-06-14
Use Go 1.13 error handling verb


SuperQ:
When using `fmt.Errorf()`, we should use the new [`%w` error verb](https://blog.golang.org/go1.13-errors) so that error messages can be unwrapped.

This would make the minimum Go version be 1.13, so we should make sure client_golang is OK with this.

SuperQ:
cc @beorn7 about the client_golang interaction.

beorn7:
There is no fixed policy, but unless it's prohibitively cumbersome, I'd like to support three to four past Go versions, which is currently back to Go 1.12 or 1.11.

Could we use a build tag to keep supporting older build versions for a while?

SuperQ:
We could, but there's no rush to do this change. It's just a nice to have to allow more verbose library errors, while still being able to parse "no such file or directory" errors.

conallob:
Is this not a dupe of https://github.com/prometheus/procfs/issues/358 ? If so, it has been partially implemented, with https://github.com/prometheus/procfs/issues/519 waiting for a major release in order to change the minimum Go version.

SuperQ:
Yea, this can be closed.

PR #357, opened 2021-02-09
Update error message handling

SuperQ:
Use Go 1.13 `%w` error formatting
* Update minimum Go version to 1.13.
* Bump vendoring.
* Update all error format strings to use %w.
* Cleanup consistency of use.
* Cleanup use of `errors.New("string" + var)` to use `fmt.Errorf()`.

https://github.com/prometheus/procfs/issues/310

### DIFF
--- a/cpuinfo.go
+++ b/cpuinfo.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
@@ -77,7 +78,7 @@ func parseCPUInfoX86(info []byte) ([]CPUInfo, error) {
 	// find the first "processor" line
 	firstLine := firstNonEmptyLine(scanner)
 	if !strings.HasPrefix(firstLine, "processor") || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
 	v, err := strconv.ParseUint(field[1], 0, 32)
@@ -192,7 +193,7 @@ func parseCPUInfoARM(info []byte) ([]CPUInfo, error) {
 	firstLine := firstNonEmptyLine(scanner)
 	match, _ := regexp.MatchString("^[Pp]rocessor", firstLine)
 	if !match || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
 	cpuinfo := []CPUInfo{}
@@ -256,7 +257,7 @@ func parseCPUInfoS390X(info []byte) ([]CPUInfo, error) {
 
 	firstLine := firstNonEmptyLine(scanner)
 	if !strings.HasPrefix(firstLine, "vendor_id") || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
 	cpuinfo := []CPUInfo{}
@@ -281,7 +282,7 @@ func parseCPUInfoS390X(info []byte) ([]CPUInfo, error) {
 		if strings.HasPrefix(line, "processor") {
 			match := cpuinfoS390XProcessorRegexp.FindStringSubmatch(line)
 			if len(match) < 2 {
-				return nil, errors.New("Invalid line found in cpuinfo: " + line)
+				return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 			}
 			cpu := commonCPUInfo
 			v, err := strconv.ParseUint(match[1], 0, 32)
@@ -341,7 +342,7 @@ func parseCPUInfoMips(info []byte) ([]CPUInfo, error) {
 	// find the first "processor" line
 	firstLine := firstNonEmptyLine(scanner)
 	if !strings.HasPrefix(firstLine, "system type") || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
 	cpuinfo := []CPUInfo{}
@@ -383,7 +384,7 @@ func parseCPUInfoPPC(info []byte) ([]CPUInfo, error) {
 
 	firstLine := firstNonEmptyLine(scanner)
 	if !strings.HasPrefix(firstLine, "processor") || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
 	v, err := strconv.ParseUint(field[1], 0, 32)
@@ -428,7 +429,7 @@ func parseCPUInfoRISCV(info []byte) ([]CPUInfo, error) {
 
 	firstLine := firstNonEmptyLine(scanner)
 	if !strings.HasPrefix(firstLine, "processor") || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
 	}
 	field := strings.SplitN(firstLine, ": ", 2)
 	v, err := strconv.ParseUint(field[1], 0, 32)
--- a/iscsi/get.go
+++ b/iscsi/get.go
@@ -36,7 +36,7 @@ func GetStats(iqnPath string) (*Stats, error) {
 
 	matches, err := filepath.Glob(filepath.Join(iqnPath, "tpgt*"))
 	if err != nil {
-		return nil, fmt.Errorf("iscsi: GetStats: get TPGT path error %v", err)
+		return nil, fmt.Errorf("iscsi: GetStats: get TPGT path error %w", err)
 	}
 	istats.Tpgt = make([]TPGT, len(matches))
 
@@ -63,11 +63,11 @@ func GetStats(iqnPath string) (*Stats, error) {
 func isPathEnable(path string) (bool, error) {
 	enableReadout, err := ioutil.ReadFile(filepath.Join(path, "enable"))
 	if err != nil {
-		return false, fmt.Errorf("iscsi: isPathEnable ReadFile error %v", err)
+		return false, fmt.Errorf("iscsi: isPathEnable ReadFile error %w", err)
 	}
 	isEnable, err := strconv.ParseBool(strings.TrimSpace(string(enableReadout)))
 	if err != nil {
-		return false, fmt.Errorf("iscsi: isPathEnable ParseBool error %v", err)
+		return false, fmt.Errorf("iscsi: isPathEnable ParseBool error %w", err)
 	}
 	return isEnable, nil
 }
@@ -77,14 +77,14 @@ func getLunLinkTarget(lunPath string) (lunObject LUN, err error) {
 	lunObject.LunPath = lunPath
 	files, err := ioutil.ReadDir(lunPath)
 	if err != nil {
-		return lunObject, fmt.Errorf("getLunLinkTarget: ReadDir path %s error %v", lunPath, err)
+		return lunObject, fmt.Errorf("getLunLinkTarget: ReadDir path %q: %w", lunPath, err)
 	}
 	for _, file := range files {
 		fileInfo, _ := os.Lstat(filepath.Join(lunPath, file.Name()))
 		if fileInfo.Mode()&os.ModeSymlink != 0 {
 			target, err := os.Readlink(filepath.Join(lunPath, fileInfo.Name()))
 			if err != nil {
-				return lunObject, fmt.Errorf("getLunLinkTarget: Readlink err %v", err)
+				return lunObject, fmt.Errorf("getLunLinkTarget: Readlink: %w", err)
 			}
 			targetPath, objectName := filepath.Split(target)
 			_, typeWithNumber := filepath.Split(filepath.Clean(targetPath))
@@ -112,24 +112,21 @@ func ReadWriteOPS(iqnPath string, tpgt string, lun string) (readmb uint64,
 		"statistics/scsi_tgt_port/read_mbytes")
 	readmb, err = util.ReadUintFromFile(readmbPath)
 	if err != nil {
-		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: read_mbytes error file %s and %v",
-			readmbPath, err)
+		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: read_mbytes error file %q: %w", readmbPath, err)
 	}
 
 	writembPath := filepath.Join(iqnPath, tpgt, "lun", lun,
 		"statistics/scsi_tgt_port/write_mbytes")
 	writemb, err = util.ReadUintFromFile(writembPath)
 	if err != nil {
-		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: write_mbytes error file %s and %v",
-			writembPath, err)
+		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: write_mbytes error file %q: %w", writembPath, err)
 	}
 
 	iopsPath := filepath.Join(iqnPath, tpgt, "lun", lun,
 		"statistics/scsi_tgt_port/in_cmds")
 	iops, err = util.ReadUintFromFile(iopsPath)
 	if err != nil {
-		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: in_cmds error file %s and %v",
-			iopsPath, err)
+		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: in_cmds error file %q: %w", iopsPath, err)
 	}
 
 	return readmb, writemb, iops, nil
@@ -150,7 +147,7 @@ func (fs FS) GetFileioUdev(fileioNumber string, objectName string) (*FILEIO, err
 	}
 	filename, err := ioutil.ReadFile(udevPath)
 	if err != nil {
-		return nil, fmt.Errorf("iscsi: GetFileioUdev: Cannot read filename from udev link :%s", udevPath)
+		return nil, fmt.Errorf("iscsi: GetFileioUdev: Cannot read filename from udev link %q", udevPath)
 	}
 	fileio.Filename = strings.TrimSpace(string(filename))
 
@@ -172,7 +169,7 @@ func (fs FS) GetIblockUdev(iblockNumber string, objectName string) (*IBLOCK, err
 	}
 	filename, err := ioutil.ReadFile(udevPath)
 	if err != nil {
-		return nil, fmt.Errorf("iscsi: GetIBlockUdev: Cannot read iblock from udev link :%s", udevPath)
+		return nil, fmt.Errorf("iscsi: GetIBlockUdev: Cannot read iblock from udev link %q", udevPath)
 	}
 	iblock.Iblock = strings.TrimSpace(string(filename))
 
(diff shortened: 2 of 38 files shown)


roidelapluie (approved):

beorn7:
I guess with go1.16 on the horizon, we can do this now.

cpuinfo.go
```diff
@@ -77,7 +77,7 @@ func parseCPUInfoX86(info []byte) ([]CPUInfo, error) {
 	// find the first "processor" line
 	firstLine := firstNonEmptyLine(scanner)
 	if !strings.HasPrefix(firstLine, "processor") || !strings.Contains(firstLine, ":") {
-		return nil, errors.New("invalid cpuinfo file: " + firstLine)
+		return nil, fmt.Errorf("invalid cpuinfo file: %q", firstLine)
```
beorn7, line 80: Do you think we need quoting after a colon? (We didn't quote before, so I'm wondering if this change is deliberate.)

loadavg.go
```diff
@@ -44,14 +44,14 @@ func parseLoad(loadavgBytes []byte) (*LoadAvg, error) {
 	loads := make([]float64, 3)
 	parts := strings.Fields(string(loadavgBytes))
 	if len(parts) < 3 {
-		return nil, fmt.Errorf("malformed loadavg line: too few fields in loadavg string: %s", string(loadavgBytes))
+		return nil, fmt.Errorf("malformed loadavg line: too few fields in loadavg string: %q", string(loadavgBytes))
```
beorn7, line 47: Same note as above about %q vs. %s.

iscsi/get.go
```diff
@@ -112,24 +112,21 @@ func ReadWriteOPS(iqnPath string, tpgt string, lun string) (readmb uint64,
...
+		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: write_mbytes error file %s: %w", writembPath, err)
 	}
 
 	iopsPath := filepath.Join(iqnPath, tpgt, "lun", lun,
 		"statistics/scsi_tgt_port/in_cmds")
 	iops, err = util.ReadUintFromFile(iopsPath)
 	if err != nil {
-		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: in_cmds error file %s and %v",
-			iopsPath, err)
+		return 0, 0, 0, fmt.Errorf("iscsi: ReadWriteOPS: in_cmds error file %s: %w", iopsPath, err)
```
beorn7, line 129: Here (and similarly above), I think, a `%q` would actually make more sense than a `%s` because it is in the middle of the message, not after the final `:`. WDYT?

mdstat.go
```diff
@@ -59,7 +59,7 @@ func (fs FS) MDStat() ([]MDStat, error) {
 	}
 	mdstat, err := parseMDStat(data)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing mdstat %s: %s", fs.proc.Path("mdstat"), err)
+		return nil, fmt.Errorf("error parsing mdstat %s: %w", fs.proc.Path("mdstat"), err)
```
beorn7, line 62: Again, here I think `%q` is better than `%s`.

beorn7 (commented):
I sense a somewhat inconsistent use of `%s` vs `%q`. Before adding a million comments, perhaps you should first tell us what the intention is.

cpuinfo.go
SuperQ: It is deliberate. I was thinking we should be better about passing random kernel strings to error messages with quoting.

The logging in this library is very inconsistent, so I was only trying to do some minimal cleanup.

iscsi/get.go
SuperQ: Yes, that is one we should cleanup. Will add it.

SuperQ:
Updated with some `s/%s/%q`.

beorn7:
So the guideline is to quote if in doubt? I'll make a pass following that guideline.

iscsi/get.go
```diff
@@ -150,7 +147,7 @@ func (fs FS) GetFileioUdev(fileioNumber string, objectName string) (*FILEIO, err
 	}
 	filename, err := ioutil.ReadFile(udevPath)
 	if err != nil {
-		return nil, fmt.Errorf("iscsi: GetFileioUdev: Cannot read filename from udev link :%s", udevPath)
+		return nil, fmt.Errorf("iscsi: GetFileioUdev: Cannot read filename from udev link: %s", udevPath)
```
beorn7, line 150: %s → %q

net_ip_socket.go
```diff
@@ -129,8 +129,7 @@ func parseIP(hexIP string) (net.IP, error) {
 	var byteIP []byte
 	byteIP, err := hex.DecodeString(hexIP)
 	if err != nil {
-		return nil, fmt.Errorf(
-			"cannot parse address field in socket line: %s", hexIP)
+		return nil, fmt.Errorf("cannot parse address field in socket line: %s", hexIP)
```
beorn7, line 132: This is presumably a hexadecimal number, but I guess it could be anything if the decoding has gone wrong. So let's use `%q` here, too?

net_ip_socket.go
```diff
@@ -162,45 +161,39 @@ func parseNetIPSocketLine(fields []string) (*netIPSocketLine, error) {
 	// sl
 	s := strings.Split(fields[0], ":")
 	if len(s) != 2 {
-		return nil, fmt.Errorf(
-			"cannot parse sl field in socket line: %s", fields[0])
+		return nil, fmt.Errorf("cannot parse sl field in socket line: %s", fields[0])
```
beorn7, line 164: %s → %q

net_ip_socket.go
```diff
@@ -162,45 +161,39 @@ func parseNetIPSocketLine(fields []string) (*netIPSocketLine, error) {
...
 	if line.Sl, err = strconv.ParseUint(s[0], 0, 64); err != nil {
-		return nil, fmt.Errorf("cannot parse sl value in socket line: %s", err)
+		return nil, fmt.Errorf("cannot parse sl value in socket line: %w", err)
 	}
 	// local_address
 	l := strings.Split(fields[1], ":")
 	if len(l) != 2 {
-		return nil, fmt.Errorf(
-			"cannot parse local_address field in socket line: %s", fields[1])
+		return nil, fmt.Errorf("cannot parse local_address field in socket line: %s", fields[1])
```
beorn7, line 173: %s → %q

net_ip_socket.go
```diff
@@ -162,45 +161,39 @@ func parseNetIPSocketLine(fields []string) (*netIPSocketLine, error) {
...
-			"cannot parse local_address port value in socket line: %s", err)
+		return nil, fmt.Errorf("cannot parse local_address port value in socket line: %w", err)
 	}
 
 	// remote_address
 	r := strings.Split(fields[2], ":")
 	if len(r) != 2 {
-		return nil, fmt.Errorf(
-			"cannot parse rem_address field in socket line: %s", fields[1])
+		return nil, fmt.Errorf("cannot parse rem_address field in socket line: %s", fields[1])
```
beorn7, line 185: %s → %q

proc_limits.go
```diff
@@ -103,8 +103,7 @@ func (p Proc) Limits() (ProcLimits, error) {
 		//fields := limitsMatch.Split(s.Text(), limitsFields)
 		fields := limitsMatch.FindStringSubmatch(s.Text())
 		if len(fields) != limitsFields {
-			return ProcLimits{}, fmt.Errorf(
-				"couldn't parse %s line %s", f.Name(), s.Text())
+			return ProcLimits{}, fmt.Errorf("couldn't parse %s line %s", f.Name(), s.Text())
```
beorn7, line 106: Perhaps `"couldn't parse %s line: %q"` would be more consistent? Depending on the nature of `f.Name()`, it might deserve a %q, too.

proc_ns.go
```diff
@@ -58,7 +58,7 @@ func (p Proc) Namespaces() (Namespaces, error) {
 		typ := fields[0]
 		inode, err := strconv.ParseUint(strings.Trim(fields[1], "[]"), 10, 32)
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse inode from '%v': %v", fields[1], err)
+			return nil, fmt.Errorf("failed to parse inode from '%v': %w", fields[1], err)
```
beorn7, line 61: `'%v'` → `%q`

proc_stat.go
```diff
@@ -127,10 +127,7 @@ func (p Proc) Stat() (ProcStat, error) {
 	)
 
 	if l < 0 || r < 0 {
-		return ProcStat{}, fmt.Errorf(
-			"unexpected format, couldn't extract comm: %s",
-			data,
-		)
+		return ProcStat{}, fmt.Errorf("unexpected format, couldn't extract comm: %s", data)
```
beorn7, line 130: %s → %q

schedstat.go
```diff
@@ -95,24 +95,27 @@ func (fs FS) Schedstat() (*Schedstat, error) {
...
 		}
 
 		stats.RunTimeslices, err = strconv.ParseUint(match[3], 10, 64)
-		return
+		return stats, err
 	}
 
-	err = errors.New("could not parse schedstat")
-	return
+	return stats, errors.New("could not parse schedstat")
```
beorn7, line 120: You have replaced similar instances with `fmt.Errorf` above. Personally, I'd still use `errors.New(...)` for errors that don't actually need formatting, but in any case, I'd be consistent.

cpuinfo.go
```diff
@@ -464,7 +464,7 @@ func parseCPUInfoRISCV(info []byte) ([]CPUInfo, error) {
 }
 
 func parseCPUInfoDummy(_ []byte) ([]CPUInfo, error) { // nolint:unused,deadcode
-	return nil, errors.New("not implemented")
+	return nil, fmt.Errorf("not implemented")
```
beorn7, line 467: This could just stay `errors.New`, but if you want to replace this case, too, I'd do it consistently (see comment below).

stat.go
```diff
@@ -93,7 +93,7 @@ func parseCPUStat(line string) (CPUStat, int64, error) {
 		&cpuStat.Guest, &cpuStat.GuestNice)
 
 	if err != nil && err != io.EOF {
-		return CPUStat{}, -1, fmt.Errorf("couldn't parse %s (cpu): %s", line, err)
+		return CPUStat{}, -1, fmt.Errorf("couldn't parse %s (cpu): %w", line, err)
```
beorn7, line 96: %s → %q

stat.go
```diff
@@ -116,7 +116,7 @@ func parseCPUStat(line string) (CPUStat, int64, error) {
 
 	cpuID, err := strconv.ParseInt(cpu[3:], 10, 64)
 	if err != nil {
-		return CPUStat{}, -1, fmt.Errorf("couldn't parse %s (cpu/cpuid): %s", line, err)
+		return CPUStat{}, -1, fmt.Errorf("couldn't parse %s (cpu/cpuid): %w", line, err)
```
beorn7, line 119: %s → %q

stat.go
```diff
@@ -136,7 +136,7 @@ func parseSoftIRQStat(line string) (SoftIRQStat, uint64, error) {
 		&softIRQStat.Hrtimer, &softIRQStat.Rcu)
 
 	if err != nil {
-		return SoftIRQStat{}, 0, fmt.Errorf("couldn't parse %s (softirq): %s", line, err)
+		return SoftIRQStat{}, 0, fmt.Errorf("couldn't parse %s (softirq): %w", line, err)
```
beorn7, line 139: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
 		switch {
 		case parts[0] == "btime":
 			if stat.BootTime, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
-				return Stat{}, fmt.Errorf("couldn't parse %s (btime): %s", parts[1], err)
+				return Stat{}, fmt.Errorf("couldn't parse %s (btime): %w", parts[1], err)
```
beorn7, line 187: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
...
 			}
 		case parts[0] == "intr":
 			if stat.IRQTotal, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
-				return Stat{}, fmt.Errorf("couldn't parse %s (intr): %s", parts[1], err)
+				return Stat{}, fmt.Errorf("couldn't parse %s (intr): %w", parts[1], err)
```
beorn7, line 191: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
...
 			}
 			numberedIRQs := parts[2:]
 			stat.IRQ = make([]uint64, len(numberedIRQs))
 			for i, count := range numberedIRQs {
 				if stat.IRQ[i], err = strconv.ParseUint(count, 10, 64); err != nil {
-					return Stat{}, fmt.Errorf("couldn't parse %s (intr%d): %s", count, i, err)
+					return Stat{}, fmt.Errorf("couldn't parse %s (intr%d): %w", count, i, err)
```
beorn7, line 197: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
...
 				}
 			}
 		case parts[0] == "ctxt":
 			if stat.ContextSwitches, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
-				return Stat{}, fmt.Errorf("couldn't parse %s (ctxt): %s", parts[1], err)
+				return Stat{}, fmt.Errorf("couldn't parse %s (ctxt): %w", parts[1], err)
```
beorn7, line 202: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
...
 			}
 		case parts[0] == "processes":
 			if stat.ProcessCreated, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
-				return Stat{}, fmt.Errorf("couldn't parse %s (processes): %s", parts[1], err)
+				return Stat{}, fmt.Errorf("couldn't parse %s (processes): %w", parts[1], err)
```
beorn7, line 206: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
...
 			}
 		case parts[0] == "procs_running":
 			if stat.ProcessesRunning, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
-				return Stat{}, fmt.Errorf("couldn't parse %s (procs_running): %s", parts[1], err)
+				return Stat{}, fmt.Errorf("couldn't parse %s (procs_running): %w", parts[1], err)
```
beorn7, line 210: %s → %q

stat.go
```diff
@@ -184,34 +184,34 @@ func (fs FS) Stat() (Stat, error) {
...
 			}
 		case parts[0] == "procs_blocked":
 			if stat.ProcessesBlocked, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
-				return Stat{}, fmt.Errorf("couldn't parse %s (procs_blocked): %s", parts[1], err)
+				return Stat{}, fmt.Errorf("couldn't parse %s (procs_blocked): %w", parts[1], err)
```
beorn7, line 214: %s → %q

xfrm.go
```diff
@@ -112,8 +112,7 @@ func (fs FS) NewXfrmStat() (XfrmStat, error) {
 		fields := strings.Fields(s.Text())
 
 		if len(fields) != 2 {
-			return XfrmStat{}, fmt.Errorf(
-				"couldn't parse %s line %s", file.Name(), s.Text())
+			return XfrmStat{}, fmt.Errorf("couldn't parse %s line %s", file.Name(), s.Text())
```
beorn7, line 115: As above, I'd suggest `"couldn't parse %s line: %q"` or even `"couldn't parse %q line: %q"`.

beorn7 (changes requested):
OK, I have found a number of more instances that I believe should be `%q`.

In summary, I wholeheartedly agree that potentially unknown or misformatted strings in the middle of an error message should be quoted. I'm way more relaxed about anything after a colon, after which you only have a blank and then said string. But I can see the argument for using %q there, too, especially if the string might contain line breaks or tabs.

beorn7:
And to be clear: I have reviewed now with the assumption that we only ever want to use `%s` on strings where we can be certain they don't contain "weird stuff".

SuperQ:
@beorn7 Yes, that seems like a reasonable policy, thanks for the great reviewing.

net_ip_socket.go
SuperQ: Yes, it's going to be a malformed string, so quoting is a good idea.

SuperQ:
@beorn7 Ok, I think I've got it cleaned up. PTAL.

iscsi/get.go
```diff
@@ -172,7 +169,7 @@ func (fs FS) GetIblockUdev(iblockNumber string, objectName string) (*IBLOCK, err
 	}
 	filename, err := ioutil.ReadFile(udevPath)
 	if err != nil {
-		return nil, fmt.Errorf("iscsi: GetIBlockUdev: Cannot read iblock from udev link :%s", udevPath)
+		return nil, fmt.Errorf("iscsi: GetIBlockUdev: Cannot read iblock from udev link %s", udevPath)
```
beorn7, line 172: %s → %q

net_conntrackstat.go
```diff
@@ -147,7 +147,7 @@ func parseConntrackStatEntry(fields []string) (*ConntrackStatEntry, error) {
 func parseConntrackStatField(field string) (uint64, error) {
 	val, err := strconv.ParseUint(field, 16, 64)
 	if err != nil {
-		return 0, fmt.Errorf("couldn't parse \"%s\" field: %s", field, err)
+		return 0, fmt.Errorf("couldn't parse \"%s\" field: %w", field, err)
```
beorn7, line 150: `\"%s\"` → `%q`

net_unix.go
```diff
@@ -136,29 +136,29 @@ func (u *NetUNIX) parseLine(line string, hasInode bool, min int) (*NetUNIXLine,
 
 	users, err := u.parseUsers(fields[1])
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse ref count(%s): %v", fields[1], err)
+		return nil, fmt.Errorf("failed to parse ref count(%s): %w", fields[1], err)
```
beorn7, line 139: I guess this should be `"failed to parse ref count %q: %w"` in line with similar messages above.

net_unix.go
```diff
@@ -136,29 +136,29 @@ func (u *NetUNIX) parseLine(line string, hasInode bool, min int) (*NetUNIXLine,
...
 	}
 
 	flags, err := u.parseFlags(fields[3])
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse flags(%s): %v", fields[3], err)
+		return nil, fmt.Errorf("failed to parse flags(%s): %w", fields[3], err)
```
beorn7, line 144: Same.

net_unix.go
```diff
@@ -136,29 +136,29 @@ func (u *NetUNIX) parseLine(line string, hasInode bool, min int) (*NetUNIXLine,
...
 	}
 
 	typ, err := u.parseType(fields[4])
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse type(%s): %v", fields[4], err)
+		return nil, fmt.Errorf("failed to parse type(%s): %w", fields[4], err)
```
beorn7, line 149: Same.

net_unix.go
```diff
@@ -136,29 +136,29 @@ func (u *NetUNIX) parseLine(line string, hasInode bool, min int) (*NetUNIXLine,
...
 	}
 
 	state, err := u.parseState(fields[5])
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse state(%s): %v", fields[5], err)
+		return nil, fmt.Errorf("failed to parse state(%s): %w", fields[5], err)
```
beorn7, line 154: Same.

net_unix.go
```diff
@@ -136,29 +136,29 @@ func (u *NetUNIX) parseLine(line string, hasInode bool, min int) (*NetUNIXLine,
...
 	}
 
 	var inode uint64
 	if hasInode {
 		inode, err = u.parseInode(fields[6])
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse inode(%s): %v", fields[6], err)
+			return nil, fmt.Errorf("failed to parse inode(%s): %w", fields[6], err)
```
beorn7, line 161: Same.

sysfs/net_class.go
```diff
@@ -69,7 +69,7 @@ func (fs FS) NetClassDevices() ([]string, error) {
 
 	devices, err := ioutil.ReadDir(path)
 	if err != nil {
-		return res, fmt.Errorf("cannot access %s dir %s", path, err)
+		return res, fmt.Errorf("cannot access %s dir %w", path, err)
```
beorn7, line 72: How about `"cannot access dir %q: %w"`?

beorn7 (changes requested):
I found a few more. :duck:

net_unix.go
SuperQ: Yea, so much inconsistency her. :frowning_face:

net_conntrackstat.go
SuperQ: lol://:facepalm:

sysfs/net_class.go
SuperQ: Yes, I like it.

beorn7 (approved):
\o/ The world is a better place now!

Issue #361, opened 2021-02-14
Infiniband parseRate fails on Melanox card


SuperQ:
Linux: 4.14.24
mlx5_ib: Mellanox Connect-IB Infiniband driver v4.3-0.1.6

The format of the `rate` file is slightly different:
```
$ cat /sys/class/infiniband/mlx5_1/ports/1/rate
0 GB/sec
```

This happens with the SFP+ slot is unpoulated.

PR #362, opened 2021-02-14
Fix infiniband rate parsing

SuperQ:
Fix parsing of infiniband rate for unconnected/unpopulated ports.

Fixes: https://github.com/prometheus/procfs/issues/361

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -175,9 +175,9 @@ func parseState(s string) (uint, string, error) {
 
 // Parse rate (example: "100 Gb/sec (4X EDR)") and return it as bytes/second
 func parseRate(s string) (uint64, error) {
-	parts := strings.Split(s, "Gb/sec")
+	parts := strings.SplitAfterN(s, " ", 2)
 	if len(parts) != 2 {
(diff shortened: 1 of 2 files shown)


roidelapluie (approved):

sysfs/class_infiniband.go
```diff
@@ -175,9 +175,9 @@ func parseState(s string) (uint, string, error) {
 
 // Parse rate (example: "100 Gb/sec (4X EDR)") and return it as bytes/second
 func parseRate(s string) (uint64, error) {
-	parts := strings.Split(s, "Gb/sec")
+	parts := strings.SplitAfterN(s, " ", 2)
```
roidelapluie, line 178: should we call to lower to make sure it is gb/s ?

SuperQ: I don't think we really care what the string is, based on the [kernel source](https://github.com/torvalds/linux/blob/master/drivers/infiniband/core/sysfs.c#L293-L295) it's always going to represent Gb/sec.

I'm not sure what's up with why my card driver, I can't find the upper-case version anywhere in the source.

discordianfish (approved):

PR #365, opened 2021-02-26
Mark sysfs as Linux-only

### LINKED ISSUE #359: FreeBSD build error: undefined: syscall.ENODATA
#300 seems to have introduced a regression on FreeBSD:

```
vendor/github.com/prometheus/procfs/sysfs/class_thermal.go:51:22: undefined: syscall.ENODATA
```

You can see from the [Go source](https://golang.org/src/syscall/zerrors_freebsd_amd64.go) that ENODATA isn't an error on FreeBSD.

Unfortunately, I don't understand this enough myself to try to contribute a fix, but this is preventing me from building node_exporter for FreeBSD.

rfratto:
The sysfs package is imported by node_exporter. `class_thermal` uses `ENODATA`, which doesn't exist on FreeBSD, so the build fails. 

Since sysfs is a Linux concept, this PR marks most of the functions within the package as Linux-only. Closes #359.

### DIFF
--- a/sysfs/class_cooling_device.go
+++ b/sysfs/class_cooling_device.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !windows
+// +build linux
 
 package sysfs
 
--- a/sysfs/class_cooling_device_test.go
+++ b/sysfs/class_cooling_device_test.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !windows
+// +build linux
 
 package sysfs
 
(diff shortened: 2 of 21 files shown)


SuperQ:
Maybe we should mark everything under `/sysfs` as Linux-only?

rfratto:
> Maybe we should mark everything under /sysfs as Linux-only?

Yep, agreed. Changed the PR to do this. Thanks!

discordianfish (approved):
LGTM

rfratto:
Do we need a second pair of eyes to get this merged?

SuperQ (approved):
LGTM

SuperQ:
Sorry, fell out of my TODO list.

PR #370, opened 2021-03-14
Parse /sys/class/drm/cardN/device/ stats

siavashs:
This PR adds support for reading card stats exposed by drm.
Only `amdgpu` driver currently exposes these stats but it should be easy to add more driver support if required.
FreeBSD uses the same driver through `kms-drm` but it does not expose the attributes we are interested in through niether `sysctl` or `linsysfs`.

### DIFF
--- a/sysfs/class_drm.go
+++ b/sysfs/class_drm.go
@@ -1,0 +1,26 @@
+// Copyright 2021 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+package sysfs
+
+import (
+	"path/filepath"
+
+	"github.com/prometheus/procfs/internal/util"
+)
+
+func readDRMCardField(card, field string) (string, error) {
+	return util.SysReadFile(filepath.Join(card, "device", field))
+}
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -3455,6 +3455,460 @@ Mode: 664
 Directory: fixtures/sys/class
 Mode: 775
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/drm
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/drm/card0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/drm/card0/device
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/drm/card0/device/aer_dev_correctable
+Lines: 9
+RxErr 0
+BadTLP 0
+BadDLLP 0
+Rollover 0
+Timeout 0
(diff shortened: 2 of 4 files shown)


sysfs/class_drm.go
```diff
@@ -0,0 +1,124 @@
...
+var (
+	// list of supported device drivers.
+	drmDeviceDrivers = []string{"amdgpu"}
+)
+
+// ClassDrmCardStats contains info from files in /sys/class/drm/card<card>/device
+// for a single card.
+// https://www.kernel.org/doc/html/latest/gpu/drivers.html
+// https://www.kernel.org/doc/html/latest/gpu/amdgpu.html
+type ClassDrmCardStats struct {
```
discordianfish, line 39: nit: Acronyms like DRM should be uppercase: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

sysfs/class_drm.go
```diff
@@ -0,0 +1,124 @@
...
+	Name                          string // The card name.
+	Driver                        string // The kernel driver loaded for this card.
+	GpuBusyPercent                uint64 `amdgpu:"gpu_busy_percent"`                  // How busy the GPU is as a percentage.
```
discordianfish, line 42: Same for the acronyms like GPU, GTT, VRAM and DPM

sysfs/class_drm.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			return stats, err
+		}
+		if match {
+			stats.Driver = driver
+			break
+		}
+	}
+
+	// Use struct tags to dynamically read and parse values depending on the driver.
```
discordianfish, line 94: So to support other drivers, we'd add a new tag to an existing field to read from a different sysfs field?
So far we tried to avoid using reflect where possible. But seems like a reasonable pattern, just that if we want to go that route, we should do it this way for similar cases as well. @pgier @SuperQ wdyt?

siavashs: Correct, I came up with this solution to avoid adding custom function per driver.
But if using `reflect` is an issue, I can drop it.

juliusv: I think while it's a smart approach and yields a condensed mapping in the struct above, as someone who is not really familiar with how the reflection API works, the code below is a bit hard for me to understand. It can also be surprising to readers because it's not an established reflection use case pattern in Go. So personally I'd prefer a bit more verbosity for improved clarity / simplicity - the Go way :)

siavashs: Maybe we can create a package similar to https://github.com/Netflix/go-env but for mapping `procfs` and `sysfs` entries to struct fields later.

discordianfish: If we go down that route, we'd need something that works for other parts of procfs.
So I'd say you can either change this to use a simple amd specific struct or propose a struct tagged based option with examples how it would look like for other parts of procfs. The later is more involved and there will be probably quite some back and forth..

siavashs: I spent sometime last month and I have 3 proposals:
1. Use struct tags and reflection (what we discussed above but at a generic lib)
2. Use interfaces by defining `Class`, `Device`, etc. and a library which accepts those interfaces. Then we can define a `GPU` structure which satisfies the interfaces methods. The library handles reading and generic parsing while the `GPU` structure handles any adjustments or conversions needed on read values, etc.
3. Use code generation, a command will be added to read a `proc` or `sysfs` path which will then generate all the structures and methods for reading and accessing the read values, then custom code can be added on top to handle adjustments or conversions needed on the read values.

Sorry for the late response and the long comment.
Let me know what you think is the proper way to proceed.

discordianfish: 2. sounds like the interfaces might be a bit to abstract to be useful. I think I'd prefer 1 but not convinced it's worth it over just using specific hardcoded structs, at least until we have the need to support multiple variants.

siavashs: I'll start working on 2 then.

siavashs: I misread your comment yesterday so ignore my deleted comment, and indeed interfaces make it abstract.
I'll rewrite the PR to use static way already used in the project.

discordianfish (approved):
LGTM

SuperQ (approved):
LGTM

SuperQ:
Hopefully other GPU vendors also add similar stats.

PR #376, opened 2021-04-11
Add NetClassByIface

corny:
See also: prometheus/node_exporter#1915

Do we need to keep the `NetClass()` method?

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -82,6 +82,19 @@ func (fs FS) NetClassDevices() ([]string, error) {
 	return res, nil
 }
 
+// NetClassByIface returns info for a single net interfaces (iface)
+func (fs FS) NetClassByIface(devicePath string) (*NetClassIface, error) {
+	path := fs.sys.Path(netclassPath)
+
+	interfaceClass, err := parseNetClassIface(filepath.Join(path, devicePath))
+	if err != nil {
+		return nil, err
+	}
+	interfaceClass.Name = devicePath
+
+	return interfaceClass, nil
+}
+
 // NetClass returns info for all net interfaces (iface) read from /sys/class/net/<iface>.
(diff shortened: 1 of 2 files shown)


jan--f:
Sorry I only saw this after opening #378. In any case it's a different approach, would appreciate if you would take a look. It should solve #1915 too.

discordianfish:
Hi,

can you both outline how the changes to node-exporter will look like? That would help figuring out the best way to go

corny:
I prefer to keep out the matching logic from the procfs package and to have only two methods:
* one method for fetching the list of all devices - just the device names/paths
* another method to fetch the info of a single device

discordianfish:
@corny But there is no function for just listing the devices, right?
But I tend to agree that regex based filtering is something that should be done in the consumer of this lib (e.g node-exporter).
@SuperQ wdyt?

jan--f:
> @corny But there is no function for just listing the devices, right?

There is `NetClassDevices()`.
> But I tend to agree that regex based filtering is something that should be done in the consumer of this lib (e.g node-exporter).
> @SuperQ wdyt?

Fwiw this sound good to me. I'd have the node_exporter change ready to go here: https://github.com/prometheus/node_exporter/compare/master...jan--f:netclass-filter-before-parsing?expand=1

discordianfish (approved):
AH gotcha. Okay if we all agree on that, let's go with this one. LGTM

jan--f:
@discordianfish is anything blocking a merge here? I would love to finish up https://github.com/prometheus/node_exporter/pull/2033 and ship that soon.

discordianfish:
Nothing on my side but still needs a second LGTM (from @SuperQ or @pgier)

SuperQ (approved):
LGTM

PR #378, opened 2021-04-19
sysfs/net_class: filter devices by name

jan--f:
This adds regexp-based device filtering to the NetClass function in order to avoid
parsing devices files that are uninteresting.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

Fixes: https://github.com/prometheus/node_exporter/issues/1915

This will have an accompanying PR in the node_exporter, but I wanted to check first if the API change should be made more backwards compatible by introducing a new function that takes the regex as an argument, instead of changing the signature of `NetClassDevices()`.

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/prometheus/procfs/internal/util"
 )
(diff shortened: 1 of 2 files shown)


jan--f:
@discordianfish the accompanying node_exporter patch looks something like this: https://github.com/prometheus/node_exporter/compare/master...jan--f:pass-netclass-filter-to-procfs?expand=1

jan--f:
Keeping the filtering in the consumer is preferred, see #376.

PR #380, opened 2021-04-22
add additional stats from mdstat

johnseekins:
In the recovery line for mdstat, we can also track percentage complete, estimated time to completion, and current recovery write speed. This MR adds those additional stats.


discordianfish:
We need to make sure this works for all kernels we support. If that's the case, LGTM!

johnseekins:
This should gracefully handle different kernels...and I _believe_ the mdstat file format hasn't changed in quite some time.

johnseekins:
@discordianfish Is there a standard way to prove out the "kernels supported"? I know this works on 4.19, for example.

SuperQ:
Typically it involves a lot of browsing the kernel source tree. :-(

discordianfish:
Unfortunately I don't think we have a way. We should probably collect fixtures for multiple kernel versions. It's silly but I'd probably spin up a VM with e.g 2.6.23 and run tests there..

johnseekins:
If that's the case...how did y'all validate this collector in the first place? While I appreciate that we should try to validate this as much as possible, it _does_ fail gracefully on missing stats.

SuperQ:
Lots of trial and error. The current mdstat fixture is a collection of various examples from bugs reported by users.

discordianfish (approved):
Fair enough, I think we should extend the fixtures to cover more kernel versions but until then, I think it's fair to support the procfs files as shown in the fixtures. So LGTM.
@SuperQ wdyt?

SuperQ (approved):
Yup, adding new features while we maintain backwards compatibility is what I desire.

LGTM

PR #381, opened 2021-04-24
Use /proc/self/net instead of /proc/net to avoid kernel crash.

haosdent:
This change is to fix this issue
https://github.com/prometheus/node_exporter/issues/2027 .

### DIFF
--- a/arp.go
+++ b/arp.go
@@ -34,9 +34,9 @@ type ARPEntry struct {
 // GatherARPEntries retrieves all the ARP entries, parse the relevant columns,
 // and then return a slice of ARPEntry's.
 func (fs FS) GatherARPEntries() ([]ARPEntry, error) {
-	data, err := ioutil.ReadFile(fs.proc.Path("net/arp"))
+	data, err := ioutil.ReadFile(fs.proc.Path("self/net/arp"))
 	if err != nil {
-		return nil, fmt.Errorf("error reading arp %q: %w", fs.proc.Path("net/arp"), err)
+		return nil, fmt.Errorf("error reading arp %q: %w", fs.proc.Path("self/net/arp"), err)
 	}
 
 	return parseARPEntries(data)
--- a/ipvs.go
+++ b/ipvs.go
@@ -67,7 +67,7 @@ type IPVSBackendStatus struct {
 
 // IPVSStats reads the IPVS statistics from the specified `proc` filesystem.
 func (fs FS) IPVSStats() (IPVSStats, error) {
-	data, err := util.ReadFileNoStat(fs.proc.Path("net/ip_vs_stats"))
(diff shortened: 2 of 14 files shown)


SuperQ:
I do't see how this is going to help, /proc/self is also a symlink.

haosdent:
@SuperQ how about I change to read `/proc/0/`

In 2 crash dump cases, we just found it could not get the link of `/proc/net` .

haosdent:
```
  d_name = {
    {
      {
        hash = 2805607892,
        len = 3
      },
      hash_len = 15690509780
    },
    name = 0xffff8a4281d01ef8 "net"
  },
  struct inode *d_inode = 0x0  <---- Could not get link inode for `/proc/net` then get "BUG: unable to handle kernel NULL pointer dereference " error.
  -> NULL
```

SuperQ:
That would be a kernel bug then. Not a node_exporter bug.

haosdent:
> That would be a kernel bug then. Not a node_exporter bug.

@SuperQ thanks a lot for your comment, yep, it is kernel bug. but upgrade kernel requires reboot server, if we could workaround in node-exporter, it would be helpful.

discordianfish:
The question is what is the right one in general? The doc says we're using  `/proc/self` https://github.com/prometheus/procfs/blob/5162bec877a860b5ff140b5d13db31ebb0643dd3/proc.go#L43 but it doesn't look like that's the case.

wenjianhn:
It seems like a memory corruption issue:

>        s_shrink = {
>          count_objects = 0xffffffff81878e80 <perf_trace_consume_skb>, 
>          scan_objects = 0xffffffff8187a6d0 <super_cache_scan>, 

The above count_objects was expected to be super_cache_count instead of perf_trace_consume_skb.

The kernel is likely to panic again with or without the patch since the root cause is not fixed.
For example, the kernel is going to panic when reclaim slab pages (if there is not enough free memory or memory is badly fragmented).

haosdent:
@wenjianhn Orz, where you find the complete dentry, I didn't post at here before.

haosdent:
To sync our later finding here in case other guys encounter the similar problem. 
We are able to reproduce the issue via this tiny program.
```
// proc_test.go
package main

import (
        "io/ioutil"
        "math/rand"
        "sync"
        "time"
)

func readSockstat() {
        for {
                _, _ = ioutil.ReadFile("/proc/net/sockstat")
                time.Sleep(time.Millisecond * 10)
        }
}

func main() {
        var wg sync.WaitGroup

        for i := 0; i <= 800; i++ {
                wg.Add(1)
                n := rand.Intn(10)
                time.Sleep(time.Millisecond * time.Duration(n))
                go readSockstat()
        }

        wg.Wait()
}
```
And we found it is fixed in upstream in https://lore.kernel.org/patchwork/patch/1009682/
After applying this patch, could not reproduce the issue any more.

PR #387, opened 2021-06-17
Initial fix for non-detection of OmniPath cards

### LINKED ISSUE #341: parseInfinibandDevice fails with Intel Omni-Path
It looks like [`parseInfinibandDevice`](https://github.com/prometheus/procfs/blob/c5c7bc5412c9e3a99be2ae1d89a3943e0836176e/sysfs/class_infiniband.go#L121) expects to see a file `/sys/class/infiniband/<Name>/hca_type` for each Infiniband device. (As well as `board_id` and `fw_ver`.)

Unfortunately, [this doesn't seem to be provided by the Omni-Path driver](https://github.com/torvalds/linux/blob/98913408c5465ac477f80da7affe347b413edaa4/Documentation/ABI/stable/sysfs-class-infiniband#L616), for some reason.

Downstream, this causes the node exporter to be unable to pick up Omni-Path devices with the Infiniband collector. (Which, I guess they're *strictly* not, but I think it's supposed to cover them as well.)

I'm not 100% sure what's supposed to come out of this file for other types of device: maybe the information from `board_id` is enough? Here's an example of the contents of that:

```
$ cat /sys/class/infiniband/hfi1_0/board_id
Intel Omni-Path Host Fabric Interface Adapter 100 Series
```

And there's also `hw_rev`:

```
$ cat /sys/class/infiniband/hfi1_0/hw_rev
11
```

Or would not producing an error if the `hca_type` file isn't found be appropriate?

ikirker:
Mellanox drivers provide a particular file in the `/sys` pseudo-filesystem that not all InfiniBand drivers do, and OmniPath drivers don't. This means that the InfiniBand detection fails to detect these cards.

This naïve fix makes OmniPath cards detectable by making the InfiniBand detection happy to ignore whether that file is missing.

It could probably use a look in future from someone with more expertise (or at least, more documentation) on the specific info provided by the driver.

This fixes #341 and I think also the downstream issue in the node_exporter: <https://github.com/prometheus/node_exporter/issues/2023>.

### DIFF
--- a/fixtures.ttar
+++ b/fixtures.ttar
@@ -3592,6 +3592,128 @@ Mode: 644
 Directory: fixtures/sys/class/infiniband
 Mode: 755
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Directory: fixtures/sys/class/infiniband/hfi1_0
+Mode: 755
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/infiniband/hfi1_0/board_id
+Lines: 1
+HPE 100Gb 1-port OP101 QSFP28 x16 PCIe Gen3 with Intel Omni-Path Adapter
+Mode: 644
+# ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+Path: fixtures/sys/class/infiniband/hfi1_0/fw_ver
+Lines: 1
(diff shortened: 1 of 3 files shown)


ikirker:
Ah. I had to do some finicky rebasing + merging, and I think that's where the test failures come from -- I tested before but not after, foolishly. I'll fix it up.

discordianfish (approved):
LGTM, thanks!

SuperQ (approved):
LGTM

PR #391, opened 2021-07-01
fix doc.go

wt-l00:
func (fs FS) NewStat() (Stat, error) is deprecated.
ref https://pkg.go.dev/github.com/prometheus/procfs#FS.NewStat.


SuperQ (approved):
LGTM, Thanks.

PR #393, opened 2021-07-05
Add CmdLine function to parse /proc/cmdline

bdrung:
Our company internal Prometheus exporter uses the procfs library and needs to read `/proc/cmdline`. Since `fs.proc.Path` cannot be accessed from outside, add `CmdLine` to procfs to read `/proc/cmdline`.


SuperQ (approved):
LGTM, Thanks

PR #394, opened 2021-07-05
Fix diskstats stats using sysfs

luizluca:
/sys/block/<dev> is normally a symlink while code previously checked
for a directory.

Added flush request and time spend flushing diskstats for kernel 5.5+
to sysfs.

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>


SuperQ (approved):
LGTM, Thanks.

I double checked the Scanf change, looks good.

I'm not sure why there's an IsDir() there. I'll have to look at the code history.

PR #397, opened 2021-07-14
nvme: Wrap underlying errors instead of hiding them

bdrung:
Prometheus node exporter checks the returned error variable of `NVMeClass` for being `os.ErrNotExist`. This check will never match, because `NVMeClass` uses `fmt.Errorf` to convert the error.

So wrap the error result using `%w` to allow checking for `os.ErrNotExist`.

### DIFF
--- a/sysfs/class_nvme.go
+++ b/sysfs/class_nvme.go
@@ -45,7 +45,7 @@ func (fs FS) NVMeClass() (NVMeClass, error) {
 
 	dirs, err := ioutil.ReadDir(path)
 	if err != nil {
-		return nil, fmt.Errorf("failed to list NVMe devices at %q: %v", path, err)
+		return nil, fmt.Errorf("failed to list NVMe devices at %q: %w", path, err)
 	}
 
 	nc := make(NVMeClass, len(dirs))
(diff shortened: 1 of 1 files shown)


sysfs/class_nvme.go
```diff
@@ -45,7 +45,7 @@ func (fs FS) NVMeClass() (NVMeClass, error) {
 
 	dirs, err := ioutil.ReadDir(path)
 	if err != nil {
-		return nil, fmt.Errorf("failed to list NVMe devices at %q: %v", path, err)
+		return nil, err
```
discordianfish, line 48: Thanks, good catch!
But I think we should do this instead:
```suggestion
		return nil, fmt.Errorf("failed to list NVMe devices at %q: %w, path, err)
```
That should also allow checking for the error type: https://blog.golang.org/go1.13-errors

bdrung: Thanks for the pointer to the documentation. That helps a lot since I am still quite new to Go.

Changed and adjusted commit message and merge request.

discordianfish (approved):
Thanks, LGTM!

SuperQ (approved):

Issue #401, opened 2021-07-26
strconv.ParseInt: parsing "18446744073709551615": value out of range


gdejong:
After upgrading `prometheus/procfs` from `0.6.0` to `0.7.0` the `process_cpu_seconds_total`, `process_virtual_memory_bytes` and `process_resident_memory_bytes` metrics from `prometheus/client_golang` were no longer outputted.

We have a Go service running in a Docker container.

Using the `ReportErrors` option from `ProcessCollectorOpts` I was able to display the error.
```
prometheus.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{ReportErrors: true, Namespace: "test_"}))
```

Scraping the metrics gave:
```
An error has occurred while serving metrics:

strconv.ParseInt: parsing "18446744073709551615": value out of range
```

This is the maximum value of a `uint64`

The `/proc/1/stat` file contents:
```
# cat /proc/1/stat
1 (service) S 0 1 1 0 -1 1077944576 4345 0 1 0 13 17 0 0 20 0 5 0 346180303 732237824 2193 18446744073709551615 4194304 9876697 140737276107232 140737276105984 4665027 0 0 0 2143420159 18446744073709551615 0 0 17 4 0 0 1 0 0 14888960 15220704 24559616 140737276112712 140737276112721 140737276112721 140737276112879 0
```

This is parsed in proc_stat.go -> Stat()

The value `18446744073709551615` occurs twice. On position 23 and 33.
Position 33 is written into `ignored` which is defined as an `int`. A value as large as `18446744073709551615` cannot be read into an `int` datatype and this causes the error: `strconv.ParseInt: parsing "18446744073709551615": value out of range`

Possible solution: change the datatype of `ignored` from `int` to `uint64`?

![image](https://user-images.githubusercontent.com/2346366/126991911-2f6c3a74-f469-48e1-9cf8-f785ac4f94ae.png)

gdejong:
Some extra information.

The `/proc/<pid>/stat` format is defined here: https://man7.org/linux/man-pages/man5/proc.5.html 

All fields are listed there with their proper scanf format (https://man7.org/linux/man-pages/man3/scanf.3.html)

The 25th and 35th element are defined as:

```
              (25) rsslim  %lu
                     Current soft limit in bytes on the rss of the
                     process; see the description of RLIMIT_RSS in
                     getrlimit(2).
---

              (35) wchan  %lu  [PT]
                     This is the "channel" in which the process is
[...]

PR #402, opened 2021-07-27
Fix data types for ignored stat fields #401

vykulakov:
Some ignored fields in the `/proc/[pid]/stat` file may have values that are bigger than the max value of the current Go type that is used for the ignored variable. That leads to issues in runtime on some systems that use the maximum possible values for the related fields.

Please note, that this PR fixes data types only for the ignored fields. Other fields in the stat file may be still affected especially on nodes that use 32-bit arch.

@pgier @discordianfish please have a look.

See for details:
* https://man7.org/linux/man-pages/man5/proc.5.html
* https://man7.org/linux/man-pages/man3/scanf.3.html

Fixes: https://github.com/prometheus/procfs/issues/401

Signed-off-by: Vyacheslav Kulakov <kulakov.home@gmail.com>

### DIFF
--- a/proc_stat.go
+++ b/proc_stat.go
@@ -128,7 +128,8 @@ func (p Proc) Stat() (ProcStat, error) {
 	}
 
 	var (
-		ignore int
+		ignoreInt64  int64
+		ignoreUint64 uint64
 
 		s = ProcStat{PID: p.PID, proc: p.fs}
 		l = bytes.Index(data, []byte("("))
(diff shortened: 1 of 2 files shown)


vykulakov:
The interesting thing here is that the fixtures already have examples of the stat file that cannot pass tests but they. Unfortunately, haven't been used in the stat related tests.

SuperQ (approved):
LGTM, nice find.

vykulakov:
> LGTM, nice find.

Should we deal with other fields then? Not in this PR of course.

I guess it will break compatibility but such change may be a good thing for a release (1.x.x I mean).

SuperQ:
This library is still pre 1.0. So, we can add breaking changes pretty easily.

vykulakov:
> This library is still pre 1.0. So, we can add breaking changes pretty easily.

Please, check #403 and let me know if the mailing lists are a better place for such things. Thanks.

discordianfish (approved):
Great, thanks!

discordianfish:
@vykulakov github issues are best (unless it's help/usage questions etc), so thanks for filling!

gdejong:
Thank you! Can you tell me if there is going to be a bugfix release for this?

timja:
Hi I suspect this is causing a crash on startup for arm64 MacOS.

This is the only change that went into 1.2.2 of `node_exporter` it works on 1.2.1

see `node_exporter` issue https://github.com/prometheus/node_exporter/issues/2120

Issue #403, opened 2021-07-27
Change stat data types to match the related kernel data types


vykulakov:
The [man page](https://man7.org/linux/man-pages/man5/proc.5.html) for `/proc/[pid]/stat` describes data types that should be used to parse particular values in this stat file. So the following information about data types may be obtained from there:
```
(1) pid  %d
(2) comm  %s
(3) state  %c
(4) ppid  %d
(5) pgrp  %d
(6) session  %d
(7) tty_nr  %d
(8) tpgid  %d
(9) flags  %u
(10) minflt  %lu
(11) cminflt  %lu
(12) majflt  %lu
(13) cmajflt  %lu
(14) utime  %lu
(15) stime  %lu
(16) cutime  %ld
(17) cstime  %ld
(18) priority  %ld
(19) nice  %ld
(20) num_threads  %ld
(21) itrealvalue  %ld
(22) starttime  %llu
(23) vsize  %lu
(24) rss  %ld
(25) rsslim  %lu
(26) startcode  %lu  [PT]
(27) endcode  %lu  [PT]
(28) startstack  %lu  [PT]
(29) kstkesp  %lu  [PT]
(30) kstkeip  %lu  [PT]
(31) signal  %lu
(32) blocked  %lu
(33) sigignore  %lu
(34) sigcatch  %lu
(35) wchan  %lu  [PT]
(36) nswap  %lu
(37) cnswap  %lu
(38) exit_signal  %d  (since Linux 2.1.22)
(39) processor  %d  (since Linux 2.2.8)
(40) rt_priority  %u  (since Linux 2.5.19)
(41) policy  %u  (since Linux 2.5.19)
(42) delayacct_blkio_ticks  %llu  (since Linux 2.6.18)
(43) guest_time  %lu  (since Linux 2.6.24)
(44) cguest_time  %ld  (since Linux 2.6.24)
(45) start_data  %lu  (since Linux 3.3)  [PT]
(46) end_data  %lu  (since Linux 3.3)  [PT]
(47) start_brk  %lu  (since Linux 3.3)  [PT]
(48) arg_start  %lu  (since Linux 3.5)  [PT]
(49) arg_end  %lu  (since Linux 3.5)  [PT]
(50) env_start  %lu  (since Linux 3.5)  [PT]
(51) env_end  %lu  (since Linux 3.5)  [PT]
(52) exit_code  %d  (since Linux 3.5)  [PT]
```
[...]

PR #404, opened 2021-07-29
Fix data types for CUTime and CSTime stat fields #403

vykulakov:
To @SuperQ @discordianfish 

These two stat fields (`CUTime` and `CSTime`) in the `/proc/[pid]/stat` file should have the `signed long` data type according to the documentation. But currently, in the code, their data type is just `unsigned int`. This commit fixes it and adds more tests.

See for details:
* https://man7.org/linux/man-pages/man5/proc.5.html
* https://man7.org/linux/man-pages/man3/scanf.3.html

Signed-off-by: Vyacheslav Kulakov <kulakov.home@gmail.com>


discordianfish (approved):
Thanks, LGTM!

vykulakov:
It seems I cannot merge this PR. Do we need to wait for yet another approvement?

discordianfish:
Yeah we usually need 2 LGTMs. @SuperQ can you have a look?

SuperQ:
Hrm, we just recently changed these from int to int64 intentionally. We had issues like https://github.com/prometheus/node_exporter/issues/2110.

vykulakov:
> Hrm, we just recently changed these from `int` to `int64` intentionally. We had issues like [prometheus/node_exporter#2110](https://github.com/prometheus/node_exporter/issues/2110).

@SuperQ yep, It was me who changed it to `int64`. But this was too much for the ignored field as they should be just unsigned, that's it. I reread the docs and found that we may be a bit more accurate with specifying data types for the fields. Also, there are tests for cases like in the issue you mentioned. So all should be fine.

Or I may just revert those types for the ignored fields to `int64`. But it affects nothings.

vykulakov:
@SuperQ Actually, the most important thing in the previous PR #402 was replacing `int` with `uint` for most of the ignored fields instead of replacing `int` with `int64`. I used the 64-bit versions of types because didn't know some C data types aspects that I know now. Sorry for that.

SuperQ:
@vykulakov Ok, the only thing I'm worried about is that even tho it says one thing in the procfs docs, I did some digging into the code in the kernel. I'm not so good at C, but it seems like some of the data types are "long long", hard-coding some of these values as 64-bit. But, again, I'm no kernel expert.

vykulakov:
@SuperQ could you point me to this place in the kernel, please? It'll save some time to me.

SuperQ:
I think the actual print definition is https://github.com/torvalds/linux/blob/614cb2751d3150850d459bee596c397f344a7936/fs/proc/array.c#L571-L610. 

Everything there is marked long long, or 64-bit. It seems like the man page is out of date compared to the kernel. Maybe the underlying data type is just a long, but it's now unconditionally printed as a long long.

[This seems to be the commit that changed the output](https://github.com/torvalds/linux/commit/bda7bad62bc4c4e0783348e8db51abe094153c56).

proc_stat.go
```diff
@@ -128,8 +128,8 @@ func (p Proc) Stat() (ProcStat, error) {
 	}
 
 	var (
-		ignoreInt64  int64
-		ignoreUint64 uint64
+		ignoreInt  int
+		ignoreUint uint
```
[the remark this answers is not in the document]
vykulakov: Reverted

vykulakov:
> I think the actual print definition is https://github.com/torvalds/linux/blob/614cb2751d3150850d459bee596c397f344a7936/fs/proc/array.c#L571-L610.
> 
> Everything there is marked long long, or 64-bit. It seems like the man page is out of date compared to the kernel. Maybe the underlying data type is just a long, but it's now unconditionally printed as a long long.
> 
> [This seems to be the commit that changed the output](https://github.com/torvalds/linux/commit/bda7bad62bc4c4e0783348e8db51abe094153c56).

That means that all other fields get affected again. I mean on 32-bit systems most of the fields will take 4 bytes instead of 8 bytes as in that kernel change.

So this PR doesn't make much sense now as it should change all the fields instead just `CUTime` and `CSTime`.

SuperQ (approved):
LGTM, Thanks!

proc_stat_test.go
```diff
@@ -76,16 +77,30 @@ func TestProcStat(t *testing.T) {
...
 	}
+
+	// max values of stat int fields
+	for _, test := range []struct {
+		name string
+		want int
+		have int
+	}{
+		{name: "waited for children user time", want: math.MinInt64, have: s.CUTime},
+		{name: "waited for children system time", want: math.MaxInt64, have: s.CSTime},
```
dswarbrick, line 98: Using MinInt64 / MaxInt64 here causes tests to fail on 32-bit archs, since these are untyped constants, being assigned to machine-native ints (in struct a few lines up).

Either the struct needs to explicitly use int64 types, or you need to use smaller constants.

(incidentally, this is not the first time that there has been a test regression for 32-bit arch in this package)

discordianfish: Makes sense, do you want to open an issue for testing on 32 bit archs? We need to see how to make that possible (let alone who has time to work on it) but agreed that we should track that at least.

dswarbrick: I was able to reproduce the 32-bit test failure simply by exporting `GOARCH=386` before running the tests.

PR #406, opened 2021-08-04
Fix build with GopherJS

paralin:
Fixes build with "gopherjs build ." by excluding proc_maps.go from js and wasm.


paralin:
GopherJS sets both "js" and "linux" - that file references "golang.org/x/sys/unix" which can't be referenced from GopherJS.

As for why, I'm running the Prometheus client and metrics generator in the browser. :)

discordianfish (approved):
Ah makes sense. Thanks!

PR #415, opened 2021-10-22
Check bufio.Scanner errors for CPUInfo parsers

mdlayher:

It's possible for Next to return false due to an I/O error, and the proper way to check that error is by consulting the Err method: https://pkg.go.dev/bufio#Scanner.Err.

### DIFF
--- a/cpuinfo.go
+++ b/cpuinfo.go
@@ -184,6 +184,11 @@ func parseCPUInfoX86(info []byte) ([]CPUInfo, error) {
 			cpuinfo[i].PowerManagement = field[1]
 		}
 	}
+
+	if err := scanner.Err(); err != nil {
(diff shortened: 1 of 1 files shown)


cpuinfo.go
```diff
@@ -184,6 +184,11 @@ func parseCPUInfoX86(info []byte) ([]CPUInfo, error) {
 			cpuinfo[i].PowerManagement = field[1]
 		}
 	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("failed to parse cpuinfo: %v", err)
```
SuperQ, line 189: Please format these with `%w` so the `error` is usable with `errors.Is()`.

SuperQ (changes requested):

Issue #418, opened 2021-10-28
Lnstat error: strconv.ParseUint: parsing "125ee882a": value out of range


bdrung:
After upgrading to the latest prometheus-node-exporter build from today's master branch (commit 6523fdfc4b20f42269c849cde95a413c232d38a0) with procfs 0.7.3, we see some clusters have failing lnstat collectors:

```
prometheus-node-exporter[59966]: level=error ts=2021-10-28T15:16:33.821Z caller=collector.go:169 msg="collector failed" name=lnstat duration_seconds=0.000423308 err="Lnstat error: strconv.ParseUint: parsing \"125ee882a\": value out of range"
```

The hex values comes from `/proc/net/stat/arp_cache`:
```
$ head -n2 /proc/net/stat/arp_cache 
entries  allocs destroys hash_grows  lookups hits  res_failed  rcv_probes_mcast rcv_probes_ucast  periodic_gc_runs forced_gc_runs unresolved_discards table_fulls
00000431  00223378 00425a10 00000007  126152242 122ba4ce5  000352a9  00000000 00000000  00028463 00000000 00000000 00000000
```

These hosts use a custom linux 5.4 kernel.

discordianfish:
Should be fixed by #419

PR #419, opened 2021-10-30
netstat change uint size from 32 to 64 bits

binjip978:

### DIFF
--- a/netstat.go
+++ b/netstat.go
@@ -23,8 +23,8 @@ import (
 
 // NetStat contains statistics for all the counters from one file
 type NetStat struct {
-	Filename string
 	Stats    map[string][]uint64
+	Filename string
 }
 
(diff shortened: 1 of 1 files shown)


binjip978:
fixes https://github.com/prometheus/procfs/issues/418 
also minor change to struct layout to use 16 bytes instead of 24 according to fieldalignment analyser.

netstat.go
```diff
@@ -23,8 +23,8 @@ import (
 
 // NetStat contains statistics for all the counters from one file
 type NetStat struct {
-	Filename string
 	Stats    map[string][]uint64
+	Filename string
```
SuperQ, line 27: This change seems unrelated.

SuperQ: Oh, sorry, I missed the comment on struct alignment. Thanks!

SuperQ (approved):
LGTM

PR #421, opened 2021-11-11
fix bug when /proc/net/dev contains unexpected line

bo-er:
In centos 8(4.18.0-193.28.1.el8_2.x86_64) if I try this command:
```
ip link add team0:1 type dummy
```
I will get a message: 
```
"RTNETLINK answers: Invalid argument". 

```
However if I do the same thing on Centos 7(3.10.0-862.14.4.el7.x86_64), there is no error ! And If I cat /proc/net/dev I can see there is a sub interface:
```
Inter-|   Receive                                                |  Transmit
 face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
  eth0: 2489766199  690315    0    0    0     0          0         0 96022876  713917    0    0    0     0       0          0
    lo: 402582904  916378    0    0    0     0          0         0 402582904  916378    0    0    0     0       0          0
team0:1:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
```
So the problem is currently 'procfs' cann't parse /proc/net/dev that contains odd line of interfaces.

### DIFF
--- a/net_dev.go
+++ b/net_dev.go
@@ -16,7 +16,6 @@ package procfs
 import (
 	"bufio"
 	"errors"
-	"fmt"
 	"os"
 	"sort"
 	"strconv"
@@ -76,8 +75,7 @@ func newNetDev(file string) (NetDev, error) {
 
 		line, err := netDev.parseLine(s.Text())
 		if err != nil {
-			fmt.Printf("[ERROR] failed to parse line: %v", s.Text())
-			continue
+			return netDev, err
 		}
 
 		netDev[line.Name] = *line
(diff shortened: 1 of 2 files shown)


discordianfish:
Interesting.. well the actual problem is that this assumes a device name can't include `:`: https://github.com/prometheus/procfs/pull/421/files#diff-a691799f566ae988549d796fb8df91095344d63552424e79c7687a14dcaaf601R92

But we shouldn't just ignore parse errors. Instead, if we want to fix that, we need to parse this differently.

discordianfish (changes requested):
Interesting.. well the actual problem is that this assumes a device name can't include `:`:  https://github.com/prometheus/procfs/pull/421/files#diff-a691799f566ae988549d796fb8df91095344d63552424e79c7687a14dcaaf601R92

But we shouldn't just ignore parse errors. Instead, if we want to fix that, we need to parse this differently.

discordianfish:
Thanks! Can you also add test cases for dev names that include :?

bo-er:
@discordianfish hello! 🙇🏻‍♂️ Do I need to add more test cases?

discordianfish (approved):
LGTM, thanks!

discordianfish:
you need to sign your commit though before we can merge. See the DCO check

bo-er:
> you need to sign your commit though before we can merge. See the DCO check

I have signed my commits.

SuperQ:
We should probably fix this the same way we fixed it in the node_exporter. I forgot we have yet to convert the node_exporter to this parser.

https://github.com/prometheus/node_exporter/pull/904
https://github.com/prometheus/node_exporter/pull/910

SuperQ (approved):
Actually, I like this better, it avoids use of regexp.

PR #424, opened 2021-12-03
add CgroupSummary to parse /proc/cgroups

biello:


discordianfish (commented):
LGTM, thanks!

proc_cgroups.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2020 The Prometheus Authors
```
SuperQ, line 1: Nit

```go
// Copyright 2021 The Prometheus Authors
```

SuperQ (commented):
Minor nit, otherwise LGTM.

proc_cgroups.go
biello: updated

PR #425, opened 2021-12-11
Add a missing name to the *testing.T argument.

nerdatmath:
When using t.Run, the subtest must take a named *testing.T argument. Otherwise, inner calls to t.Error etc. are invoked on the outer test object, which is pretty much always wrong.

@discordianfish please review.


discordianfish (approved):
Good catch, thanks!

PR #429, opened 2021-12-24
try best to parse data in /proc/stat

yyrdl:
Issue Related : [Issues 1882 ](https://github.com/prometheus/node_exporter/issues/1882 )

The background:

Some version of linux will report a negative value  of process number  in /proc/stat file,  it will cause an error when parse the data in procsfs/stat.go  ,  and this results in loss of CPU load data (and other metrics). Although it's a bug of kernel, but the cost of upgrading  kernel  is too expensive , maybe  let node-exporter trying best to collect  is an acceptable solution .

I will submit another pr to node-exporter after this merged.


testdata/proc1/stat
```diff
@@ -0,0 +1,16 @@
...
+ctxt 38014093
+btime 1418183276
+processes -26442
```
yyrdl, line 13: This is  what we met ...

SuperQ:
Can you provide more information on what kernel produces this? From the previous investigation, it's impossible to produce these values outside of a corrupt system. All kernel versions I could find use an unsigned long value to represent this data.

yyrdl:
> Can you provide more information on what kernel produces this? From the previous investigation, it's impossible to produce these values outside of a corrupt system. All kernel versions I could find use an unsigned long value to represent this data.

@SuperQ I've confirmed this issue ， it won't appear in linux of community version ,and it is because someone of   kernel team in Tencent made a mistake.

SuperQ:
Ok, thanks for confirming. This is a vendor bug, not something we want to support here.

PR #433, opened 2022-01-20
Add missing godoc comments for xfrm stats

tklauser:
Add missing godoc comments matching the kernel documentation, see
https://www.kernel.org/doc/html/latest/networking/xfrm_proc.html

Also rename the `xfrm*.go` files to have a `net_` prefix, as the
respective file is found in the `net/` sub-directory and this matches
filenames for other network-related stats.


discordianfish (approved):
Thanks!

PR #435, opened 2022-02-14
Introduce parsing for `/proc/<pid>/net/snmp`, `/proc/<pid>/net/snmp6` and `/proc/<pid>/net/netstat`.

nikakis:


SuperQ:
Nice, do we have to worry about missing fields in various kernel versions? We like to be able to support very old kernels in this library, since Go will run on kernels as old as 2.6.23.

nikakis:
@SuperQ I believe according to [this RFC](https://datatracker.ietf.org/doc/rfc1213/) it seems that we will be fine.

SuperQ:
Any chance I can also get you to implement snmp6 parsing in this PR too?

nikakis:
Sure i will implement snmp6 and netstat also.
Maybe i can also do the refactoring on the node_exporter side when this is done.

nikakis:
@SuperQ I added also snmp6 and netstat. Have a look please!

.gitignore
```diff
@@ -1 +1,2 @@
 /fixtures/
+.idea
```
SuperQ, line 2: We don't put editor-specific ignores in our repos.

proc_netstat.go
```diff
@@ -0,0 +1,222 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procfs
+
+import (
+	"bufio"
+	"fmt"
+	"github.com/mitchellh/mapstructure"
```
SuperQ, line 19: Hmm, I'm not sure we want to add this dependency. But maybe it would be useful for other stuff.

CC @discordianfish @roidelapluie

SuperQ: Also, we split stdlib and external deps like this:

```
import (
	"fmt"
	"os"

	"github.com/mitchellh/mapstructure"
)
```

proc_snmp.go
```diff
@@ -0,0 +1,187 @@
...
+	SndbufErrors float64
+	InCsumErrors float64
+	IgnoredMulti float64
+}
+
+func (p Proc) Snmp() (ProcSnmp, error) {
+	filename := p.path("net/snmp")
+	procSnmp := ProcSnmp{PID: p.PID}
+
+	file, err := os.Open(filename)
```
SuperQ, line 139: For files with relatively known fixed sizes, we prefer to use [`ReadFileNoStat()`](https://pkg.go.dev/github.com/prometheus/procfs@v0.7.3/internal/util#ReadFileNoStat). This reader makes sure we gather the file contents as quickly as possible with the fewest IO operations. This is because the kernel doesn't provide any mutex locking around procfs/sysfs file reads.

SuperQ (changes requested):

proc_netstat.go
nikakis: @SuperQ I resolved the rest of the comments and amend the last commit. 
How i should proceed with the `mapstructure` dependency?

discordianfish: Hrm yeah not sure either. To get this merged asap, I'd remove the dependency assign manually to the struct for now.

But I see something like that being useful for similar cases... If we want to do that, we should right away use this for these cases to establish that pattern for further contributions.

nikakis: @discordianfish Which are those cases?

nikakis: actually indeed there is a lot verbosity on the parsing of map to structure e.g. [here](https://github.com/prometheus/procfs/blob/master/blockdevice/stats.go#L346). Maybe we can merge this PR and i can start asap working on replacing those verbose parsing cases with `mapstructure`?

discordianfish: I think this deserve a longer discussion. So if you want this to get merged soon, remove mapstructure and assign these manually.

proc_netstat.go
```diff
@@ -205,19 +194,28 @@ func parseNetstat(r io.Reader, fileName string) (map[string]map[string]float64,
...
+			key := nameParts[i]
+
+			var s interface{}
+			if protocol == "TcpExt" {
+				s = &procNetstat.TcpExt
 			}
+			if protocol == "IpExt" {
+				s = &procNetstat.IpExt
+			}
+			structValue := reflect.ValueOf(s).Elem()
```
discordianfish, line 215: We try to avoid using reflect. I'd use a switch/case statement to map the values. A bit verbose but should be fine.

proc_snmp.go
```diff
@@ -170,19 +159,40 @@ func parseSnmp(r io.Reader, fileName string) (map[string]map[string]float64, err
...
-			netStats[protocol][nameParts[i]], err = strconv.ParseFloat(valueParts[i], 64)
+			value, err := strconv.ParseFloat(valueParts[i], 64)
 			if err != nil {
-				return nil, err
+				return procSnmp, err
 			}
+			key := nameParts[i]
+
+			var s interface{}
+			if protocol == "Ip" {
```
discordianfish, line 174: Use ```
switch protocol {
case "Ip"...
```
instead

File: proc_netstat.go (hunk 4 above)
Comment (nikakis) [reply proc_netstat.go:215]: @discordianfish Sure i will do, but just from curiosity why we avoid `reflect`?

**Comment by nikakis** [comment]
@discordianfish @SuperQ Have a look again please!

**Review by discordianfish — APPROVED** [review]
Great, LGTM!

**Comment by nikakis** [comment]
Hello! @discordianfish @SuperQ 
Sorry for spamming but any update with merging this?
Thanks!

**Comment by discordianfish** [comment]
Still needs a LGTM from @SuperQ or @pgier

File: proc_netstat.go [hunk 6]
```diff
@@ -0,0 +1,440 @@
...
+		scanner  = bufio.NewScanner(r)
+		procNetstat = ProcNetstat{}
+	)
+
+	for scanner.Scan() {
+		nameParts := strings.Split(scanner.Text(), " ")
+		scanner.Scan()
+		valueParts := strings.Split(scanner.Text(), " ")
+		// Remove trailing :.
+		protocol := nameParts[0][:len(nameParts[0])-1]
```
pgier, line 195: Any reason not to use `strings.TrimSuffix` instead?

proc_netstat.go
```diff
@@ -0,0 +1,440 @@
...
+		if len(nameParts) != len(valueParts) {
+			return procNetstat, fmt.Errorf("mismatch field count mismatch in %s: %s",
+				fileName, protocol)
+		}
+		for i := 1; i < len(nameParts); i++ {
```
pgier, line 200: Could probably use a range over `nameParts` here instead of the explicit counter.

proc_snmp.go
```diff
@@ -0,0 +1,353 @@
...
+		scanner  = bufio.NewScanner(r)
+		procSnmp = ProcSnmp{}
+	)
+
+	for scanner.Scan() {
+		nameParts := strings.Split(scanner.Text(), " ")
+		scanner.Scan()
+		valueParts := strings.Split(scanner.Text(), " ")
+		// Remove trailing :.
+		protocol := nameParts[0][:len(nameParts[0])-1]
```
pgier, line 160: Similar to the other file, I think you can just use `strings.TrimSuffix`

pgier (approved):
LGTM, just had a couple of optional in-line suggestions.

proc_netstat.go
nikakis: Actually i need to pick elements from both `nameParts` and `valueParts`, i am not sure if a range over is possible here.

pgier: hmm, I guess that wouldn't work because you want to start at index 1 instead of 0.  So you'd have to add something like
```
nameParts = nameParts[1:]
valueParts = valueParts[1:]
```
So I think just leaving it as with the explicit counter is fine, or whatever you prefer.

nikakis: @pgier I think yes, lets keep it with the explicit counter.

proc_snmp6.go
```diff
@@ -0,0 +1,381 @@
...
+	procSnmp6, err := parseSNMP6Stats(bytes.NewReader(data))
+	procSnmp6.PID = p.PID
+	return procSnmp6, err
+}
+
+// parseSnmp6 parses the metrics from proc/<pid>/net/snmp6 file
+// and returns a map contains those metrics.
+func parseSNMP6Stats(r io.Reader) (ProcSnmp6, error) {
+	var (
+		scanner  = bufio.NewScanner(r)
```
discordianfish, line 163: How did this gofmt issue not surfaced in the PR? @SuperQ @pgier any idea?

nikakis: I will fix this fmt issue and also try to investigate why this did worked properly

SuperQ: Weird. Maybe out of date source branch?

Issue #432, opened 2022-01-14
TestProcStatLimits: Int overflow on 32 bits arches


eclipseo:
On both i686 and armv7hl, TestProcStatLimits overflows:

```
# github.com/prometheus/procfs [github.com/prometheus/procfs.test]
./proc_stat_test.go:97:49: cannot use math.MinInt64 (untyped int constant -9223372036854775808) as int value in struct literal (overflows)
./proc_stat_test.go:98:51: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in struct literal (overflows)
```

dswarbrick:
Linking to PR #404 which introduced this regression.

The crux of the issue is that the shipped testdata fixtures will overflow ints in the ProcStat struct on 32-bit archs.

dswarbrick:
In terms of fixing this, we ran into a pretty similar situation before, which I fixed in #314, i.e. split the tests into 32-bit and 64-bit variants, using Go build tags to conditionally perform the right test on the right arch.

I guess it boils down to how imperative it is that we ship testdata fixtures which exercise the full range of machine-native ints / uints.

PR #437, opened 2022-03-09
proc_stat_test.go: architecture dependent MinInt and MaxInt constants

TimRots:
to avoid `math.MinInt64` and `math.MaxInt64` overflowing `int` type
when running `TestProcStatLimits` on 32bit architectures:
```
./proc_stat_test.go:97:49: cannot use math.MinInt64 (untyped int constant -9223372036854775808) as int value in struct literal (overflows)
./proc_stat_test.go:98:51: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in struct literal (overflows)
```

Fixes: #432

Signed-off-by: Tim Rots <tim.rots@protonmail.ch>

### DIFF
--- a/proc_stat_test.go
+++ b/proc_stat_test.go
@@ -14,7 +14,7 @@
 package procfs
 
 import (
-	"math"
+	"math/bits"
 	"os"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


discordianfish:
Hrmm how is this working with the fixture having the same values on all archs?

TimRots:
I clearly didn't understand the issue. Sorry for this

PR #438, opened 2022-03-15
parsing /proc/stat by CPU ID

alexjx:
We have a case that some of the cpu (threads) are offline intentionally (see lscpu below),  procfs will "interpolate" those offline values. This patch made the /proc/stat parsing with the real cpu ID.

```
Architecture:         x86_64
CPU op-mode(s):       32-bit, 64-bit
Byte Order:           Little Endian
CPU(s):               16
On-line CPU(s) list:  0,2,4,6,8,10,12,14
Off-line CPU(s) list: 1,3,5,7,9,11,13,15
Thread(s) per core:   1
Core(s) per socket:   8
Socket(s):            1
NUMA node(s):         1
Vendor ID:            GenuineIntel
CPU family:           6
Model:                106
Model name:           Intel(R) Xeon(R) Gold 6354 CPU @ 3.00GHz
Stepping:             6
CPU MHz:              2993.062
BogoMIPS:             5986.12
Virtualization:       VT-x
Hypervisor vendor:    KVM
Virtualization type:  full
L1d cache:            32K
L1i cache:            32K
L2 cache:             4096K
L3 cache:             16384K
NUMA node0 CPU(s):    0,2,4,6,8,10,12,14
Flags:                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon rep_good nopl xtopology cpuid tsc_known_freq pni pclmulqdq vmx ssse3 fma cx16 pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch cpuid_fault invpcid_single ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves wbnoinvd arat avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq rdpid md_clear arch_capabilities
```

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -62,7 +62,7 @@ type Stat struct {
 	// Summed up cpu statistics.
 	CPUTotal CPUStat
 	// Per-CPU statistics.
-	CPU []CPUStat
+	CPU map[int64]CPUStat
 	// Number of times interrupts were handled, which contains numbered and unnumbered IRQs.
 	IRQTotal uint64
 	// Number of times a numbered IRQ was triggered.
@@ -171,7 +171,9 @@ func (fs FS) Stat() (Stat, error) {
 		return Stat{}, err
 	}
 
-	stat := Stat{}
+	stat := Stat{
+		CPU: make(map[int64]CPUStat),
+	}
 
 	scanner := bufio.NewScanner(bytes.NewReader(data))
 	for scanner.Scan() {
@@ -228,9 +230,6 @@ func (fs FS) Stat() (Stat, error) {
 			if cpuID == -1 {
 				stat.CPUTotal = cpuStat
 			} else {
-				for int64(len(stat.CPU)) <= cpuID {
-					stat.CPU = append(stat.CPU, CPUStat{})
-				}
 				stat.CPU[cpuID] = cpuStat
 			}
 		}


discordianfish:
Just the newStat() thing, generally this seems reasonable, thanks!

stat.go
```diff
@@ -142,14 +142,20 @@ func parseSoftIRQStat(line string) (SoftIRQStat, uint64, error) {
...
+
 // NewStat returns information about current cpu/process statistics.
 // See https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 //
 // Deprecated: Use fs.Stat() instead.
 func NewStat() (Stat, error) {
 	fs, err := NewFS(fs.DefaultProcMountPoint)
 	if err != nil {
-		return Stat{}, err
+		return newStat(), err
```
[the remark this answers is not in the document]
alexjx: Since we've changed the CPUs into a map, we have to initialzie it. Return Stat{} will initialize the map in nil, which burdens initialization in the usage site.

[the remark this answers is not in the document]
discordianfish: I think just returning Stat{} with map being nil is fine. If an error is returned, the caller shouldn't use it anyway, right?

stat.go
```diff
@@ -62,7 +62,7 @@ type Stat struct {
 	// Summed up cpu statistics.
 	CPUTotal CPUStat
 	// Per-CPU statistics.
-	CPU []CPUStat
+	CPU map[int64]*CPUStat
```
pgier, line 65: I think it would be better to just use a `CPUStat` instead of pointer `*CPUStat` here.  The struct is relatively small, so I think memory shouldn't be an issue.

stat.go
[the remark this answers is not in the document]
pgier: I agree with @discordianfish, I think just returning an empty Stat{} with nil map should be fine.

stat.go
```diff
@@ -168,10 +174,10 @@ func (fs FS) Stat() (Stat, error) {
 	fileName := fs.proc.Path("stat")
 	data, err := util.ReadFileNoStat(fileName)
 	if err != nil {
-		return Stat{}, err
+		return newStat(), err
 	}
 
-	stat := Stat{}
+	stat := newStat()
```
pgier, line 180: I think you can just do an in-line initialization here instead of the call to newStat()

stat.go
alexjx: We started with `CPUStat` initially, the problem is in node_exporter, it tracks `CPUStat` with the same data structure. Working with references makes both implementation clean. If pointer should be avoided here, I could see what I can do in the node_exporter side.

stat.go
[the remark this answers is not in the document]
alexjx: OK.

stat.go
discordianfish: @alexjx Why does it make implementation cleaner?

alexjx: Sorry for the long delay. What I thought back then was that by making it a pointer, we have the ablity to test if the value "exist" in certain senarios. But reviewing node_exporter implementation, It seem the case is rare. I will try updating the patch.

discordianfish (approved):
Thanks!

PR #439, opened 2022-03-21
Process sysfs/class/net/*/bonding

bewing:
As part of prometheus/node_exporter#1604, parse information from `/sys/class/net/*/bonding*/*` to provide information regarding current bonding state for bonding controllers and their devices.

I have some style/format questions that I would appreciate input on as part of the process:

* There does not appear to be consistency on pointers vs values in many modules of this project.  I notice that `net_class.go` uses pointers for integers, but not strings.  Should I follow that, or is there a different preference from the maintainers? It would appear that using pointers would allow for `nil` values to indicate kernel-level support for a specific file (if the kernel doesn't have the file present, better to indicate with a nil pointer rather than a default empty string or integer?), but I'm curious what the consensus is in the project for indicating this
* The kernel driver has many different sized ints (`uint8`, `int64`, `uint64`, etc) -- should I try to match what the current kernel defines these as, or just use `uint64`s as the worst case storage requirement?
* The bonding driver has many files of the format `<uint> <human readable value string>` -- should this be implemented as constants in the module itself, or imported constants from a reference module (assuming one exists), or some other approach I haven't thought of?
* Should I be using pointers to other `NetClassIface` structs to reference controllers and devices (IE, L34, L64 ?)
* Should any special care/documentation be taken regarding fields that need CAP_NET_ADMIN to read?

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -18,8 +18,11 @@ package sysfs
 import (
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/prometheus/procfs/internal/util"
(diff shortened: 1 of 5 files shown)


discordianfish:
> There does not appear to be consistency on pointers vs values in many modules of this project. I notice that net_class.go uses pointers for integers, but not strings. Should I follow that, or is there a different preference from the maintainers? It would appear that using pointers would allow for nil values to indicate kernel-level support for a specific file (if the kernel doesn't have the file present, better to indicate with a nil pointer rather than a default empty string or integer?), but I'm curious what the consensus is in the project for indicating this

yes, we use pointers to distinguish between a null value and e.g non existing file as in your example. Other reasons would be larger structs where you want to avoid copying them when you pass them to functions.

> The kernel driver has many different sized ints (uint8, int64, uint64, etc) -- should I try to match what the current kernel defines these as, or just use uint64s as the worst case storage requirement?

Yes we try to mimic the underlying sized ints

>  The bonding driver has many files of the format <uint> <human readable value string> -- should this be implemented as constants in the module itself, or imported constants from a reference module (assuming one exists), or some other approach I haven't thought of?

Not sure what you mean by that.

> Should I be using pointers to other NetClassIface structs to reference controllers and devices (IE, L34, L64 ?)

If you retrieved the data for them I see why not, but we probably shouldn't waste cycles on getting them if we don't need them in which case I'd just use some string? identifier.

> Should any special care/documentation be taken regarding fields that need CAP_NET_ADMIN to read?

I don't think we have any convention for that.

bewing:
> > The bonding driver has many files of the format   -- should this be implemented as constants in the module itself, or imported constants from a reference module (assuming one exists), or some other approach I haven't thought of?
> 
> Not sure what you mean by that.
> 

For example, `sys/class/net/bond0/bonding/ad_select` :

```
$ cat /sys/class/net/bond0/bonding/ad_select
stable 0
```

The actual underlying stored kernel value for ad_select is an integer, but [bond_opt_get_val](https://github.com/torvalds/linux/blob/d888c83fcec75194a8a48ccd283953bdba7b2550/drivers/net/bonding/bond_options.c#L504) is used to resolve that integer 0 to the string "stable" via a constant lookup table [bond_ad_select_tbl](https://github.com/torvalds/linux/blob/d888c83fcec75194a8a48ccd283953bdba7b2550/drivers/net/bonding/bond_options.c#L158-L163), and then both are present in the sysfs file.  Should we return the integer, just the string, or a generic bond_options struct with both values?

bewing:
I'm not able to run tests in node_exporter using this branch, so there's still more digging/dev work needed before this is mergable

sysfs/net_class.go
```diff
@@ -194,6 +256,247 @@ func parseNetClassIface(devicePath string) (*NetClassIface, error) {
...
+				if err != nil {
+					return fmt.Errorf("unable to read %s", filepath.Join(path, "slaves"))
+				}
+				for _, device := range strings.Split(devices, " ") {
+					if intf, exists := (*netClass)[device]; exists {
+						netClassIface.BondAttrs.Devices = append(netClassIface.BondAttrs.Devices, &intf)
+					} else {
+						return fmt.Errorf("unable to find device %s", device)
+					}
+				}
```
bewing, line 459: So it turns out the errors that I was seeing trying to test node_exporter against my branch are because they have an incomplete fixture set:
https://github.com/prometheus/node_exporter/blob/b52bf958f8e2d4ed1624b8122d08af4d12da9322/collector/fixtures/sys.ttar#L1094-L1098

Member devices eth1, eth4, and eth5 are not present in the fixtures.  Does it make more sense to fix the upstream fixtures to be complete (either by removing the references to the non-existing devices, or adding dummy fixtures for them), or by not resolving them to pointers here and not returning an error at all?

discordianfish: If we're sure that on real systems the member would be present, which I expect, I'd say to update the fixtures.
I'd include the other interfaces if it's not inflating the fixture sizes to much, otherwise remove the references.

discordianfish:
> The actual underlying stored kernel value for ad_select is an integer, but [bond_opt_get_val](https://github.com/torvalds/linux/blob/d888c83fcec75194a8a48ccd283953bdba7b2550/drivers/net/bonding/bond_options.c#L504) is used to resolve that integer 0 to the string "stable" via a constant lookup table [bond_ad_select_tbl](https://github.com/torvalds/linux/blob/d888c83fcec75194a8a48ccd283953bdba7b2550/drivers/net/bonding/bond_options.c#L158-L163), and then both are present in the sysfs file. Should we return the integer, just the string, or a generic bond_options struct with both values?

I'd say just the string but I wonder why they expose both the string and the int in procfs and if we should take that decisiopn into account here..

bewing:
Oh what a difference 4 years makes.

Updated branch, node_exporter tests no longer fail.

Can we re-touch on the questions I had in the initial PR, and see if we can get this to a mergable state?

```
$ git diff
diff --git a/go.mod b/go.mod
index 31d63c20..1bc7d351 100644
--- a/go.mod
+++ b/go.mod
@@ -31,6 +31,8 @@ require (
        howett.net/plist v1.0.1
 )

+replace github.com/prometheus/procfs => ../procfs
+
 require (
        cyphar.com/go-pathrs v0.2.2 // indirect
        github.com/alecthomas/units v0.0.0-20240927000941-0f3dac36c52b // indirect
diff --git a/go.sum b/go.sum
index ae4bd585..25cfca5d 100644
--- a/go.sum
+++ b/go.sum
@@ -84,8 +84,6 @@ github.com/prometheus/common v0.67.5 h1:pIgK94WWlQt1WLwAC5j2ynLaBRDiinoAb86HZHTU
 github.com/prometheus/common v0.67.5/go.mod h1:SjE/0MzDEEAyrdr5Gqc6G+sXI67maCxzaT3A2+HqjUw=
 github.com/prometheus/exporter-toolkit v0.16.0 h1:xT/j7L2XKF+VJd6B4fpUw6xWabHrSmsUf6mYmFqyu0s=
 github.com/prometheus/exporter-toolkit v0.16.0/go.mod h1:d1EL8Z9674xQe/iWhwP2wDyCEoBPbXVeqDbqAUsgJWY=
-github.com/prometheus/procfs v0.20.1 h1:XwbrGOIplXW/AU3YhIhLODXMJYyC1isLFfYCsTEycfc=
-github.com/prometheus/procfs v0.20.1/go.mod h1:o9EMBZGRyvDrSPH1RqdxhojkuXstoe4UlK79eF5TGGo=
 github.com/safchain/ethtool v0.7.0 h1:rlJzfDetsVvT61uz8x1YIcFn12akMfuPulHtZjtb7Is=
 github.com/safchain/ethtool v0.7.0/go.mod h1:MenQKEjXdfkjD3mp2QdCk8B/hwvkrlOTm/FD4gTpFxQ=
 github.com/siebenmann/go-kstat v0.0.0-20210513183136-173c9b0a9973 h1:GfSdC6wKfTGcgCS7BtzF5694Amne1pGCSTY252WhlEY=

$ go clean -testcache
$ make test
>> running tests
go test -short -race ./...
ok      github.com/prometheus/node_exporter     1.107s
ok      github.com/prometheus/node_exporter/collector   1.602s
ok      github.com/prometheus/node_exporter/collector/utils     1.047s
?       github.com/prometheus/node_exporter/tools       [no test files]
```

bewing:
> I'd say just the string but I wonder why they expose both the string and the int in procfs and if we should take that decisiopn into account here..

It appears the pattern that was used in the Infiniband section was to include both:  
https://github.com/prometheus/procfs/blob/506a32c52055a5c2d8eb8068c3a0027ae3ae9ec8/sysfs/class_infiniband.go#L113-L116

SuperQ:
Thanks for coming back to working on this. Looks like there's some testing issues.

bewing:
> Thanks for coming back to working on this. Looks like there's some testing issues.

Pushed new commit with linting fixes.

bewing:
Weird that gofmt didn't fix this for me.  

```nowrap
$ golangci-lint run ./sysfs/
sysfs/net_class_test.go:23:1: File is not properly formatted (goimports)
        "github.com/google/go-cmp/cmp"
^
1 issues:
* goimports: 1

```

pushed new changeset after running golangci-lint.

SuperQ:
Yea, we have some `golangci-lint` format tools here.

bewing:
@discordianfish any chance we can get a review on this?  Would love to use this in node_exporter to monitor things like speed mismatches, partial outages, etc on bonded interfaces

jforce:
@discordianfish - sorry to chase, can we get a review of this please? :)

PR #442, opened 2022-04-14
Fixup linting

SuperQ:
* Fix check license issues.
* Ignore some revive naming issues.
* Update minimum Go versions.
* Update go fmt for newer Go versions
* Update Go CI versions.
* Repac fixtures.
* Use cimg/go.

### DIFF
--- a/proc_netstat.go
+++ b/proc_netstat.go
@@ -32,7 +32,7 @@ type ProcNetstat struct {
 	IpExt
 }
 
-type TcpExt struct {
+type TcpExt struct { // nolint:revive
 	SyncookiesSent            float64
 	SyncookiesRecv            float64
 	SyncookiesFailed          float64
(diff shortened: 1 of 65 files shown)


proc_netstat.go
```diff
@@ -32,7 +32,7 @@ type ProcNetstat struct {
 	IpExt
 }
 
-type TcpExt struct {
+type TcpExt struct { // nolint:revive
```
pgier, line 35: Since this was just recently added, should we just change it to TCPExt so we don't need to skip linting?  And similar question for the other recently added lint failures.

pgier (approved):
LGTM

proc_netstat.go
SuperQ: I was thinking about following up with a big breaking change to clean up the nolint.

PR #443, opened 2022-04-15
Reformat the proc_netstat.go file to go fmt format

nikakis:


pgier:
I think this is fixed by fba575dda75ac3de1e94ee7660f93c9367151935

PR #446, opened 2022-04-22
bcache: remove sys/fs cache_readaheads metrics

joshniec:
Support for bcache cache_readahead_totals was removed in Kernel Version 5.13: https://lkml.org/lkml/2021/5/31/1602

This is a pre-requisite to removing support for cache_readaheads in node_exporter to resolve https://github.com/prometheus/node_exporter/issues/2103


lechaimwilson:
👍🏽

fixtures.ttar
```diff
@@ -4879,7 +4879,7 @@ Mode: 664
 # ttar - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 Path: fixtures/sys/class/nvme/nvme0/model
 Lines: 1
-Samsung SSD 970 PRO 512GB               
+Samsung SSD 970 PRO 512GB
```
SuperQ, line 4882: It looks like you're re-writing unrelated fixtures. Some of these fixtures have intentional extra whitespace and other things since the source kernel files have them.

Please only change fixtures that are related to your change.

bcache/bcache.go
```diff
@@ -81,7 +81,6 @@ type PeriodStats struct {
 	CacheHits           uint64
 	CacheMissCollisions uint64
 	CacheMisses         uint64
-	CacheReadaheads     uint64
```
SuperQ, line 84: Hmm, do we want to remove this completely? Or detect the removed metric and provide a `nil`.

joshniec: You're right, I was thinking about this after submitting this patch too. This metric is obviously still available on kernel version < 5.13.

joshniec: @SuperQ I'm new to Golang (and not a dev by trade), but I think I was able to accomplish detecting when the metric is removed, but I wasn't able to get it work when returning nil (understandable because the return type is `unit64`).

SuperQ (approved):
LGTM, thanks

I think we should eventually refactor this. The `readValue()` is a bit non-standard compared to the usual methods used by this library. A task for another day.

bcache/bcache.go
SuperQ: Looks great!

PR #447, opened 2022-05-04
Move fixtures to testdata folder

dprotaso:
This prevents the test fixtures from being shipped around in the module
cache. ie. Prior if project A has a dependency on this project then running
`go mod vendor` will pull the fixtures.ttar into the vendor folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
-/fixtures/
+/testdata/fixtures/
+/fixtures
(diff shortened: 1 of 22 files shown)


dprotaso:
@SuperQ this is ready

discordianfish:
Oh I did not know that testdata/ is treated differently. Interesting. But LGTM in general.

Issue #452, opened 2022-06-05
Add support for Linux SAS classes


scottlaird:
Linux exports ~5 useful metrics per SAS PHY, which can be useful for debugging SAS performance problems and errors.  I'm planning on adding support for this to node_exporter (see https://github.com/prometheus/node_exporter/issues/2386), but I need to extract information from /sys first before I can do this.

The useful metrics are all in `/sys/class/sas_phy`, but we need data from `/sys/class/sas_end_device`, `/sys/class/sas_port`, and `/sys/class/sas_expander` in order to make sense of the PHY data and know what's connected to a specific PHY.

scottlaird:
Ah, I think I can close this one, but the node exporter issue is still open.  I have the code for that, but I never put a PR together.  I'll work on it soon.

PR #453, opened 2022-06-05
Add support for several /sys/class/sas_* classes

scottlaird:
This adds support for several SAS classes from `/sys/class`:

  * `sas_host`
  * `sas_device`
  * `sas_end_device`
  * `sas_expander`
  * `sas_phy`
  * `sas_port`

These are for issue #452, which is needed for https://github.com/prometheus/node_exporter/issues/2386.  Once that is complete, it'll be possible to track SAS communication errors on each individual SAS link in a system, and then map those links to specific block devices.

Use case: I recently discovered that I had a bad SAS cable in my system and was seeing weird, intermittent disk timeouts and errors which caused ZFS problems.  Being able to track SAS issues for each node should make detecting and debugging this sort of problem much easier.

### DIFF
--- a/sysfs/class_sas_device.go
+++ b/sysfs/class_sas_device.go
@@ -32,6 +32,7 @@ type SASDevice struct {
 	SASAddress string   // /sys/class/sas_device/<Name>/sas_address
 	SASPhys    []string // /sys/class/sas_device/<Name>/device/phy-*
 	SASPorts   []string // /sys/class/sas_device/<Name>/device/ports-*
+	BlockDevices []string  // /sys/class/sas_device/<Name>/device/target*/*/block/*
 }
 
 type SASDeviceClass map[string]SASDevice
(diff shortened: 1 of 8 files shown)


sysfs/class_sas_device.go
```diff
@@ -0,0 +1,174 @@
...
+package sysfs
+
+import (
+	"github.com/prometheus/procfs/internal/util"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+)
+
+const sasDeviceClassPath = "class/sas_device"
```
discordianfish, line 26: nit: turn this into a const block

sysfs/class_sas_device.go
```diff
@@ -0,0 +1,174 @@
...
+	device := SASDevice{Name: name}
+
+	devicepath := fs.sys.Path(filepath.Join(sasDeviceClassPath, name, "device"))
+
+	dirs, err := ioutil.ReadDir(devicepath)
+	if err != nil {
+		return nil, err
+	}
+
+	phyDevice := regexp.MustCompile(`^phy-[0-9:]+$`)
```
discordianfish, line 96: Probably makes sense to make this a global var to avoid recompiling the regex

sysfs/class_sas_host.go
```diff
@@ -0,0 +1,91 @@
...
+	host := SASHost{Name: name}
+
+	devicepath := fs.sys.Path(filepath.Join(sasHostClassPath, name, "device"))
+
+	dirs, err := ioutil.ReadDir(devicepath)
+	if err != nil {
+		return nil, err
+	}
+
+	phyDevice := regexp.MustCompile(`^phy-[0-9:]+$`)
```
discordianfish, line 78: Same

discordianfish:
Some minor comments but LGTM in general. Thanks!

sysfs/class_sas_device.go
scottlaird: Done

sysfs/class_sas_device.go
scottlaird: Done (and elsewhere).  I debated moving all of the regexes to either class_sas_head.go or their own file (due to global naming conflicts), but ended up removing duplicates and mostly leaving them in the file where they're first used.

scottlaird:
I started writing my `node_exporter` code, and found a couple bugs, which I just fixed.  There are also a couple convenience functions that should really be in here.  I may or may not have time for them today.

scottlaird:
Okay, that ended up being bigger than expected.  OTOH, the `node_exporter` code shrank quite a bit, especially the bits where it has to chain through multiple classes to find some random bit of data.  Let me know if there's a different pattern that you'd like to see or if there are any other changes that you'd suggest.  Thanks.

discordianfish (approved):
LGTM, thanks!

scottlaird:
@SuperQ any chance this could get a review?

SuperQ (approved):
LGTM

PR #460, opened 2022-07-14
Add sysctl

discordianfish:


ttar
```diff
@@ -30,15 +30,15 @@ path=""
 CMD=""
 ARG_STRING="$*"
 
+PYTHON=${PYTHON:-python}
```
SuperQ, line 33: Hmm, don't we take this ttar from some upstream?

SuperQ (approved):
LGTM

PR #461, opened 2022-07-17
add processor field for /proc/[pid]/stat

xixiliguo:
When analysing performance issue, it is useful to know process run on which cpu.

Signed-off-by: peter wang wangbo.red@gmail.com


discordianfish (approved):
Makes sense, thanks!

SuperQ (approved):
Thanks

PR #467, opened 2022-09-24
Update codespell

SuperQ:
Switch to ignore file, update with new ignores.


discordianfish (approved):
Thanks!

PR #469, opened 2022-10-07
fix: Avoid trimming characters incorrectly from ProcStatus

irvinlim:
Hello, thank you for the wonderful library!

We found some issues in parsing ProcStatus, in particular those whose `Name` starts with a `k`. For a simple repro (I copied out the existing code from https://github.com/prometheus/procfs/blob/bf361025088d9df7a76662269f07facf20f8711e/proc_status.go):

```go
package main

import (
	"bytes"
	"fmt"
	"strconv"
	"strings"

	"github.com/prometheus/procfs"
)

const (
	pid  = 6187
	data = `Name:   kube-proxy`
)

func main() {
	s := procfs.ProcStatus{PID: pid}

	lines := strings.Split(data, "\n")
	for _, line := range lines {
		if !bytes.Contains([]byte(line), []byte(":")) {
			continue
		}

		kv := strings.SplitN(line, ":", 2)

		// removes spaces
		k := string(strings.TrimSpace(kv[0]))
		v := string(strings.TrimSpace(kv[1]))
		// removes "kB"
		v = string(bytes.Trim([]byte(v), " kB"))

		// value to int when possible
		// we can skip error check here, 'cause vKBytes is not used when value is a string
		vKBytes, _ := strconv.ParseUint(v, 10, 64)
		// convert kB to B
		vBytes := vKBytes * 1024

		fillStatus(&s, k, v, vKBytes, vBytes)
	}

	fmt.Printf("%v\n", s.Name)
}
```

The output is as follows:

```
ube-proxy
```

The reason is because `bytes.Trim` takes in a cutSet, not a suffix that should be trimmed. As such, this PR modifies `bytes.Trim` to `strings.TrimSuffix`.

### DIFF
--- a/proc_status.go
+++ b/proc_status.go
@@ -96,10 +96,10 @@ func (p Proc) NewStatus() (ProcStatus, error) {
 		kv := strings.SplitN(line, ":", 2)
 
 		// removes spaces
-		k := string(strings.TrimSpace(kv[0]))
-		v := string(strings.TrimSpace(kv[1]))
+		k := strings.TrimSpace(kv[0])
+		v := strings.TrimSpace(kv[1])
 		// removes "kB"
-		v = string(bytes.Trim([]byte(v), " kB"))
+		v = strings.TrimSuffix(v, " kB")
 
 		// value to int when possible
 		// we can skip error check here, 'cause vKBytes is not used when value is a string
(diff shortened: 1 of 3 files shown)


discordianfish (approved):
Good catch, thanks!

SuperQ (approved):
Thanks!

PR #472, opened 2022-10-13
Separate NetStat parsing logic

Dentrax:
As procfs already did use `parseX(io.Reader)` functions for other filesystem formats, this was missing in `netstat` parsing. It should not do both _reading_ and _parsing_ stuff in the same function. Anyone who want to just access parser func using `go:linkname` directive, now it's possible. I had to copy-paste the actual parser logic since the function itself also _reads_ a file.

### DIFF
--- a/netstat.go
+++ b/netstat.go
@@ -15,6 +15,7 @@ package procfs
 
 import (
 	"bufio"
+	"io"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


discordianfish (approved):
Makes sense in general but neither functions are exposed, so they can't be used by consumers of the package anyway. We might want to consider exposing them. Do you have a use case for this?
But otherwise just for consistency alone this makes sense, so LGTM

SuperQ (approved):
Yes, let's make the parsing functions public. That would be very nice.

PR #474, opened 2022-10-29
Separate /proc/stat parsing logic

Dentrax:

Same motivation as PR https://github.com/prometheus/procfs/pull/472


discordianfish (approved):
Makes sense, thanks! LGTM.

SuperQ (approved):
Also same, let's make the parsing public.

PR #476, opened 2022-11-02
Add `/proc/modules` support

Dentrax:

I have some use-cases for this implementation:

* want to parse my `/proc/interrupts` file using procfs package (with go:linkname)
* move module parsing logic [from node-problem-detector](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/util/metrics/system/module_stats.go) to here

All tests passes.


proc_modules.go
```diff
@@ -0,0 +1,105 @@
...
+		parts := strings.Fields(scanner.Text())
+
+		module := Module{
+			Name:         parts[0],
+			Dependencies: []string{},
+			State:        parts[4],
+			Taints:       []string{},
+		}
+
+		if size, err := strconv.ParseUint(parts[1], 10, 64); err == nil {
```
discordianfish, line 78: Is it expected that the parsing might fail? If not, we should just abort here and return the error then silentely ignoring it.

proc_modules.go
```diff
@@ -0,0 +1,105 @@
...
+// - https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/4/html/reference_guide/s2-proc-modules
+// - https://unix.stackexchange.com/a/152527/300614
+func parseModules(r io.Reader) ([]Module, error) {
+	var (
+		scanner = bufio.NewScanner(r)
+		modules = make([]Module, 0)
+	)
+
+	for scanner.Scan() {
+		parts := strings.Fields(scanner.Text())
```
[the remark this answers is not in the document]
Dentrax: Done!

proc_modules.go
Dentrax: > Is it expected that the parsing might fail?

Yes, I think it's possible. Checking the error could also prevent from spec changes. (too low possibility, but still would be better than not checking instead.)

proc_modules.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, err
+	}
+	return parseModules(bytes.NewReader(data))
+}
+
+// parseModules parses the metrics from /proc/modules file
+// and returns a []Module structure.
+// - https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/4/html/reference_guide/s2-proc-modules
+// - https://unix.stackexchange.com/a/152527/300614
+func parseModules(r io.Reader) ([]Module, error) {
```
SuperQ, line 63: CC @discordianfish 

Should we make this public?

SuperQ (changes requested):

proc_modules.go
discordianfish: Yes, lets make it public

Dentrax: I've created another issue about keeping these _parse_ functions to _public_ instead. Let's discuss the idea here: https://github.com/prometheus/procfs/issues/478

I'm marking this one as resolved. Feel free to reopen. I can make it public. Other reviews are done, thanks!

proc_modules.go
discordianfish: yeah but I don't like that we silentely ignore the error here. Why not return an error if err != nil?

proc_modules_test.go
```diff
@@ -0,0 +1,97 @@
...
+// limitations under the License.
+
+package procfs
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestProcInterrupts(t *testing.T) {
```
discordianfish, line 21: Copy paste error I guess. Should be TestProcModules?

Dentrax: Oh, my mad. Done!

proc_modules.go
Dentrax: > Why not return an error if err != nil?

That's because I noticed this `err == nil` pattern in some places. For example, arp: https://github.com/prometheus/procfs/blob/0c4b3aa46606db022c89b9653f384ca8eefadb7e/arp.go#L98 zoneinfo: https://github.com/prometheus/procfs/blob/0c4b3aa46606db022c89b9653f384ca8eefadb7e/zoneinfo.go#L186

I agree with you, returning an error would better. I can return an error if you want. Wdyt?

discordianfish: @SuperQ wdyt? But yeah I think returning an error would be better

SuperQ: Yes, returning the error would be better.

SuperQ (changes requested):

rexagod:
@Dentrax Just a nit, but can you add `Fixes #478` to the description so that the original issue is closed when this gets in?

proc_modules.go
discordianfish: @Dentrax Can you change this so we can merge it?

PR #477, opened 2022-11-04
Separate /proc/stat disk stat parsing logic

Dentrax:

Same motivation as PR https://github.com/prometheus/procfs/pull/472


SuperQ (approved):
Same here, let's make these public.

PR #483, opened 2022-12-22
Use errors.Is() for invalid argument

SuperQ:
Switch to using Go errors.Is() to check for the "invalid argument" error.

### DIFF
--- a/sysfs/class_fibrechannel.go
+++ b/sysfs/class_fibrechannel.go
@@ -17,6 +17,7 @@
 package sysfs
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


discordianfish (approved):
LGTM

dcbw:
@SuperQ @discordianfish turns out this PR now prematurely terminates parsing on at least some systems (RHEL, Fedora) because the "invalid argument" error is actually a `syscall.Errno` EINVAL because SysReadFile() uses syscall.Read() instead of os.ReadFile() or os.Read().

Should I do a patch with a function that checks errors.Is(err, syscall.Errno) and compares to EINVAL, or should we revert?

PR #487, opened 2023-01-11
add avgRTT to nfs mountstats

ksankeerth:
Hi Team,
I'm sending this PR to add avgRTT to nfs mount stats. https://github.com/prometheus/node_exporter/issues/2550

I followed the same approach as linux kernel to calculate `avgRTT`. https://git.linux-nfs.org/?p=bfields/nfs-utils.git;a=blob;f=tools/nfs-iostat/nfs-iostat.py;h=1df74ba822b59f68c556530e6bf825a1e57f611d;hb=HEAD#l342

Let me know your feedback.

cc : @discordianfish 

Thanks!

### DIFF
--- a/mountstats.go
+++ b/mountstats.go
@@ -186,6 +186,8 @@ type NFSOperationStats struct {
 	CumulativeTotalResponseMilliseconds uint64
 	// Duration from when a request was enqueued to when it was completely handled.
 	CumulativeTotalRequestMilliseconds uint64
+	// The average time from the point the client sends RPC requests until it receives the response.
+	AverageRTTMilliseconds float64
 	// The count of operations that complete with tk_status < 0.  These statuses usually indicate error conditions.
 	Errors uint64
 }
@@ -534,7 +536,6 @@ func parseNFSOperationStats(s *bufio.Scanner) ([]NFSOperationStats, error) {
 
 			ns = append(ns, n)
 		}
-
 		opStats := NFSOperationStats{
 			Operation:                           strings.TrimSuffix(ss[0], ":"),
 			Requests:                            ns[0],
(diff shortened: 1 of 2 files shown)


mountstats.go
```diff
@@ -534,17 +536,33 @@ func parseNFSOperationStats(s *bufio.Scanner) ([]NFSOperationStats, error) {
...
-			Requests:                            ns[0],
-			Transmissions:                       ns[1],
-			MajorTimeouts:                       ns[2],
-			BytesSent:                           ns[3],
-			BytesReceived:                       ns[4],
-			CumulativeQueueMilliseconds:         ns[5],
-			CumulativeTotalResponseMilliseconds: ns[6],
-			CumulativeTotalRequestMilliseconds:  ns[7],
+		var opStats NFSOperationStats
+		if ns[0] != 0 {
```
discordianfish, line 540: Instead of this assign opStats without AverageRTT, then check if ns[0] != 0 and add AverageRTT

ksankeerth: thanks for review. I made the changes.

discordianfish:
DCO sign-off missing but LGTM otherwise

ksankeerth:
> DCO sign-off missing but LGTM otherwise

Thanks for reviewing again. Fixed the DCO issue and renamed `avgRTT` to `avgRTTMilliSeconds`

discordianfish (approved):

ksankeerth:
Hi @SuperQ,

Could you please review this PR when you have time?

Thanks in advance!

SuperQ (approved):
We should cleanup the Go docs for the struct, but since the rest are broken this is fine.

SuperQ:
I just realized that this feature is not something we should have in the code. It's computing an artificial value from existing data, which is not something we do in our libraries.

If you want to compute this value, you can do it upstream.

PR #489, opened 2023-01-20
Add support `/sys/devices/system/node/nodeN/meminfo`

khunafin:


sysfs/meminfo_numa.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2019 The Prometheus Authors
```
discordianfish, line 1: Update year

discordianfish (approved):
Great, LGTM!

sysfs/meminfo_numa.go
khunafin: done

SuperQ (changes requested):
Needs to fix the Go linter errors.

tjhop:
I stumbled across this PR while investigating for #565. If the approach used in #569 is accepted, it would seem like a similar approach may be needed here as well

PiotrProkop:
@khunafin are you planning to update this PR? If not I am willing to help with this work.

PR #490, opened 2023-01-25
add support CpusAllowedList for /proc/[pid]/status

Vatson112:
Hi, 

It may be usefull to collect CPU affinity info about processes. 

For example, we can monitor:

- Is out processes running on the same group of cores
- Is there too many proceeses running on the same group of cores

Tests passes.

### DIFF
--- a/proc_status.go
+++ b/proc_status.go
@@ -78,7 +78,7 @@ type ProcStatus struct {
 	// GIDs of the process (Real, effective, saved set, and filesystem GIDs)
 	GIDs [4]string
 
-	// CPUSET: list of cpu cores, allowed to run process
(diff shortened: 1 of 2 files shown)


discordianfish (approved):
LGTM, thanks!

proc_status.go
```diff
@@ -76,6 +77,9 @@ type ProcStatus struct {
 	UIDs [4]string
 	// GIDs of the process (Real, effective, saved set, and filesystem GIDs)
 	GIDs [4]string
+
+	// CPUSET: list of cpu cores, allowed to run process
```
SuperQ, line 81: Minor nit Go documentation comments need to match.

```suggestion
	// CpusAllowedList: List of cpu cores processes are allowed to run on.
```

PR #494, opened 2023-02-23
Fix parseVMStatNUMA return wrong VMStat in nr_zone_inactive_file

gaius-qi:
Hi, @discordianfish✋!
Thanks for maintaining this repository!


SuperQ (approved):
Oops! Thanks for the fix.

Issue #499, opened 2023-03-14
Failed to retrieve conntrack stats on an armv5 based Synology NAS


bruzzz:
### Host operating system: output of uname -a

Linux DS 2.6.32.12 #25556 Thu Jul 1 14:25:48 CST 2021 armv5tel GNU/Linux synology_88f6282_411

### node_exporter version: output of node_exporter --version

node_exporter, version 1.5.0 (branch: HEAD, revision: https://github.com/prometheus/node_exporter/commit/1b48970ffcf5630534fb00bb0687d73c66d1c959)
build user: root@6e7732a7b81b
build date: 20221129-19:00:13
go version: go1.19.3
platform: linux/arm

As originally reported in [this ticket](https://github.com/prometheus/node_exporter/issues/2635), the conntrack collector on the node_exporter seems to have an issue because of some expected fields beeing missing.

### The error meassage with node_exporter v 1.2.2 looks as following (node_exporter log output):
`ts=2023-03-13T19:04:46.789Z caller=collector.go:169 level=error msg="collector failed" name=conntrack duration_seconds=0.000592692 err="failed to retrieve conntrack stats: failed to read conntrack stats from \"/proc/net/stat/nf_conntrack\": invalid conntrackstat entry, missing fields"`

### With version 1.5.0 it seems to be even worse:
```
ts=2023-03-13T21:26:20.985Z caller=collector.go:169 level=error msg="collector failed" name=conntrack duration_seconds=0.000755877 err="failed to retrieve conntrack stats: failed to read conntrack stats from \"/proc/net/stat/nf_conntrack\": invalid conntrackstat entry, missing fields"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x66de9c]

goroutine 59 [running]:
github.com/prometheus/node_exporter/collector.parseNetlinkStats({0x1209ec0, 0x3, 0x3}, 0x105dda8, {0x8be790, 0x105dd70})
        /home/dietpi/node_exporter/node_exporter-1.5.0/collector/netdev_linux.go:69 +0x1dc
github.com/prometheus/node_exporter/collector.netlinkStats(0x105dda8, {0x8be790, 0x105dd70})
        /home/dietpi/node_exporter/node_exporter-1.5.0/collector/netdev_linux.go:52 +0x114
[...]

PR #500, opened 2023-03-15
Support old conntrack stats

SuperQ:
Linux < 2.6.35 was missing the `search_restart` field in nf_conntrack.
* Refactor stat parsing to simplify code.
* Support 16-field nf_conntrack entries.

Fixes: https://github.com/prometheus/procfs/issues/499


internal/util/parse.go
```diff
@@ -64,6 +64,21 @@ func ParsePInt64s(ss []string) ([]*int64, error) {
 	return us, nil
 }
 
+// Parses a uint64 from given hex in string.
+func ParseHexUint64s(ss []string) ([]*uint64, error) {
+	us := make([]*uint64, 0, len(ss))
```
pgier, line 69: Could this return a `uint64` slice instead of `*uint64`?

pgier (approved):
LGTM, just left one question inline.

internal/util/parse.go
SuperQ: It could, I was following the pattern of the other string parsers here.

discordianfish (approved):

Issue #502, opened 2023-04-01
Feature Request: support loading net.ipv4 config


Jack-Kingdom:

SuperQ:
I think this would be more appropriate for a dedicated library package.

For example: https://github.com/lorenzosaino/go-sysctl

PR #504, opened 2023-04-03
feat: add LoadSysConfig method for read sys config easier #502

Jack-Kingdom:


SuperQ (changes requested):
I don't see what this is useful for. It appears to be a trivial wrapper around ReadFileNoStat. This seems somewhat useless as it's just unwrapping and then re-wrapping the path string.

Jack-Kingdom:
@SuperQ yeah, you know there is a batch of sysctl config located under procfs.
Under sysctl config file, They looks like this:
```
# enable fast open
net.ipv4.tcp_fastopen = 3
# turn on path MTU discovery
net.ipv4.tcp_mtu_probing = 1
# tcp window scaling
net.ipv4.tcp_window_scaling = 1
```

and under procfs, it's saparated files located at different path.
for example:
```
/proc/sys/net/ipv4/tcp_fastopen
```

at first, I want to provide a struct and read those config into struct. but thosed config very scattered and commonly a file only contains a single config. so I created this helper method to load those config. Is there a better way ?

SuperQ:
Oh, you want to access sysctls. I would recommend using a dedicated package like https://github.com/lorenzosaino/go-sysctl.

SuperQ:
I don't think we need a sysctl library here.

PR #505, opened 2023-04-13
Support parse raid type for mdstat

ImSingee:
For something like

```text
md6 : active raid1 sdb2[2](F) sdc[1](S) sda2[0]
md11 : active (auto-read-only) raid1 sdb2[0] sdc2[1] sdc3[2](F) hda[4](S) ssdc2[3](S)
```

We can also get the raid type from /proc/mdstat

### DIFF
--- a/mdstat.go
+++ b/mdstat.go
@@ -97,11 +97,14 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
 		mdName := deviceFields[0] // mdx
 		state := deviceFields[2]  // active or inactive
 
-		mdType := "unknown"
-		if len(deviceFields) > 3 && isRaidType(deviceFields[3]) {
-			mdType = deviceFields[3] // raid1, raid5, etc.
-		} else if len(deviceFields) > 4 && isRaidType(deviceFields[4]) {
-			mdType = deviceFields[4]
+		mdType := "unknown"        // raid1, raid5, etc.
+		if len(deviceFields) > 3 { // mdType may be in the 3rd or 4th field
+			if isRaidType(deviceFields[3]) {
+				mdType = deviceFields[3]
+			} else if len(deviceFields) > 4 && isRaidType(deviceFields[4]) {
+				// if the 3rd field is (...), the 4th field is the mdType
+				mdType = deviceFields[4]
+			}
 		}
 
 		if len(lines) <= i+3 {
@@ -175,8 +178,11 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
 	return mdStats, nil
 }
 
+// check if a string's format is like the mdType
+// Rule 1: mdType should not be like (...)
+// Rule 2: mdType should not be like sda[0]
 func isRaidType(mdType string) bool {
-	return strings.HasPrefix(mdType, "raid") || mdType == "linear"
+	return !strings.ContainsAny(mdType, "([")
 }
 
 func evalStatusLine(deviceLine, statusLine string) (active, total, down, size int64, err error) {


SuperQ (changes requested):

mdstat_test.go
```diff
@@ -25,23 +25,23 @@ func TestFS_MDStat(t *testing.T) {
...
+		"md12":  {Name: "md12", Type: "raid0", ActivityState: "active", DisksActive: 2, DisksTotal: 2, DisksSpare: 0, DisksDown: 0, DisksFailed: 0, BlocksTotal: 3886394368, BlocksSynced: 3886394368, BlocksSyncedPct: 0, BlocksSyncedFinishTime: 0, BlocksSyncedSpeed: 0, Devices: []string{"sdc2", "sdd2"}},
+		"md120": {Name: "md120", Type: "linear", ActivityState: "active", DisksActive: 2, DisksTotal: 2, DisksFailed: 0, DisksDown: 0, DisksSpare: 0, BlocksTotal: 2095104, BlocksSynced: 2095104, BlocksSyncedPct: 0, BlocksSyncedFinishTime: 0, BlocksSyncedSpeed: 0, Devices: []string{"sda1", "sdb1"}},
+		"md126": {Name: "md126", Type: "raid0", ActivityState: "active", DisksActive: 2, DisksTotal: 2, DisksFailed: 0, DisksDown: 0, DisksSpare: 0, BlocksTotal: 1855870976, BlocksSynced: 1855870976, BlocksSyncedPct: 0, BlocksSyncedFinishTime: 0, BlocksSyncedSpeed: 0, Devices: []string{"sdb", "sdc"}},
+		"md219": {Name: "md219", Type: "", ActivityState: "inactive", DisksTotal: 0, DisksFailed: 0, DisksActive: 0, DisksDown: 0, DisksSpare: 3, BlocksTotal: 7932, BlocksSynced: 7932, BlocksSyncedPct: 0, BlocksSyncedFinishTime: 0, BlocksSyncedSpeed: 0, Devices: []string{"sdc", "sda"}},
```
SuperQ, line 41: We should probably handle inactive by setting the type to `unknown`.

mdstat.go
```diff
@@ -95,6 +97,13 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
 		mdName := deviceFields[0] // mdx
 		state := deviceFields[2]  // active or inactive
 
+		mdType := ""
```
SuperQ, line 100: Maybe set the default to `unknown`.

SuperQ (changes requested):

mdstat.go
```diff
@@ -95,6 +97,13 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
...
+		if len(deviceFields) > 3 && strings.HasPrefix(deviceFields[3], "raid") {
```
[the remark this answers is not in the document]
SuperQ: Yea, that's better. Parsing this file can be pretty tricky.

SuperQ:

mdstat.go
```diff
@@ -165,6 +175,10 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
 	return mdStats, nil
 }
 
+func isRaidType(mdType string) bool {
+	return strings.HasPrefix(mdType, "raid") || mdType == "linear"
```
SuperQ, line 179: No, I don't think this is going to work. Like I said, there are a whole bunch of other types other than `raid...`. Since this can change depending on the kernel/mdadm version. We can't support an exhaustive list here.

SuperQ (changes requested):
I think the original checking for `(...)` in the word after active/inactive was better.

mdstat.go
```diff
@@ -95,6 +97,13 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
 		mdName := deviceFields[0] // mdx
 		state := deviceFields[2]  // active or inactive
 
+		mdType := "unknown"
+		if len(deviceFields) > 3 && isRaidType(deviceFields[3]) {
```
SuperQ, line 101: If you move this to after `if len(lines) <= i+3 {` you can reduce the word checking to just the check for `(...)`.

SuperQ: Actually, maybe this whole piece should be moved to the `evalStatusLine()` function.

SuperQ (commented):
I think we only get a type if state is `active`. Maybe check that first before checking the next word or two.

SuperQ:
Ahh, crap, yea, that's a problem.

mdstat.go
```diff
@@ -165,6 +178,13 @@ func parseMDStat(mdStatData []byte) ([]MDStat, error) {
 	return mdStats, nil
 }
 
+// check if a string's format is like the mdType
+// Rule 1: mdType should not be like (...)
+// Rule 2: mdType should not be like sda[0]
+func isRaidType(mdType string) bool {
```
discordianfish, line 184: Ewww hah..
There must be a better way. Is there an explicit way to query the md type? E.g by reasing another procfs file?

robbat2:
@ImSingee  please see my PR https://github.com/prometheus/procfs/pull/674 that fixes the outstanding issues here; but your commit are missing some signed-off tags - could you please fix that in this PR and i'll rebase my PR on top of that.

ImSingee:
@robbat2 Sure, I will update this later this day!

ImSingee:
@robbat2 I have added the Sign-Off

robbat2:
> @robbat2 I have added the Sign-Off

Thank you; rebased my PR now

PR #509, opened 2023-04-24
Implement mdraid sysfs parsing

dswarbrick:
Modernised method of fetching mdraid statistics via machine-readable sysfs entries, instead of parsing human-readable `/proc/mdstat`.

This would go a long way towards addressing https://github.com/prometheus/node_exporter/issues/1085, as well as superseding / obsoleting various other pending PRs (#329) that attempt to squeeze a bit more info out of `/proc/mdstat` and issues, e.g., https://github.com/prometheus/node_exporter/issues/1874

The mdraid sysfs entries contain significantly more detailed information that what is presented in the simplified and human-readable `/proc/mdstat`.

I'd like to get a few more eyeballs on this code, and have it alongside the existing /proc/mdstat parser initially, with a view to perhaps making the `MDStat` function a wrapper around the new sysfs parser eventually (or refactor node_exporter to use the new function directly). Existing `MDStat` members ActivityState, DisksDown, DisksFailed, DisksSpare, DisksTotal, DisksActive can be derived / counted / calculated from information exposed by the new `Mdraid` struct.

_Some_ of the `MDStat` struct information is not exposed directly by `/sys/block/md*/md`, however it is obtainable via other means. For example:
- BlocksTotal - this is technically redundant, since it is just the md block device size in KiB. This should be implemented generically for _all_ block device types, i.e., by exposing `/sys/block/*/size` (measured in 512-byte sectors - yes, even on 4K sector drives).
- BlocksSynced - sysfs will only expose this _whilst an array is resyncing / recovering_ (or checking). Otherwise, it is to be assumed that all blocks are in sync.

Other `MDStat` struct members which are **not** currently used by node_exporter:
- BlocksSyncedPct - superseded by SyncCompleted, which is a fraction (0-1) representing the completion status of current sync operation.
- BlocksSyncedSpeed - no equivalent in sysfs, however this could be calculated in promql with the `rate()` function.
- BlocksSyncedFinishTime - no equivalent in sysfs, however this could be calculated in promql by dividing the remaining blocks to sync by the sync speed.

@SuperQ I would greatly appreciate your review / input on this.

### DIFF
--- a/blockdevice/stats_test.go
+++ b/blockdevice/stats_test.go
@@ -73,7 +73,7 @@ func TestBlockDevice(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expectedNumOfDevices := 2
+	expectedNumOfDevices := 8
 	if len(devices) != expectedNumOfDevices {
 		t.Fatalf(failMsgFormat, "Incorrect number of devices", expectedNumOfDevices, len(devices))
 	}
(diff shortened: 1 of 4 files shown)


SuperQ:
Nice, do we know how far back this kernel support goes?

discordianfish (approved):
Nice! LGMT. We can still support parsing both in the node-exporter.

SuperQ (approved):
Awesome.

PR #511, opened 2023-05-03
net class: split out attribute parser and export it

dcbw:
Can be used by other projects (eg node_exporter) to read a single sysfs file rather than all of them at once, which can cause lock contention in the kernel as many sysfs net files grab the RTNL lock.

Immediate consumer: https://github.com/prometheus/node_exporter/pull/2669

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -117,6 +117,77 @@ func (fs FS) NetClass() (NetClass, error) {
 	return netClass, nil
 }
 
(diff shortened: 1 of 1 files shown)


dcbw:
@discordianfish is this what you were thinking of?

discordianfish (approved):
Perfect, thanks!

SuperQ (approved):
Thanks

PR #512, opened 2023-05-04
static analyzer issue fix

artemseleznev:
issue was found by Svace (https://www.ispras.ru/en/technologies/svace/)


discordianfish:
Looks like this is a valid issue but using defer like this will keep all files open until the function returns. I'd either move this to a function that is called per file and uses defer or just wrap it in a anonymous function.

SuperQ:
Yea, refactoring the reading to a function would be a good idea.

PR #516, opened 2023-05-10
net_class: catch syscall.EINVAL and better handle read errors

dcbw:
https://github.com/prometheus/procfs/pull/483 introduced a bug that terminates attribute reading when the returned error is `syscall.Errno`, which is what `util.SysReadFile()` will typically return. Handle that case specifically.

While doing that pull the error checking into `ParseNetClassAttribute()` for clarity and to allow external callers to handle errors correctly.

@SuperQ @discordianfish

### DIFF
--- a/sysfs/net_class.go
+++ b/sysfs/net_class.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	"github.com/prometheus/procfs/internal/util"
 )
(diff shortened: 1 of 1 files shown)


sysfs/net_class.go
```diff
@@ -117,14 +118,35 @@ func (fs FS) NetClass() (NetClass, error) {
...
+// It returns an error if the file cannot be read and the error is fatal.
 func ParseNetClassAttribute(devicePath, attrName string, interfaceClass *NetClassIface) error {
 	attrPath := filepath.Join(devicePath, attrName)
 	value, err := util.SysReadFile(attrPath)
 	if err != nil {
-		return fmt.Errorf("failed to read file %q: %w", attrPath, err)
+		if isFatalError(err) {
+			return fmt.Errorf("failed to read file %q: %w", attrPath, err)
+		}
+		return nil
```
SuperQ, line 149: Hmm, what errors are we going to ignore by doing this?

SuperQ:
Maybe we should inverse this, all errors are fatal unless they're in a list of allowed non-fatal errors.

sysfs/net_class.go
dcbw: @SuperQ No more errors than we ignored before... It uses the same logic, but now additionally ignores the syscall.Errno(EINVAL)

dcbw:
> Maybe we should inverse this, all errors are fatal unless they're in a list of allowed non-fatal errors.

@SuperQ that's effectively what the code is already doing (and was doing before); but perhaps my function name is not as clear. I can reverse the meaning if you'd like.

dcbw:
@SuperQ flipped meaning to `canIgnoreError()` and updated the docstring to better explain. PTAL thanks!

Issue #358, opened 2021-02-09
unify error messages


roidelapluie:
currently we have:

could not parse
failed to parse
error parsing
failed to read file ...
fail to scan

We should unify the error messages.

alexal:
@roidelapluie I just created PR #360. Please check it out and let me know if I need to change something. Thanks.

conallob:
Is PR #360 still ongoing or is it stalled? If there are no objections, I'm interested in taking this on

SuperQ:
#360 is stalled, feel free to chery-pick and continue with it.

conallob:
@SuperQ @discordianfish this was implemented in https://github.com/prometheus/procfs/pull/526

Could somebody assign this issue to me and mark it as closed please?

PR #518, opened 2023-05-11
Unify error messages and prepare to adopt wrapping error messages, part 1

conallob:
Unify error messages by using the errors module. Prepare for wrapping errors once Go v1.20 is the default.

This is part 1 of an implementation for #358

### DIFF
--- a/net_softnet.go
+++ b/net_softnet.go
@@ -76,6 +76,7 @@ func parseSoftnet(r io.Reader) ([]SoftnetStat, error) {
 	s := bufio.NewScanner(r)
 
 	var stats []SoftnetStat
+	cpuIndex := 0
 	for s.Scan() {
 		columns := strings.Fields(s.Text())
 		width := len(columns)
@@ -127,9 +128,13 @@ func parseSoftnet(r io.Reader) ([]SoftnetStat, error) {
 
 			softnetStat.SoftnetBacklogLen = us[0]
 			softnetStat.Index = us[1]
+		} else {
+			// For older kernels, create the Index based on the scan line number.
+			softnetStat.Index = uint32(cpuIndex)
 		}
 		softnetStat.Width = width
 		stats = append(stats, softnetStat)
+		cpuIndex++
 	}
 
 	return stats, nil
--- a/net_softnet_test.go
+++ b/net_softnet_test.go
@@ -55,6 +55,7 @@ func TestNetSoftnet(t *testing.T) {
 			CPUCollision:   0x00000004,
 			ReceivedRps:    0x00000003,
 			FlowLimitCount: 0x00000002,
+			Index:          0x00000002,
 			Width:          11,
 		},
 		{
@@ -62,6 +63,7 @@ func TestNetSoftnet(t *testing.T) {
 			Dropped:      0x00000000,
 			TimeSqueezed: 0x0109a4,
 			CPUCollision: 0x00020e76,
+			Index:        0x00000003,
 			Width:        9,
 		},
 		{
@@ -70,6 +72,7 @@ func TestNetSoftnet(t *testing.T) {
 			TimeSqueezed: 0x00000011,
 			CPUCollision: 0x00000020,
 			ReceivedRps:  0x00000010,
+			Index:        0x00000004,
 			Width:        10,
 		},
 	}


SuperQ:

loadavg.go
```diff
@@ -44,14 +44,14 @@ func parseLoad(loadavgBytes []byte) (*LoadAvg, error) {
 	loads := make([]float64, 3)
 	parts := strings.Fields(string(loadavgBytes))
 	if len(parts) < 3 {
-		return nil, fmt.Errorf("malformed loadavg line: too few fields in loadavg string: %q", string(loadavgBytes))
+		return nil, fmt.Errorf("%w: Malformed line %q: %w", ErrFileParse, string(loadavgBytes), err)
```
SuperQ, line 47:
```suggestion
		return nil, fmt.Errorf("%w: Malformed line, too few fields in loadavg string: %q", ErrFileParse, string(loadavgBytes))
```

mdstat.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Prometheus Authors
+// Copyright 2018 The Prometheur Authors
```
SuperQ, line 1: derp

conallob (commented):
I now believe the test failures are related to the golang version of the test environment.

Per https://go.dev/doc/go1.20#errors , multiple %w atoms requires Go >v1.20.x

SuperQ:
Yes, I guess we need to either string some errors, or wait a bit. We typically try and support the currently supported Go versions. So we need to stay compatible with 1.19 for now.

conallob:
I could adapt this PR to use https://github.com/hashicorp/go-multierror instead and file a new issue to migrate to the errors package once Go v1.20 is the default

SuperQ:
No, I think just using some `%s` is fine for now. For cases where there is multi-error, prefer to pass the underlying error, not the new error type.

conallob:
> No, I think just using some `%s` is fine for now. For cases where there is multi-error, prefer to pass the underlying error, not the new error type.

Done.


I've also filed https://github.com/prometheus/procfs/issues/519 to track replacing these `%s` with `%w` again, once Go v1.20 is the default

fscache.go
```diff
@@ -263,7 +263,7 @@ func parseFscacheinfo(r io.Reader) (*Fscacheinfo, error) {
 	for s.Scan() {
 		fields := strings.Fields(s.Text())
 		if len(fields) < 2 {
-			return nil, fmt.Errorf("malformed Fscacheinfo line: %q", s.Text())
+			return nil, fmt.Errorf("%w: %q", ErrFileParse, s.Text())
```
SuperQ, line 266: I wonder if we should keep some of these details.

```suggestion
			return nil, fmt.Errorf("%w: malformed Fscacheinfo line: %q", ErrFileParse, s.Text())
```

SuperQ (approved):

conallob:
https://github.com/prometheus/procfs/pull/522 was intended to be a complimentary PR, but I managed to goof the git branches, so all changes are now in this PR


PTAL

arp.go
```diff
@@ -78,11 +78,11 @@ func parseARPEntries(data []byte) ([]ARPEntry, error) {
 		} else if width == expectedDataWidth {
 			entry, err := parseARPEntry(columns)
 			if err != nil {
-				return []ARPEntry{}, fmt.Errorf("failed to parse ARP entry: %w", err)
+				return []ARPEntry{}, fmt.Errorf("%s: %v: %w", ErrFileParse, entry, err)
```
SuperQ, line 81: Keep original details.

buddyinfo.go
```diff
@@ -55,7 +55,7 @@ func parseBuddyInfo(r io.Reader) ([]BuddyInfo, error) {
 		parts := strings.Fields(line)
 
 		if len(parts) < 4 {
-			return nil, fmt.Errorf("invalid number of fields when parsing buddyinfo")
+			return nil, fmt.Errorf("%w: %v", ErrFileParse, parts)
```
SuperQ, line 58: Keep original details.

buddyinfo.go
```diff
@@ -66,15 +66,15 @@ func parseBuddyInfo(r io.Reader) ([]BuddyInfo, error) {
...
+				return nil, fmt.Errorf("%w: mismatch in number of buddyinfo buckets, previous count %d, new count %d", ErrFileParse, bucketCount, arraySize)
 			}
 		}
 
 		sizes := make([]float64, arraySize)
 		for i := 0; i < arraySize; i++ {
 			sizes[i], err = strconv.ParseFloat(parts[i+4], 64)
 			if err != nil {
-				return nil, fmt.Errorf("invalid value in buddyinfo: %w", err)
+				return nil, fmt.Errorf("%s: %f: %w", ErrFileParse, sizes[i], err)
```
SuperQ, line 77: Keep original details.

crypto.go
```diff
@@ -83,7 +84,7 @@ func parseCrypto(r io.Reader) ([]Crypto, error) {
 
 		kv := strings.Split(text, ":")
 		if len(kv) != 2 {
-			return nil, fmt.Errorf("malformed crypto line: %q", text)
+			return nil, fmt.Errorf("%w: %q", ErrFileParse, text)
```
SuperQ, line 87: Keep original details.

fscache.go
```diff
@@ -245,7 +245,7 @@ func (fs FS) Fscacheinfo() (Fscacheinfo, error) {
 func setFSCacheFields(fields []string, setFields ...*uint64) error {
 	var err error
 	if len(fields) < len(setFields) {
-		return fmt.Errorf("Insufficient number of fields, expected %v, got %v", len(setFields), len(fields))
+		return fmt.Errorf("%s: Expected %d, but got %d: %w", ErrFileParse, len(setFields), len(fields), err)
```
SuperQ, line 248: Keep original details.

loadavg.go
```diff
@@ -44,14 +44,14 @@ func parseLoad(loadavgBytes []byte) (*LoadAvg, error) {
...
-		return nil, fmt.Errorf("malformed loadavg line: too few fields in loadavg string: %q", string(loadavgBytes))
+		return nil, fmt.Errorf("%w: Malformed line %q", ErrFileParse, string(loadavgBytes))
 	}
 
 	var err error
 	for i, load := range parts[0:3] {
 		loads[i], err = strconv.ParseFloat(load, 64)
 		if err != nil {
-			return nil, fmt.Errorf("could not parse load %q: %w", load, err)
+			return nil, fmt.Errorf("%s: %f: %w", ErrFileParse, loads[i], err)
```
SuperQ, line 54: Keep original details.

mdstat.go
```diff
@@ -70,7 +70,7 @@ func (fs FS) MDStat() ([]MDStat, error) {
 	}
 	mdstat, err := parseMDStat(data)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing mdstat %q: %w", fs.proc.Path("mdstat"), err)
+		return nil, fmt.Errorf("%s: %v: %w", ErrFileParse, fs.proc.Path("mdstat"), err)
```
SuperQ, line 73: Keep original details.

SuperQ (changes requested):
There are still a lot of errors that drop the original context of what part of parsing failed. Please make sure that we keep the original details of the parsing error intact.

I've highlighted a number of them.

conallob:
Thanks, I have been fixing up entries that didn't give specifics, but I may have missed a few early on

conallob:
Fixed up all the vague, non specific errors (`grep "fmt.Errorf" *.go | grep -e "%s: %.: %w"` confirms they're all gone)

I do have 2 commits which aren't DCO signed (should work out how to config that per repository) and I'm not sure how to amend those commits

conallob:
Recreated as https://github.com/prometheus/procfs/pull/526 , but without all the messed up git history

Issue #519, opened 2023-05-15
Use Go 1.20+ Error Wrapping


conallob:
Follow on from https://github.com/prometheus/procfs/issues/358

https://go.dev/doc/go1.20#errors introduces the ability to wrap multiple errors using multiple `%w` atoms in `fmt.Errorf()`

Once Go v1.20 is the default version, `fmt.Errorf()` entries should be updated to use multiple `%w` atoms. Thanks to https://github.com/prometheus/procfs/issues/358 , this should be replacing `%s` for `%w`

PR #526, opened 2023-05-23
Create new error types in proc.go and use them to unify error messages

conallob:
Start using new errors to unify error messages, complete with groundwork for wrapping errors once Go v1.20+ is the default


SuperQ (approved):

conallob:
https://github.com/prometheus/procfs/issues/519 already exists to track wrapping errors, once Go v1.20 is the default version

Issue #527, opened 2023-05-28
Release v0.10.0 does not build with GOOS=windows + GOARCH=amd64


adrianosela:
## Release v0.10.0 does not build with GOOS=windows + GOARCH=amd64

Packages importing v0.10.0 of this module (github.com/prometheus/procfs) cannot build for amd64 windows.

Issue likely introduced via https://github.com/prometheus/procfs/commit/80824fa6b8fa49fbc0b58cf6a2271ec78777d595 and solved for certain operating systems via https://github.com/prometheus/procfs/pull/523 (but not windows!)

### Example

```
10:41 $ GOOS=windows GOARCH=amd64 go build .
# github.com/prometheus/procfs
../../../../pkg/mod/github.com/prometheus/procfs@v0.10.0/fs_statfs_type.go:25:18: undefined: syscall.Statfs_t
../../../../pkg/mod/github.com/prometheus/procfs@v0.10.0/fs_statfs_type.go:26:17: undefined: syscall.Statfs
```

Release v0.10.0 is a breaking change. Kindly look into this.

SuperQ:
Released as https://github.com/prometheus/procfs/releases/tag/v0.10.1

PR #523, opened 2023-05-17
fs: Statfs_t{} doesn't have a Type field on NetBSD, OpenBSD, or Solaris

dcbw:
Other projects that import procfs, like node_exporter, apparently build for OSs that don't have Statfs_t.Type.

Leads to errors like (for node_exporter):

```
/go/bin/promu --config .promu.yml build --prefix .build/openbsd-amd64 
 >   node_exporter
# github.com/prometheus/procfs
/go/pkg/mod/github.com/prometheus/procfs@v0.9.1-0.20230517120836-7acd99d0499a/fs.go:63:14: stat.Type undefined (type syscall.Statfs_t has no field or method Type)
!! command failed: build -o .build/openbsd-amd64/node_exporter -ldflags -X github.com/prometheus/common/version.Version=1.5.0 -X github.com/prometheus/common/version.Revision=849b6076b81bf64249b4f75e0c3134d25a1da9a3 -X github.com/prometheus/common/version.Branch=pull/2669 -X github.com/prometheus/common/version.BuildUser=root@d9577d73a62f -X github.com/prometheus/common/version.BuildDate=20230517-20:49:53  -extldflags '-static' -a -tags 'netgo osusergo static_build' github.com/prometheus/node_exporter: exit status 1
```

PR #528, opened 2023-05-28
fs: syscall.Statfs_t object doesn't have a Type field on Windows

adrianosela:
## fs: syscall.Statfs_t object doesn't have a Type field on Windows

Fixes build for `GOOS=windows` (https://github.com/prometheus/procfs/issues/527)

This PR extends the solution from https://github.com/prometheus/procfs/pull/523 to include windows in the list of OSes for which the `syscall.Statfs_t` object does not have a `Type` field

### Before:

```
✔ ~/go/src/github.com/adrianosela/procfs [master|✔]
11:22 $ GOOS=windows GOARCH=amd64 go build
./fs_statfs_type.go:25:18: undefined: syscall.Statfs_t
./fs_statfs_type.go:26:17: undefined: syscall.Statfs
```

### After:

```
✔ ~/go/src/github.com/adrianosela/procfs [master|✔]
11:24 $ GOOS=windows GOARCH=amd64 go build
✔ ~/go/src/github.com/adrianosela/procfs [master|✔]
```

### DIFF
--- a/fs_statfs_notype.go
+++ b/fs_statfs_notype.go
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build netbsd || openbsd || solaris
-// +build netbsd openbsd solaris
+//go:build netbsd || openbsd || solaris || windows
+// +build netbsd openbsd solaris windows
 
 package procfs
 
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
Oops, thanks!

adrianosela:
Cross posting from: https://github.com/prometheus/procfs/pull/529

Hey @SuperQ , would be good to go over the list here https://github.com/golang/go/blob/master/src/go/build/syslist.go and deal with all of the OSes.... there's a list of unix-like OSses there too, so maybe worth just going off of that?

PR #529, opened 2023-05-28
fs: Statfs_t{} doesn't have a Type field on Windows (#528)

SuperQ:


adrianosela:
Hey @SuperQ , would be good to go over the list here https://github.com/golang/go/blob/master/src/go/build/syslist.go and deal with all of the OSes.... there's a list of unix-like OSses there too, so maybe worth just going off of that?

PR #532, opened 2023-06-02
Fix CPUVulnerabilities() reporting from sysfs

### LINKED ISSUE #531: CPU vulnerability reporting from sysfs is broken
I was working on: https://github.com/prometheus/node_exporter/issues/1046 and got a panic when trying to get cpu vulnerabilities:

```
panic: Couldn't create metrics handler: couldn't create collector: failed to get vulnerabilities: unknown vulnerability state for itlb_multihit: Not affected
```

Turns out that the parser is looking for `"Not Affected"` rather than `"Not affected"`.  I'm not sure where the capital letter came from. I don't see it in the sysfs code. I suspect it was copied from the comment in the relevant PR: https://github.com/prometheus/procfs/pull/106#issuecomment-414981407 and this actually never worked.

mwasilew2:
closes #531 

This PR contains a few changes (each in a separate commit):
- adds fixtures and tests for parsing cpu vulnerabilities from sysfs, fixes the typo in the vulnerability string
- switches to using a map of vulnerabilities instead of a slice (makes working with the result easier, e.g. when testing)
- switches to using encoding of the state
- adds license in the test file
- adds linux build flags
- adds test cases for vulnerable state, brings state string to lower cases (to try and make it case insensitive)

### DIFF
--- a/sysfs/vulnerability.go
+++ b/sysfs/vulnerability.go
@@ -81,29 +81,24 @@ func parseVulnerability(name, rawContent string) (*Vulnerability, error) {
 	v := &Vulnerability{CodeName: name}
 	rawContent = strings.TrimSpace(rawContent)
 	rawContentLower := strings.ToLower(rawContent)
-
-	if strings.HasPrefix(rawContentLower, notAffected) {
+	switch {
+	case strings.HasPrefix(rawContentLower, notAffected):
 		v.State = VulnerabilityStateNotAffected
-		return v, nil
-	}
-
-	if strings.HasPrefix(rawContentLower, vulnerable) {
+	case strings.HasPrefix(rawContentLower, vulnerable):
 		v.State = VulnerabilityStateVulnerable
 		m := strings.Fields(rawContent)
 		if len(m) > 1 {
 			v.Mitigation = strings.Join(m[1:], " ")
 		}
-		return v, nil
-	}
-
-	if strings.HasPrefix(rawContentLower, mitigation) {
+	case strings.HasPrefix(rawContentLower, mitigation):
 		v.State = VulnerabilityStateMitigation
 		m := strings.Fields(rawContent)
 		if len(m) > 1 {
 			v.Mitigation = strings.Join(m[1:], " ")
 		}
-		return v, nil
-	}
+	default:
+		return nil, fmt.Errorf("unknown vulnerability state for %s: %s", name, rawContent)
 
-	return nil, fmt.Errorf("unknown vulnerability state for %s: %s", name, rawContent)
+	}
+	return v, nil
 }
--- a/sysfs/vulnerability.go
+++ b/sysfs/vulnerability.go
@@ -36,6 +36,8 @@ const (
 )
 
 var (
+	// VulnerabilityHumanEncoding allows mapping the vulnerability state (encoded as an int) onto a human friendly
+	// string. It can be used by consumers of this library to expose to the user the state of the vulnerability.
 	VulnerabilityHumanEncoding = map[int]string{
 		VulnerabilityStateNotAffected: notAffected,
 		VulnerabilityStateVulnerable:  vulnerable,
--- a/sysfs/vulnerability.go
+++ b/sysfs/vulnerability.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	notAffected = "Not Affected"
+	notAffected = "Not affected"
 	vulnerable  = "Vulnerable"
 	mitigation  = "Mitigation"
 )
--- a/sysfs/vulnerability.go
+++ b/sysfs/vulnerability.go
@@ -24,33 +24,33 @@ import (
 )
 
 const (
-	notAffected = "Not affected"
+	notAffected = "Not affected" // based on: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu
 	vulnerable  = "Vulnerable"
 	mitigation  = "Mitigation"
 )
 
 // CPUVulnerabilities retrieves a map of vulnerability names to their mitigations.
-func (fs FS) CPUVulnerabilities() ([]Vulnerability, error) {
-	matches, err := filepath.Glob(fs.sys.Path("devices/system/cpu/vulnerabilities/*"))
+func (fs FS) CPUVulnerabilities() (map[string]*Vulnerability, error) {
+	matchingFilepaths, err := filepath.Glob(fs.sys.Path("devices/system/cpu/vulnerabilities/*"))
 	if err != nil {
 		return nil, err
 	}
 
-	vulnerabilities := make([]Vulnerability, 0, len(matches))
-	for _, match := range matches {
-		name := filepath.Base(match)
+	vulnerabilities := make(map[string]*Vulnerability, len(matchingFilepaths))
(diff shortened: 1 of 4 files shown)


sysfs/vulnerability.go
```diff
@@ -59,29 +73,29 @@ func (fs FS) CPUVulnerabilities() ([]Vulnerability, error) {
...
 
-func parseVulnerability(name, value string) (Vulnerability, error) {
-	v := Vulnerability{CodeName: name}
-	value = strings.TrimSpace(value)
-	if value == notAffected {
-		v.State = notAffected
+func parseVulnerability(name, rawContent string) (*Vulnerability, error) {
+	v := &Vulnerability{CodeName: name}
+	rawContent = strings.TrimSpace(rawContent)
+	if rawContent == notAffected {
```
SuperQ, line 83: I wonder if we should `strings.ToLower(rawContent)` to make this less fragile.

mwasilew2: that's a good point

I've pushed another version which unifies character cases when parsing the state, let me know what you think.

I've also added test cases for vulnerable state

sysfs/vulnerability_test.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2019 The Prometheus Authors
```
SuperQ, line 1: Nit, new files get the current year.

```suggestion
// Copyright 2023 The Prometheus Authors
```

sysfs/vulnerability.go
```diff
@@ -59,29 +73,37 @@ func (fs FS) CPUVulnerabilities() ([]Vulnerability, error) {
...
 
-func parseVulnerability(name, value string) (Vulnerability, error) {
-	v := Vulnerability{CodeName: name}
-	value = strings.TrimSpace(value)
-	if value == notAffected {
-		v.State = notAffected
+func parseVulnerability(name, rawContent string) (*Vulnerability, error) {
+	v := &Vulnerability{CodeName: name}
+	rawContent = strings.TrimSpace(rawContent)
+	rawContentLower := strings.ToLower(rawContent)
```
SuperQ, line 83: Maybe simplify and just use the lower string for everything.

```suggestion
	vulnerabilityState := strings.ToLower(strings.TrimSpace(rawContent))
```

SuperQ (changes requested):
Minor nits, otherwise great.

SuperQ (changes requested):
Minor nits, otherwise great.

sysfs/vulnerability.go
mwasilew2: @SuperQ 

I didn't want to lose upper cases in the mitigation text, for example:
```
Retpolines, IBPB: conditional, STIBP: always-on, RSB filling, PBRSB-eIBRS: Not affected
```

It felt like bringing the whole text to lower cases would result in reduced legibility. I lean towards leaving cases as they are, but I don't feel particularly strongly about it. If you think that simplifying this to a single var is more important than keeping the cases than let me know and I'll push another patch.

sysfs/vulnerability.go
```diff
@@ -24,33 +24,47 @@ import (
...
+)
+
+const (
+	VulnerabilityStateNotAffected = iota
+	VulnerabilityStateVulnerable
+	VulnerabilityStateMitigation
+)
+
+var (
+	VulnerabilityHumanEncoding = map[int]string{
```
discordianfish, line 39: Can't find this being used anywhere?

sysfs/vulnerability.go
```diff
@@ -59,29 +73,37 @@ func (fs FS) CPUVulnerabilities() ([]Vulnerability, error) {
...
-	v := Vulnerability{CodeName: name}
-	value = strings.TrimSpace(value)
-	if value == notAffected {
-		v.State = notAffected
+func parseVulnerability(name, rawContent string) (*Vulnerability, error) {
+	v := &Vulnerability{CodeName: name}
+	rawContent = strings.TrimSpace(rawContent)
+	rawContentLower := strings.ToLower(rawContent)
+
+	if strings.HasPrefix(rawContentLower, notAffected) {
```
discordianfish, line 85: Turn these into switch cases

discordianfish: Also why HasPrefix and not ==?

sysfs/vulnerability.go
mwasilew2: It's just a package variable holding the mapping between the state and a string describing the state. It's not used anywhere in this PR. It can be used for example by external consumers of this library to get a string describing the state of the vulnerability.

I've got a `node_exporter` PR waiting to be opened which is blocked on this PR. The `node_exporter` PR requires strings in order to fill labels. `node_exporter` will get a map containing vulnerabilities with encoded states and then will use this variable to map the states to more human friendly strings.

It's just an optimization I thought about. Rather than storing the exact same strings multiple times (in each Vulnerability) I thought it would make sense to encode those and store ints instead.

One could argue that I over engineered this or that it's premature optimization. I could do some benchmarks to see if it actually makes a difference, but it's a tiny change so I'm not sure if it's worth the effort. One of the maintainers could weigh in and say that simplicity is far more important at this stage (which makes total sense to me as well).

I'm open to any suggestions and feedback. Let me know what you think.

sysfs/vulnerability.go
mwasilew2: > Turn these into switch cases

yup, makes sense, will push a change in a sec

> Also why HasPrefix and not ==?

that's because the actual content returned from sysfs can have multiple different formats, e.g.:
- `Vulnerable`
- `Vulnerable: Clear CPU buffers attempted, no microcode`
- `Mitigation`
- `Mitigation: Retpolines, IBPB: conditional, STIBP: always-on, RSB filling, PBRSB-eIBRS: Not affected`
- `Not affected`

(see fixtures for more examples)

And checking the prefix was one of the simplest solutions I could come up with for identifying the state. I don't really like this approach, I'm open to any suggestion on how this could be done better

mwasilew2:
@discordianfish @SuperQ is there anything else I can help with here?

sysfs/vulnerability.go
SuperQ: Sure, that makes sense to keep the original string caps.

sysfs/vulnerability.go
SuperQ: Adding a Go doc string here would help.

```suggestion
	// VulnerabilityHumanEncoding something something
	VulnerabilityHumanEncoding = map[int]string{
```

SuperQ (approved):
LGTM

mwasilew2:
rebased onto upstream

mwasilew2:
the codespell job is failing, but it's unrelated to this change

Here's the job definition: https://github.com/prometheus/procfs/blob/abbfda0b3c04c4536faa158b44a4047e357ede23/.circleci/config.yml#L38-L45

the docker image hasn't changed recently: https://hub.docker.com/r/circleci/python/tags

codespell pip package was updated yesterday: https://pypi.org/project/codespell/#history

I think this is unrelated and should be handled in a separate issue/PR. I'll open a separate one

discordianfish (approved):

SuperQ:
Going to ignore the codespell, it's fixed in HEAD.

Issue #506, opened 2023-04-15
support UDP drops


funkiestj:
on newer linux systems /proc/net/udp (and /proc/net/udp6) have a drop column on the far right:
```

$ uname -a
Linux ip-172-31-10-56 5.4.0-1045-aws #47~18.04.1-Ubuntu SMP Tue Apr 13 15:58:14 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux 

$ cat /proc/net/udp
  sl  local_address rem_address   st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode ref pointer drops
 4000: 3500007F:0035 00000000:0000 07 00000000:00000000 00:00000000 00000000   101        0 256693 2 0000000000000000 0
 4015: 380A1FAC:0044 00000000:0000 07 00000000:00000000 00:00000000 00000000   100        0 1016179 2 0000000000000000 0
 5890: 0100007F:8797 0100007F:C351 01 00000000:00000000 00:00000000 00000000  1001        0 1016395 2 0000000000000000 0
 8736: 00000000:12B5 00000000:0000 07 00000000:00000000 00:00000000 00000000     0        0 32052 2 0000000000000000 0
12419: 00000000:2118 00000000:0000 07 00000000:00000000 00:00000000 00000000     0        0 24157 2 0000000000000000 0
```

alebsys:
@discordianfish, can you please assign this to me? tnx

alebsys:
The TCP and UDP interfaces are very different after the `inode` column in `/proc/net/udp` | `/proc/net/tcp` files.
Maybe it's worth creating separate structures for UDP and TCP instead of a single `netIPSocketLine`?

PR #538, opened 2023-06-18
Add udp drops

alebsys:
Changes add parsing of the field drops in /proc/net/udp{,6} files.
Fixes: #506

### DIFF
--- a/net_ip_socket.go
+++ b/net_ip_socket.go
@@ -24,12 +24,6 @@ import (
 	"strings"
 )
 
-var (
-	// isUDP is used to determine the type of network socket UDP or TCP.
-	// Further its value will be used to parse UDP-specific fields from /proc/net/udp{,6} files.
-	isUDP bool
-)
-
 const (
 	// readLimit is used by io.LimitReader while reading the content of the
 	// /proc/net/udp{,6} files. The number of lines inside such a file is dynamic
@@ -87,7 +81,7 @@ func newNetIPSocket(file string) (NetIPSocket, error) {
 	defer f.Close()
 
 	var netIPSocket NetIPSocket
-	var isUDP bool
+	var isUDP bool // is used to determine the type of network socket UDP or TCP
 
 	if strings.Contains(file, "udp") {
 		isUDP = true
@@ -120,6 +114,7 @@ func newNetIPSocketSummary(file string) (*NetIPSocketSummary, error) {
 
 	var netIPSocketSummary NetIPSocketSummary
 	var udpPacketDrops uint64
+	var isUDP bool // is used to determine the type of network socket UDP6 or TCP6
 
 	if strings.Contains(file, "udp") {
 		isUDP = true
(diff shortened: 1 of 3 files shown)


net_ip_socket.go
```diff
@@ -24,6 +24,12 @@ import (
 	"strings"
 )
 
+var (
+	// isUDP is used to determine the type of network socket UDP or TCP.
+	// Further its value will be used to parse UDP-specific fields from /proc/net/udp{,6} files.
+	isUDP bool
```
discordianfish, line 30: Why is this a global variable?

net_ip_socket.go
```diff
@@ -224,5 +248,14 @@ func parseNetIPSocketLine(fields []string) (*netIPSocketLine, error) {
 		return nil, fmt.Errorf("%s: Cannot parse inode value in %q: %w", ErrFileParse, line.Inode, err)
 	}
 
+	// drops
+	if isUDP {
+		drops, err := strconv.ParseUint(fields[12], 0, 64)
```
discordianfish, line 253: What is fields[12] for non udp? Would it make sense to always set drops if fields[12] exist?

alebsys: Unfortunately, after the inode column, the file structure is different - in the `/proc/net/udp` file it is drops, but in `/proc/net/tcp` it is [retransmit timeout](https://www.kernel.org/doc/Documentation/networking/proc_net_tcp.txt).

net_ip_socket.go
alebsys: You are right, I will change to a local variable.

net_ip_socket.go
discordianfish: Ok makes sense

discordianfish (approved):
LGTM

net_ip_socket.go
```diff
@@ -77,13 +81,18 @@ func newNetIPSocket(file string) (NetIPSocket, error) {
 	defer f.Close()
 
 	var netIPSocket NetIPSocket
+	var isUDP bool // is used to determine the type of network socket UDP or TCP
+
+	if strings.Contains(file, "udp") {
+		isUDP = true
+	}
```
SuperQ, line 88:
```suggestion
	isUDP := strings.Contains(file, "udp")
```

net_ip_socket.go
```diff
@@ -104,19 +113,29 @@ func newNetIPSocketSummary(file string) (*NetIPSocketSummary, error) {
 	defer f.Close()
 
 	var netIPSocketSummary NetIPSocketSummary
+	var udpPacketDrops uint64
+	var isUDP bool // is used to determine the type of network socket UDP6 or TCP6
+
+	if strings.Contains(file, "udp") {
+		isUDP = true
+	}
```
SuperQ, line 121:
```suggestion
	isUDP := strings.Contains(file, "udp")
```

SuperQ (changes requested):
A nice simplification.

discordianfish:
@SuperQ needs to approve/merge

darkrift:
Any chance we can see this move forward ? @SuperQ

darkrift:
@discordianfish Can you do something to get this move forward ?

SuperQ (approved):

Issue #492, opened 2023-02-15
nfsd v4 ops index mismatch


atkinsam:
Hi, I believe the nfsd v4 ops parsing here uses indices which are mismatched on some ops. https://github.com/prometheus/procfs/blob/6efaedecce726b918b26265fdebc9548008f0e5b/nfs/parse.go#L269 

I can see that many/most are correct:
 - [access uses index 4](https://github.com/prometheus/procfs/blob/6efaedecce726b918b26265fdebc9548008f0e5b/nfs/parse.go#L279), which corresponds with [index 3 in the kernel](https://elixir.bootlin.com/linux/v5.10.162/source/include/linux/nfs4.h#L79) (the off-by-one appears to be due to the parser leaving the [LAST_NFS4_OP](https://elixir.bootlin.com/linux/v5.10.162/C/ident/LAST_NFS4_OP) value [in the array](https://github.com/prometheus/procfs/blob/6efaedecce726b918b26265fdebc9548008f0e5b/nfs/parse.go#L270)
 - [close uses index 5](https://github.com/prometheus/procfs/blob/6efaedecce726b918b26265fdebc9548008f0e5b/nfs/parse.go#L280), which corresponds with [index 4 in the kernel](https://elixir.bootlin.com/linux/v5.10.162/source/include/linux/nfs4.h#L80).

But the write operation, for example, appears to be incorrect. The [parser uses index 37](https://github.com/prometheus/procfs/blob/6efaedecce726b918b26265fdebc9548008f0e5b/nfs/parse.go#L312), but the write operation uses [index 38 in the kernel](https://elixir.bootlin.com/linux/v5.10.162/source/include/linux/nfs4.h#L114). So the parser should really be using index 39. 

I noticed this by comparing stats emitted by [node_exporter](https://github.com/prometheus/node_exporter) with what I see in `nfsstat -s`.

discordianfish:
Thanks! Can you submit a PR to fix this?

jjaakkol:
I can confirm that this pull request fixes the issue. My Grafana dashboard now correctly reports NFSv4 writes.

PR #543, opened 2023-07-04
[fix]: fix nfsd v4 ops index miss match

dongjiang1989:
#492 

/kind fix

1. fix nfsd v4 ops index miss match
2. add unittest for `proc4ops 72` , `proc4ops 59`, `proc4ops 40`


dongjiang1989:
@discordianfish  PTAL

discordianfish:
Are we sure there are no systems with only 39 fields when using nfsd v4? I would assume this worked in the past?

dongjiang1989:
> Are we sure there are no systems with only 39 fields when using nfsd v4? I would assume this worked in the past?

The previous version of procfs supports the `OP_RELEASE_LOCKOWNER` field. It’s just that the lack of `OP_SETCLIENTID` and `OP_SETCLIENTID_CONFIRM` caused the index miss match.

dongjiang1989:
@SuperQ  PTAL!  Thinks.

dongjiang1989:
@discordianfish @SuperQ  When is this likely to be merged and a release available? so I can manage expectations.

discordianfish:
We don't have a fixed release schedule. First @SuperQ needs to review and merge this.

jjaakkol:
Ok, I just noticed today that Write operations won't get counted at all in my dashboard and wrote almost exactly the same patch myself, before noticing this.  I don't think this can break any old systems. The numbering comes directly from NFSv4 RFC and write request there is number 38. Linux kernel just uses those exact same numbers. Thanks @dongjiang1989 !

https://www.rfc-editor.org/rfc/rfc7530.html#page-299

dongjiang1989:
> Ok, I just noticed today that Write operations won't get counted at all in my dashboard and wrote almost exactly the same patch myself, before noticing this. I don't think this can break any old systems. The numbering comes directly from NFSv4 RFC and write request there is number 38. Linux kernel just uses those exact same numbers. Thanks @dongjiang1989 !
> 
> https://www.rfc-editor.org/rfc/rfc7530.html#page-299

@jjaakkol  Thank you for your code review. If you have any other questions about this pull request, please submit them to me!

Issue #546, opened 2023-07-09
Couldn't parse \"/proc/stat\": bufio.Scanner: token too long


drdeimos:
### Host operating system: 
Linux k42jv 6.4.2 prometheus/node_exporter#1 SMP PREEMPT_DYNAMIC Thu Jul  6 15:29:24 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### node_exporter version: 

```
node_exporter, version 1.6.0 (branch: HEAD, revision: ff7f9d69b645cb691dd3e84dc3afc88f5c006962)
  build user:       root@f9c3ed0cfbd3
  build date:       20230527-12:03:54
  go version:       go1.20.4
  platform:         linux/amd64
  tags:             netgo osusergo static_build
```

### node_exporter command line flags
```
/usr/local/bin/node_exporter --web.listen-address=127.0.0.1:9101 --collector.netdev.address-info --collector.ntp --collector.ntp.server-is-local --collector.filesystem.mount-points-exclude (^/(dev|proc|sys|run|var/lib/kubelet)($|/))|(^/var/lib/docker/) --collector.filesystem.fs-types-exclude ^(autofs|binfmt_misc|cgroup|configfs|debugfs|devpts|devtmpfs|fusectl|fuse\.lxcfs|hugetlbfs|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|sysfs|tracefs|squashfs)$
```

### node_exporter log output
<details>
  <summary>log</summary>
  
  ```
Jul 07 21:58:49 k42jv systemd[1]: Started Prometheus exporter for hardware and OS metrics.
Jul 07 21:58:50 k42jv node_exporter[484]: ts=2023-07-07T18:58:50.626Z caller=node_exporter.go:180 level=info msg="Starting node_exporter" version="(version=1.6.0, branch=HEAD, revision=ff7f9d69b645cb691dd3e84dc3afc88f5c006962)"
Jul 07 21:58:50 k42jv node_exporter[484]: ts=2023-07-07T18:58:50.635Z caller=node_exporter.go:181 level=info msg="Build context" build_context="(go=go1.20.4, platform=linux/amd64, user=root@f9c3ed0cfbd3, date=20230527-12:03:54, tags=netgo osusergo static_build)"
[...]

PR #547, opened 2023-07-10
Increase Stat() buffer size

SuperQ:
Increase the /proc/stat scanner tokens size from the default of 65k to 1024k. This allows for scanning of large `intr` lines with > 65k columns.
* Increase the scanner buffer from the default 4kB to 8kB bytes at a time.

Fixes: https://github.com/prometheus/procfs/issues/546


discordianfish (approved):
LGTM

stat.go
```diff
@@ -187,6 +187,9 @@ func parseStat(r io.Reader, fileName string) (Stat, error) {
 		err error
 	)
 
+	buf := make([]byte, 0, 8*1024)
```
discordianfish, line 190: Maybe add a comment that we increased this due to long intr lines. Also might want to make this a const? But fine with me either way

SuperQ: Added a comment.

PR #549, opened 2023-07-11
Add `hw_counters` for infiniband device.

aztecher:
This change adds `hw_counters` to the existing infiniband class so that additional counter information for infiniband devices can be obtained.

In preparing this PR, I have confirmed that `hw_counters` for Infiniband devices can be obtained from the `mlx5_core` kernel module included in multiple Linux distributions and multiple OS versions.
I have also confirmed that it is possible to get `hw_counters` from the `mlx5_core` kernel module included in the [OFED driver](https://network.nvidia.com/products/infiniband-drivers/linux/mlnx_ofed/).
However, I have not checked for other cases (e.g. mlx4_core), so this implementation is limited to mlx5 only.

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -67,6 +67,42 @@ type InfiniBandCounters struct {
 	VL15Dropped                  *uint64 // counters/VL15_dropped
 }
 
+// InfiniBandHwCounters contains counter value from files in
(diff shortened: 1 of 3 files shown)


SuperQ (approved):

gangxie112:
Which release of the node-exporter will include this change?

discordianfish:
> Which release of the node-exporter will include this change?

The next one, we'll release a new version probably soon

PR #552, opened 2023-07-19
Add inode number to fdinfo

ctrox:
This simply adds the inode number to `ProcFDInfo` as I noticed its absence. I left it a string to be consistent with the other fields.


discordianfish (approved):
LGTM, thanks!

SuperQ (approved):

PR #553, opened 2023-07-26
Custom build tag "nostatfs" disables statfs

mhutchinson:
This allows users to specify the "nostatfs" build tag to disable this feature if needed.

In our case, we are using Tamago (https://github.com/usbarmory/tamago), which does not support syscall.Statfs. Without this change attempting to build returns 'undefined: syscall.Statfs' and 'undefined: syscall.Statfs_t'. With this change, we can build successfully build our tamago executables provided that "nostatfs" is added to the build tags.

This option seems preferable to adding "tamago" as a build tag here, and will scale to other use cases too.

### DIFF
--- a/fs_statfs_notype.go
+++ b/fs_statfs_notype.go
@@ -11,8 +11,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build netbsd || openbsd || solaris || windows
-// +build netbsd openbsd solaris windows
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
Thanks!

mhutchinson:
Thanks for the quick review @SuperQ! So I can manage expectations, when is this likely to be merged and a release available?

SuperQ:
Release v0.11.1 tagged.

PR #556, opened 2023-08-05
infiniband: do not fail if board_id is not present

dswarbrick:
Whilst a large majority do, not _all_ InfiniBand drivers expose a board_id in sysfs.

Fixes: https://github.com/prometheus/node_exporter/issues/2769

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -121,16 +121,24 @@ func (fs FS) InfiniBandClass() (InfiniBandClass, error) {
 }
 
 // Parse one InfiniBand device.
+// Refer to https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-infiniband
 func (fs FS) parseInfiniBandDevice(name string) (*InfiniBandDevice, error) {
 	path := fs.sys.Path(infinibandClassPath, name)
 	device := InfiniBandDevice{Name: name}
 
-	for _, f := range [...]string{"board_id", "fw_ver", "hca_type"} {
+	// fw_ver is exposed by all InfiniBand drivers since kernel version 4.10.
+	value, err := util.SysReadFile(filepath.Join(path, "fw_ver"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to read HCA firmware version: %w", err)
+	}
+	device.FirmwareVersion = value
+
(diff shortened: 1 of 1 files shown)


sysfs/class_infiniband.go
```diff
@@ -121,16 +121,24 @@ func (fs FS) InfiniBandClass() (InfiniBandClass, error) {
 }
 
 // Parse one InfiniBand device.
+// Refer to https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-infiniband
 func (fs FS) parseInfiniBandDevice(name string) (*InfiniBandDevice, error) {
 	path := fs.sys.Path(infinibandClassPath, name)
 	device := InfiniBandDevice{Name: name}
 
-	for _, f := range [...]string{"board_id", "fw_ver", "hca_type"} {
+	// fw_ver is exposed by all InfiniBand drivers since kernel version 4.10.
```
SuperQ, line 129: We typically support older kernels than this. Maybe we should just make all of these files/values optional.

SuperQ:
I wonder if we should make the values in the struct `*string` and set it to nil if the file is not found.

sysfs/class_infiniband.go
dswarbrick: Note that this PR is does not change the behaviour for `fw_ver`, which has been a mandatory requirement of this collector since it first landed in June 2019. It only makes the `board_id` optional, and `hca_type` was already optional.

I would be very surprised if anybody is running InfiniBand on kernels older than 4.10. I don't recall a single issue being logged about this collector failing due to missing `fw_ver`.

dswarbrick:
> I wonder if we should make the values in the struct `*string` and set it to nil if the file is not found.

Doing so means that consumers of this need to check for non-nil before using the struct members. I'm not sure if that's a good thing or not. Given that Prometheus does not store empty-string labels, my gut feeling is that it's more convenient to simply leave these struct members as empty strings if not found, meaning that node_exporter can just blindly assign them to labels.

discordianfish (approved):

mustafakemalgilor:
Related-to: https://github.com/prometheus/node_exporter/issues/2769

SuperQ (approved):

Issue #565, opened 2023-09-13
parseMemInfo() ignores unit, leading to incorrect results from /proc/meminfo


tjhop:
Hi all :wave: 

Is there a reason that the parseMemInfo() function [ignores the optional unit field](https://github.com/prometheus/procfs/blob/113c5013dda3c600bda241d86c64258ec7117c7b/meminfo.go#L165)? This can lead to incorrect results on systems that _do_ supply the field, as the units are obviously not adjusted accordingly. 

A small example program/output:
`meminfo.go`
```golang
package main

import (
	"bufio"
	"fmt"
	"os"
	"strings"

	"github.com/dustin/go-humanize"
	"github.com/prometheus/procfs"
)

func main() {
	fmt.Println("Meminfo sample entries:")

	meminfoFile, err := os.Open("/proc/meminfo")
	if err != nil {
		panic(err)
	}
	defer meminfoFile.Close()

	scanner := bufio.NewScanner(meminfoFile)
	for scanner.Scan() {
		line := scanner.Text()
		fields := strings.Fields(line)
		key := strings.TrimSuffix(fields[0], ":")
		if key == "MemTotal" || key == "MemFree" {
			fmt.Println(line)

			val, err := humanize.ParseBytes(fmt.Sprintf("%s %s", fields[1], fields[2]))
			if err != nil {
				panic(err)
			}
			fmt.Printf("%s humanized: %s\n", key, humanize.Bytes(val))
		}
	}

	if scanner.Err() != nil {
		if err != nil {
			panic(err)
		}
	}

	fmt.Println()
	fmt.Println("procfs lib values:")
	fs, err := procfs.NewFS("/proc")
	if err != nil {
		panic(err)
	}

	meminfo, err := fs.Meminfo()
	if err != nil {
		panic(err)
	}

	fmt.Printf("MemTotal: %d bytes\n", *meminfo.MemTotal)
	fmt.Printf("MemTotal humanized: %s\n", humanize.Bytes(*meminfo.MemTotal))
	fmt.Printf("MemFree: %d bytes\n", *meminfo.MemFree)
	fmt.Printf("MemFree humanized: %s\n", humanize.Bytes(*meminfo.MemFree))
}
```

output:
```bash
/tmp/procfs -> go run meminfo.go
Meminfo sample entries:
MemTotal:       32588196 kB
MemTotal humanized: 33 GB
MemFree:         9787012 kB
MemFree humanized: 9.8 GB

procfs lib values:
[...]

PR #569, opened 2023-09-15
fix(meminfo): account for optional unit field so values are accurate

tjhop:
Addresses: #565

Previously, this function was ignoring the optional unit field, leading to incorrect results on systems that do report the field. This uses the humanize lib used elsewhere within the Prometheus ecosystem to normalize the value to bytes, including when a unit is provided.


SuperQ (changes requested):
I think this is a good idea, but it's going to silently change the results of the struct.

I think we need a breaking change here in order to signal to downstream code that the results are different.

Maybe we should rename the struct from `Meminfo` to `MeminfoBytes` in order to make it clear that the values returned are in bytes.

Or we add new struct fields like:

```
MemTotal: val
MemTotalBytes: scaledValue(val, unit)
```

tjhop:
I had similar thoughts and was considering suggesting a complimentary struct/function for `MeminfoBytes`, `parseMeminfoBytes()`, `fs.MeminfoBytes()`, etc so that the existing code would continue to work as expected. 

I think adding in new struct fields for the scaled values to live alongside the raw/unscaled values is a clever solution. I'll work on it 👍

tjhop:
@SuperQ I revamped a bit to try the shadow struct fields, lemme know your thoughts

meminfo.go
```diff
@@ -162,114 +217,176 @@ func parseMemInfo(r io.Reader) (*Meminfo, error) {
...
 		}
 
+		switch len(fields) {
+		case 2:
+			// No unit present, use the parsed the value as bytes directly.
+			val = v
+			valBytes = v
+		case 3:
+			// Unit present in optional 3rd field, convert and normalize to bytes.
+			bytes, err := humanize.ParseBytes(fmt.Sprintf("%s %s", fields[1], fields[2]))
```
SuperQ, line 235: `humanize.ParseBytes()` is producing incorrect results as it thinks `kB` is 1000. The kernel values for this are KiB (1024).

The kernel is [hard-coded to only have one representation](https://github.com/torvalds/linux/blob/3aba70aed91f2b283f7952be152ad76ec5c34975/fs/proc/meminfo.c#L28-L32), which is `kB`.

Let's leave out the `humanize` package and simply check for the string `kB` in the fields. If the extra field is different we can return a parse error.

tjhop: Ah, I didn't realize the kernel only had a single unit available for the optional field -- I didn't see direct docs on kernel.org related to the unit field for meminfo, and the docs I did find read to me as if other units were available. Given the misinterpretation of the unit itself and how static the underlying data is, it seems fair to just handle the `kB` case. 

Rebased and then noticend a new release just came out, so refactored on top of current master as well. I think it's ready for another look :+1:

SuperQ (approved):
Thanks!

meminfo.go
```diff
@@ -162,114 +214,179 @@ func parseMemInfo(r io.Reader) (*Meminfo, error) {
...
 	s := bufio.NewScanner(r)
 	for s.Scan() {
-		// Each line has at least a name and value; we ignore the unit.
 		fields := strings.Fields(s.Text())
-		if len(fields) < 2 {
-			return nil, fmt.Errorf("%w: Malformed line %q", ErrFileParse, s.Text())
-		}
+		var val, valBytes uint64
 
 		v, err := strconv.ParseUint(fields[1], 0, 64)
```
SuperQ, line 220: How about we change this:

```suggestion
		val, err = strconv.ParseUint(fields[1], 0, 64)
```

Then we can drop the `val = v` below.

meminfo.go
```diff
@@ -140,6 +140,58 @@ type Meminfo struct {
...
+	HardwareCorruptedBytes *uint64
+	AnonHugePagesBytes     *uint64
+	ShmemHugePagesBytes    *uint64
+	ShmemPmdMappedBytes    *uint64
+	CmaTotalBytes          *uint64
+	CmaFreeBytes           *uint64
+	HugePagesTotalBytes    *uint64
+	HugePagesFreeBytes     *uint64
+	HugePagesRsvdBytes     *uint64
+	HugePagesSurpBytes     *uint64
```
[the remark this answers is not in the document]
tjhop: Makes sense -- I'll get rid of the `Bytes` fields for these. Since those fields in the kernel explicitly never emit a unit, they should always follow the `len == 2` path and never be subject to the scaling logic, so I feel comfortable not adding in any explicit handling to avoid scaling those fields. Please lemme know if you feel differently

meminfo.go
tjhop: Addressed in f7b4621

meminfo.go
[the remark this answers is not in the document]
tjhop: Addressed in 8607561

tjhop:
Amended the last commit to fix an accidental deleted line (thank you tests). Force pushed to fix. Only the last commit was amended, so I believe the approval should still be valid since that commit wasn't touched.

discordianfish (approved):

discordianfish:
Great, thanks!

PR #574, opened 2023-09-21
NFSd: handle new wdeleg_getattr attribute in /proc/net/rpc/nfsd

klausman:
This attribute was introduced it v6.6-rc1.

I am unsure if toplevel is the right place for the metric struct-wise, but it is a sole uint64 that doesn't seem to fit any of the other categories in an obvious manner. Unfortunately, I have also been unable to find any documentation of the meaning of this stat.

In the short run, I at leas would like my node-exporters to not just not export any NFSd metrics anymore just because there is one new stat, and this seems the way to approach the matter.


SuperQ (approved):
Thanks!

SuperQ:
The golangci-lint issue is unrelated.

PR #579, opened 2023-10-10
Add support for for /proc/net/tls_data kTLS stats

defect:
Since Linux kernel 4.17 (I believe) there is support to move some TLS operations to kernel space, enabling things like TLS `sendfile` to improve performance. Statistics for this feature is exposed under `/proc/net/tls_stats`. This PR adds support to read these statistics. It is very much based on the support for `/proc/net/xfrm_stats` in `net_xfrm.go` as they both share the same format.

For more info on Linux kTLS see https://docs.kernel.org/networking/tls.html.

### DIFF
--- a/net_tls_stat.go
+++ b/net_tls_stat.go
@@ -1,0 +1,119 @@
+// Copyright 2023 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


net_tls_stat.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2013 Prometheus Team
```
discordianfish, line 1: Update year

net_tls_stat_test.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017 Prometheus Team
```
discordianfish, line 1: Same

discordianfish:
Would be nice if we could make the parsing for /proc/net/xfrm_stats more generic and re-use that here, but I think it's fine either way

discordianfish (approved):

defect:
Yeah, i was thinking about refactoring that out but didn't really find a clean way to do it. Maybe using annotated structs a la `json` and `yaml` and friends? But that feels a bit overkill, at least for now.

discordianfish:
@defect Fair enough, that might be better be part of a bigger refactoring

SuperQ (approved):

PR #597, opened 2023-12-13
Add support for guest_time and cguest_time stats.

thomasbarrett:
This PR adds support for the `guest_time` and `cguest_time` fields from `/proc/<pid>/stat`. These are useful for monitoring kvm guest processes.


SuperQ (approved):
I needed to double check, but this should be fine since these fields have been supported since Linux 2.6.24. We only support [Linux 2.6.32 and later](https://tip.golang.org/wiki/MinimumRequirements#linuxlinux) due to Go upstream minimums.

PR #601, opened 2024-01-22
golint: redundant type from array, slice, or map composite literal

manas-rust:
golint: redundant type from array, slice, or map composite literal


SuperQ (approved):
Thanks!

Issue #589, opened 2023-11-16
procfs still unable to collect IB counters


mtds:
* `procfs` is still unable to aid Node Exporter in collecting IB metrics **when** the `irdma` module is loaded but only the cards in use are those using the `mlx5_0` module.
* I have detailed the problem in the following issue in the node exporter repository: [#2846: Infiniband metrics: still not collected when irdma is loaded](https://github.com/prometheus/node_exporter/issues/2846)
* References to a similar issue and a corresponding PR which solved the problem for the latest `procfs` release (**v0.12.0**):
   - [Infiniband metrics not collected if irdma kernel module is loaded](https://github.com/prometheus/node_exporter/issues/2769)
   - #556

PR #605, opened 2024-02-15
infiniband: support Intel irdma devices

dswarbrick:
This skips parsing `/sys/class/infiniband/<device>/ports/<port-num>/counters` for irdma devices, which do not expose it, and would previously cause parseInfiniBandPort() to return an error. Instead, irdma devices expose `/sys/class/infiniband/<device>/ports/<port-num>/hw_counters`.

Some background info added as function description comments also.

Fixes: #589


SuperQ (approved):
Thanks!

PR #607, opened 2024-02-19
chore: class_fibrechannel: support optional attributes

machine424:


SuperQ (approved):

SuperQ:
It might be nice to change the struct values to be references, so we can differentiate between nil and 0.

machine424:
> It might be nice to change the struct values to be references, so we can differentiate between nil and 0. We do this in several other parsers.

You're right, it may be confusing, especially that we allow missing counters files. It's worth an issue I think.

Could we expected a `procfs` release any soon? I can help if needed.

SuperQ:
I can cut a release if we need one. Do you want to change the struct first? Or do that later?

PR #614, opened 2024-03-19
util: throw if file has !digit characters

rexagod:
Throw if file has non-numeric data, in addition to hinting the path that caused the error.

Fixes: prometheus#304, prometheus/node_exporter#1710

### DIFF
--- a/internal/util/parse.go
+++ b/internal/util/parse.go
@@ -14,9 +14,11 @@
 package util
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
+	"unicode"
 )
 
 // ParseUint32s parses a slice of strings into a slice of uint32s.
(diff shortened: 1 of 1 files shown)


discordianfish:
See ci errors

discordianfish (approved):
LGTM

internal/util/parse.go
```diff
@@ -85,6 +87,14 @@ func ReadUintFromFile(path string) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
+	for _, c := range data {
+		if string(c) == "\n" {
+			continue
+		}
+		if !unicode.IsDigit(rune(c)) {
```
SuperQ, line 94: Hmm, this is going to match a whole bunch of unicode glyph runes that are not `[0-9]`. I think we should be more strict and check like this:

 ```suggestion
		if !strings.ContainsAny(c, "0123456789") {
```

SuperQ (changes requested):

SuperQ (changes requested):
Actually, looking at the upstream issue, we already see that this function is returning `strconv.ErrSyntax` from `strconv.ParseUint()`. I think the issue is the node_exporter is just not ignoring this specific error.

I don't think we need to make this change as the upstream `strconv.ParseUint()` is already doing a rune check on the string.

Issue #507, opened 2023-04-18
strings.TrimRight is used instead of TrimSuffix


GrosQuildu:
This line aims to remove " kB" suffix, but instead removes all characters from " kB" cutset. 
https://github.com/prometheus/procfs/blob/cdc45126aef95c2e741f1d56ded3d24fdcd3d167/proc_smaps.go#L130

Similar fix as was done in this commit is required - https://github.com/prometheus/procfs/commit/b3d77ecba8e7aeb7b4ff946644b1013199656c42 .

discordianfish:
Oops, thanks! Wanna submit a quick PR to fix this?

PR #618, opened 2024-03-21
bugfix: `s/TrimRight/TrimSuffix` for certain cases

rexagod:
Use `TrimSuffix` instead of `TrimRight` in cases where an exact suffix string needs to be removed, not all occurrences of all characters in a specified cutset.

Fixes: #507


discordianfish (approved):
LGTM

rexagod:
Is this good to go in now? I'm not sure if anything is needed here from my end, but if so, PLMK, thanks!

SuperQ (approved):

Issue #612, opened 2024-03-13
class_fibrechannel: distinguish between missing and zero


machine424:
See https://github.com/prometheus/procfs/pull/607#issuecomment-1978779539

PR #623, opened 2024-03-23
enhancement: Use pointer fields for `FibreChannel*`

rexagod:
Allow fields to be `nil`-checked.

Fixes: #612

PS. This is a **breaking** change.
***

@machine424 I took the liberty of sending out a PR targeting this issue, but please feel free to close this if you already have a patch on the way. :)


discordianfish (approved):
Makes sense

sysfs/class_fibrechannel_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"k8s.io/utils/ptr"
```
SuperQ, line 23: While this is a convenient package, I'm not sure we want to add this to the dependency tree.

This package is pulled in by a lot of places and I would like to minimize the indirect requirements.

In other similar tests, it's painful, but we do string vars and then `&stringVar`.

SuperQ (changes requested):

discordianfish:
@SuperQ PTAL

SuperQ (approved):

PR #625, opened 2024-03-28
Revert add avgRTT to nfs mountstats #487

SuperQ:
This is an artificiality computed value from existing values. The goal of this library is to provide simple raw access to values, rather than pre-compute data.

Reverts: https://github.com/prometheus/procfs/pull/487


ArthurSens (approved):
Hey 👋 

I'm not sure why I'm tagged here, I haven't worked on this library before 😬 
Looking at #629 I can see you were made a maintainer, so I believe you know what you're doing :) Giving the rubber stamp just in case you need someone to unblock the PR

Unfortunately, I won't have the time to fully understand the impact of this revert, I suggest that other maintainers chime in if you're looking for a longer discussion!

PR #626, opened 2024-03-29
Update build constraints for loong64 and s390x, Compatible with lower versions of go compiler

fangwentong:
go1.16 cannot recognize the loong64 instruction set architecture, and an error will occur during compilation. This MR explicitly specifies the instruction set build constraints in the source file.

```
❯ gvm use go1.16
Now using version go1.16
❯ go build -o main main.go
vendor/github.com/prometheus/procfs/cpuinfo_x86.go:20:5: parseCPUInfo redeclared in this block
	previous declaration at vendor/github.com/prometheus/procfs/cpuinfo_loong64.go:19:5
```


discordianfish:
Not sure we want/need to support go 1.16. Why do you need to compile it with that old version? @SuperQ wdyt?

SuperQ:
As of today this library requires Go 1.20 and newer.

Issue #428, opened 2021-12-17
Instantaneous vs. cummulative cpufreq stats


iustin:
Hi,

Looking around at issues such as #122 and #123, I see that in both cases, what is collected is the current (i.e. instantaneous) state of the CPUs, either hardware (`cpufreq_*` counters) or software (`scaling_*_freq`).

However, if compiled as such, the kernel does provide cummulative statistics via the `stats/` subdirectory - see https://www.kernel.org/doc/html/latest/cpu-freq/cpufreq-stats.html.

I think the `time_in_state` is the most interesting, followed by `total_trans`. These would allow much better fidelity of historical CPU state.

Thoughts?

rexagod:
+1. Looping in @discordianfish to check if this is request makes sense, if so, I'd be happy to raise a PR. (:

discordianfish:
Agreed, I think that would make sense

iustin:
Thanks both, I forgot to track this. Would be really good to have cumulative stats.

rexagod:
`trans_table` and `time_in_state` seem to exhibit a high cardinality nature owing to the CPU frequencies they track. `total_trans` seems fine to generate metrics for, though.

iustin:
Hmm. `total_trans` is interesting, but mostly a gimmick. Looking at one of my systems that has this module compiled, `time_in_state` has 10 rows, per CPU, so indeed that's a lot of metrics, but it is the real file that gives the data.

How about hiding this behind a flag, so only people interested enable it? It needs anyway a kernel module, that might not be always available.

SuperQ:
I think `time_in_state` metrics make the most useful data.

As this is a generically useful library, the decision on metrics exposition would happen in something like the `node_exporter`.

PR #627, opened 2024-04-01
feat: Add support for `CONFIG_CPU_FREQ_STAT`

rexagod:
Extracts data from the following files:
* `total_trans`
* `trans_table`
* `time_in_state`

Refer: https://www.kernel.org/doc/html/latest/cpu-freq/cpufreq-stats.html#configuring-cpufreq-stats.

Fixes: #428

### DIFF
--- a/sysfs/system_cpu.go
+++ b/sysfs/system_cpu.go
@@ -65,6 +65,10 @@ type SystemCPUCpufreqStats struct {
 	Governor                 string
 	RelatedCpus              string
 	SetSpeed                 string
+	// Refer `CONFIG_CPU_FREQ_STAT`: https://www.kernel.org/doc/html/latest/cpu-freq/cpufreq-stats.html#configuring-cpufreq-stats
+	CpuinfoFrequencyDuration         *map[uint64]uint64
+	CpuinfoFrequencyTransitionsTotal *uint64
+	CpuinfoTransitionTable           *[][]uint64
 }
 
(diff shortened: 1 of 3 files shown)


sysfs/system_cpu.go
```diff
@@ -292,19 +296,93 @@ func parseCpufreqCpuinfo(cpuPath string) (*SystemCPUCpufreqStats, error) {
 		}
 	}
 
+	// "total_trans" is the total number of times the CPU has changed frequency.
+	var cpuinfoFrequencyTransitionsTotal *uint64 = nil
+	cpuinfoFrequencyTransitionsTotalUint, err := util.ReadUintFromFile(filepath.Join(cpuPath, "stats", "total_trans"))
+	if err != nil {
+		if !(os.IsNotExist(err) || os.IsPermission(err)) {
+			return &SystemCPUCpufreqStats{}, err
+		}
```
discordianfish, line 305: Why are you not returning the error here but continue silently?

rexagod: https://github.com/prometheus/procfs/blob/9eed74490628b5d93360bd5e5971480c03fb43e0/sysfs/system_cpu.go#L273-L275

For these cases I saw that we were skipping over silently in other parts of the codebase, and kept the norm here.

discordianfish: Ah I see, makes sense

discordianfish (approved):
LGTM

PR #630, opened 2024-04-08
style: returns procfs build-in error like other parsing methods

weidongkl:
For file parsing errors, procfs defines error ErrFileParse. Other architectures, such as x86 ,arm,mips, use ErrFileParse to return parsing errors. Loong should do this too.


SuperQ (approved):
Thanks!

Issue #576, opened 2023-09-26
procfs has data consistency issues when reading /proc/net/tcp


shaohk:
https://github.com/prometheus/procfs/blob/ff0ad85f7e8bcd5c677d99143f14a2a3aab533aa/net_tcp.go#L28

I read tcp connections by procfs like this
```
package main

import (
	"fmt"

	"github.com/prometheus/procfs"
)

func main() {
	fs, _ := procfs.NewDefaultFS()

	tcps, _ := fs.NetTCP()
	for _, tcp := range tcps {
		fmt.Println(tcp.LocalAddr, tcp.LocalPort, tcp.RemAddr, tcp.RemPort, tcp.Inode)
	}
}
```
When I execute multiple times, I find that the same TCP connection appears twice. eg：
```
Tue Sep 26 18:48 [root@host ~]# ./nettcp | sort | uniq -c | grep '2 '
      ......
      1 192.168.0.2 22815 127.0.0.1 8181 4157386808
      2 192.168.0.2 24219 127.0.0.1 8300 4192187746
      1 192.168.0.2 29727 127.0.0.1 5002 4185948034
      ......
```

But, When I execute multiple netstat command, there won't be the phenomenon mentioned above. eg:
```
Tue Sep 26 18:50 [root@host ~]# netstat -npta | awk '{print $1, $4, $5, $7}' | sort | uniq -c
      ......
      1 tcp 192.168.0.2:11609 127.0.0.1:8300 138167/python
      1 tcp 192.168.0.2:12385 127.0.0.1:8300 138188/python
      1 tcp 192.168.0.2:12465 127.0.0.1:8035 266462/agent
      ......
```

And I find netstat source code, find the code that
```net-tools-2.1.0/lib/proc.c
FILE *proc_fopen(const char *name)
{
    static char *buffer;
    static size_t pagesz;
    FILE *fd = fopen(name, "r");

    if (fd == NULL)
      return NULL;

    if (!buffer) {
      pagesz = getpagesize();
      buffer = malloc(pagesz);
    }

    setvbuf(fd, buffer, _IOFBF, pagesz);

    return fd;
}
```
netstat sets the setvbuf to _IOFBF mode when opening the net/tcp file. But golang procfs doesn't set this.

How can we ensure consistency when reading net/tcp in Golang?

discordianfish:
We essentially just parse net/tcp, can you provide the relevant /proc/net/tcp content in this situation?

dswarbrick:
[...]

PR #631, opened 2024-04-09
fix: same TCP connection appears twice

weidongkl:
/proc/net/{TCP, TCP6, UDP, UDP6} are dynamic ,
and when we read these files, we should read them all at once.
there will be data consistency issues if using line by line reading

fix: https://github.com/prometheus/procfs/issues/576

### DIFF
--- a/net_ip_socket.go
+++ b/net_ip_socket.go
@@ -14,24 +14,12 @@
 package procfs
 
 import (
-	"bufio"
 	"encoding/hex"
 	"fmt"
-	"io"
 	"net"
 	"os"
 	"strconv"
 	"strings"
-)
-
-const (
-	// readLimit is used by io.LimitReader while reading the content of the
-	// /proc/net/udp{,6} files. The number of lines inside such a file is dynamic
-	// as each line represents a single used socket.
-	// In theory, the number of available sockets is 65535 (2^16 - 1) per IP.
-	// With e.g. 150 Byte per line and the maximum number of 65535,
-	// the reader needs to handle 150 Byte * 65535 =~ 10 MB for a single IP.
-	readLimit = 4294967296 // Byte -> 4 GiB
 )
 
 // This contains generic data structures for both udp and tcp sockets.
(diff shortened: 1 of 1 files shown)


weidongkl:
net-tools used  setvbuf _IOFBF to read tcp file at once too. data consistency should be more important than the increased memory usage

weidongkl:
> @SuperQ wdyt

discordianfish:
Yeah I think we'll need to do this

discordianfish (approved):

net_ip_socket.go
```diff
@@ -74,49 +62,50 @@ type (
...
-	f, err := os.Open(file)
-	if err != nil {
-		return nil, err
-	}
-	defer f.Close()
-
 	var netIPSocketSummary NetIPSocketSummary
 	var udpPacketDrops uint64
 	isUDP := strings.Contains(file, "udp")
+	content, err := os.ReadFile(file)
```
SuperQ, line 95: Would we want to use the `util.ReadFileNoStat()` function here?

SuperQ: Ahh, nevermind, this file is beyond the 1MiB file size limit of that util function.

dswarbrick:
> net-tools used setvbuf _IOFBF to read tcp file at once too. data consistency should be more important than the increased memory usage

net-tools does not read the entire file at once. It merely uses buffered IO, with a buffer of PAGE_SIZE:
```c
FILE *proc_fopen(const char *name)
{
    static char *buffer;
    static size_t pagesz;
    FILE *fd = fopen(name, "r");

    if (fd == NULL)
      return NULL;

    if (!buffer) {
      pagesz = getpagesize();
      buffer = malloc(pagesz);
    }   

    setvbuf(fd, buffer, _IOFBF, pagesz);
    return fd; 
}
```
I suspect that an equivalent approach in Go, which would have only required a single line change, would be to replace the previous `io.LimitReader` with a `bufio.NewReaderSize(f, os.Getpagesize())`.

SuperQ:
Actually, this may be worse. `os.ReadFile()` attempts to open and read into a buffer the size of the file in one read. But only if it can stat the size of the file. Otherwise it reads 512 bytes at a time.

The proc files in question have no size.
```
$ stat /proc/net/tcp
  File: /proc/net/tcp
  Size: 0         	Blocks: 0          IO Block: 1024   regular empty file
Device: 17h/23d	Inode: 4026532077  Links: 1
Access: (0444/-r--r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2024-04-16 09:48:28.233648680 +0200
Modify: 2024-04-16 09:48:28.233648680 +0200
Change: 2024-04-16 09:48:28.233648680 +0200
 Birth: -
```

So this change likely does not fix the problem and may make it worse.

I think I'm going to revert this, without proof that this fix actually works, I think the original code is actually what we want.

dswarbrick:
Also for comparison, the (deprecated) `netstat` command:
```
openat(AT_FDCWD, "/proc/net/tcp", O_RDONLY) = 3
read(3, "  sl  local_address rem_address "..., 4096) = 4050
read(3, "  26: 0601A8C0:8938 8416AB51:05A"..., 4096) = 4050
read(3, "  53: 0601A8C0:8102 182F853E:589"..., 4096) = 4050
read(3, "  80: 0601A8C0:B44A B6126749:D98"..., 4096) = 4050
read(3, " 107: 0601A8C0:8890 6D39D766:DE9"..., 4096) = 4050
read(3, " 134: 0601A8C0:A3A4 9630C658:1AE"..., 4096) = 300
read(3, "", 4096)                       = 0
close(3)
```
So if anything, the original code most closely resembles how net-tools reads /proc/net/tcp.

Now, if _both_ prometheus/procfs _and_ net-tools produce inconsistent / duplicated entries, then I would say that it's time to use rtnetlink to get this information, as iproute2 does with the `ss` (socket statistics) tool.

SuperQ:
We already switched to netlink in the node_exporter tcpstat collector.

dswarbrick:
Exactly - so should these legacy functions that parse `/proc/net/tcp{,6}` be marked as deprecated to discourage people from using them?

SuperQ:
Yea, maybe not a bad idea. @discordianfish What do you think of marking this reader as deprecated?

discordianfish:
@SuperQ yeah that makes sense

PR #634, opened 2024-04-16
Update class_thermal.go to ignore only invalid thermal zones which raise fs.PathError

miliacristian:
Sum-up: ignore invalid thermal zones to avoid discharding of all thermal zone metric

Some thermal zones generates errors fs.PathError causing the lost of all metric data also for valid thermal zone.

Example of error log: ts=2024-04-10T14:23:57.064Z caller=collector.go:169 level=error msg="collector failed" name=thermal_zone duration_seconds=0.053034792 err="read /sys/class/thermal/thermal_zone24/temp: invalid argument"

### DIFF
--- a/sysfs/class_thermal.go
+++ b/sysfs/class_thermal.go
@@ -24,6 +24,7 @@ import (
 	"syscall"
 
 	"github.com/prometheus/procfs/internal/util"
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Thanks

PR #641, opened 2024-05-23
Fix NODEV errors on Thinkpad when parsing battery status

marvin-sinister:
On thinkpads, reading `power_supply class charge_control_end_threshold` returns `no such device` error, which results in battery metrics not being exported.

Fixes: https://github.com/prometheus/node_exporter/issues/3019


SuperQ:
LGTM, although, it seems to be only some thinkpads. My X1 Carbon seems to work OK for this file.

discordianfish (approved):

PR #645, opened 2024-05-30
Return support for Go 1.20

SuperQ:
Support the last three releases to match the versions supported by prometheus client_golang.


ArthurSens (approved):
Thanks :)

PR #648, opened 2024-06-17
Fix parsing NSpids field in /proc/{PID}/status

timuralp:
The NSpids field uses a tab (\t) to separate the list of PIDs. The procfs parser should split on \t, rather than space. This was not caught because the test fixtures themselves used the space character.

The patch updates the parser, the fixtures, and adds an explicit test for NSpids.


SuperQ (approved):
Thanks!

discordianfish (approved):

PR #650, opened 2024-07-03
feat(blockdevice): added sysblockdevicesize method and test

fs185143:
In part to support https://github.com/prometheus/node_exporter/pull/3068, I have added a new method that retrieves the value of `/sys/block/<device>/size`


fs185143:
need to add `blockdevice/testdata/fixtures/sys/block/sda/size` as `3750748848`

blockdevice/stats.go
```diff
@@ -474,3 +475,9 @@ func (fs FS) SysBlockDeviceUnderlyingDevices(device string) (UnderlyingDeviceInf
 	return UnderlyingDeviceInfo{DeviceNames: underlying}, nil
 
 }
+
+// SysBlockDeviceSize returns the size of the block device from /sys/block/<device>/size.
+// Note that this needs to be multiplied by LogicalBlockSize to get the size in bytes.
```
dswarbrick, line 480: This size is [reported](https://elixir.bootlin.com/linux/v6.9.8/source/block/genhd.c#L944) in units of `sector_t`, which are 512 bytes.
```c
ssize_t part_size_show(struct device *dev,
		       struct device_attribute *attr, char *buf)
{
	return sprintf(buf, "%llu\n", bdev_nr_sectors(dev_to_bdev(dev)));
}
```
`sector_t` is [defined](https://elixir.bootlin.com/linux/v6.9.8/source/include/linux/types.h#L134) as:
```c
/**
 * The type used for indexing onto a disc or disc partition.
 *
 * Linux always considers sectors to be 512 bytes long independently
 * of the devices real block size.
 *
 * blkcnt_t is the type of the inode's block count.
 */
typedef u64 sector_t;
...
```
See [also](https://elixir.bootlin.com/linux/v6.9.8/source/include/linux/blk_types.h#L24):
```c
/*
 * The basic unit of block I/O is a sector. It is used in a number of contexts
 * in Linux (blk, bio, genhd). The size of one sector is 512 = 2**9
 * bytes. Variables of type sector_t represent an offset or size that is a
 * multiple of 512 bytes. Hence these two constants.
 */
#ifndef SECTOR_SHIFT
#define SECTOR_SHIFT 9
#endif
#ifndef SECTOR_SIZE
#define SECTOR_SIZE (1 << SECTOR_SHIFT)
#endif
```

dswarbrick: So it is possible for the logical block size to be reported as e.g. 4096 for an Advanced Format (AF) disk, but the `/sys/block/<dev>/size` will still always be reported in units of 512-byte sectors.

blockdevice/stats.go
```diff
@@ -474,3 +475,13 @@ func (fs FS) SysBlockDeviceUnderlyingDevices(device string) (UnderlyingDeviceInf
...
 }
+
+// SysBlockDeviceSizeBytes returns the size of the block device from /sys/block/<device>/size
+// in bytes by multiplying the value by the Linux sector length of 512.
+func (fs FS) SysBlockDeviceSizeBytes(device string) (uint64, error) {
+	size, err := util.ReadUintFromFile(fs.sys.Path(sysBlockPath, device, sysBlockSize))
+	if err != nil {
+		return 0, err
+	}
+	return 512 * size, nil
```
discordianfish, line 486: Let's make this a const SectorSize

fs185143: https://github.com/prometheus/procfs/pull/650/commits/b70d052f2b0a97e986fedba7bd6260931f3e31f6

fs185143:
just following up on this cc @discordianfish

fs185143:
> just following up on this cc @discordianfish

or cc @SuperQ

fs185143:
hey could I get a review on this please? any idea who might be best to take a look @dswarbrick

dswarbrick:
@fs185143 You'll have better luck attracting the attention of maintainers if all CI tests are green, plus all commits need to be signed off to comply with the DCO.

fs185143:
closing in favour of https://github.com/prometheus/procfs/pull/658 (fixed issues like DCO)

PR #658, opened 2024-08-14
feat(blockdevice): added sysblockdevicesize method and test

fs185143:
In part to support https://github.com/prometheus/node_exporter/pull/3068, I have added a new method that retrieves the value of `/sys/block/<device>/size`

### DIFF
--- a/blockdevice/stats.go
+++ b/blockdevice/stats.go
@@ -209,6 +209,8 @@ const (
 	sysBlockQueue       = "queue"
 	sysBlockDM          = "dm"
 	sysUnderlyingDev    = "slaves"
+	sysBlockSize        = "size"
+	sectorSize          = 512
 )
 
 // FS represents the pseudo-filesystems proc and sys, which provides an
@@ -473,4 +475,14 @@ func (fs FS) SysBlockDeviceUnderlyingDevices(device string) (UnderlyingDeviceInf
 	}
 	return UnderlyingDeviceInfo{DeviceNames: underlying}, nil
 
+}
+
+// SysBlockDeviceSize returns the size of the block device from /sys/block/<device>/size.
+// in bytes by multiplying the value by the Linux sector length of 512.
+func (fs FS) SysBlockDeviceSize(device string) (uint64, error) {
+	size, err := util.ReadUintFromFile(fs.sys.Path(sysBlockPath, device, sysBlockSize))
+	if err != nil {
+		return 0, err
+	}
+	return sectorSize * size, nil
 }
--- a/blockdevice/stats.go
+++ b/blockdevice/stats.go
@@ -21,7 +21,8 @@ import (
 	"os"
 	"strings"
 
-	filesystem "github.com/prometheus/procfs/internal/fs"
+	"github.com/prometheus/procfs"
+	"github.com/prometheus/procfs/internal/fs"
 	"github.com/prometheus/procfs/internal/util"
 )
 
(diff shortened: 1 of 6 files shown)


fs185143:
fixed DCO and tests with this new PR

cc @dswarbrick

blockdevice/stats.go
```diff
@@ -209,6 +209,8 @@ const (
 	sysBlockQueue       = "queue"
 	sysBlockDM          = "dm"
 	sysUnderlyingDev    = "slaves"
+	sysBlockSize        = "size"
+	sectorSize          = 512
```
discordianfish, line 213: don't we have this somewhere else already?

discordianfish (approved):
LGtM in general though

blockdevice/stats.go
dswarbrick: The procfs/btrfs package defines it locally (albeit with a slightly misleading comment regarding _block_ size).
```go
// SectorSize contains the Linux sector size.
// > Linux always considers sectors to be 512 bytes long independently
// > of the devices real block size.
const SectorSize = 512
```
This is how the Linux kernel defines it (for reference):
```c
/*
 * The basic unit of block I/O is a sector. It is used in a number of contexts
 * in Linux (blk, bio, genhd). The size of one sector is 512 = 2**9
 * bytes. Variables of type sector_t represent an offset or size that is a
 * multiple of 512 bytes. Hence these two constants.
 */
#ifndef SECTOR_SHIFT
#define SECTOR_SHIFT 9
#endif
#ifndef SECTOR_SIZE
#define SECTOR_SIZE (1 << SECTOR_SHIFT)
#endif
```

fs185143: fixed in https://github.com/prometheus/procfs/pull/658/commits/900fc3ab8db0827d4ce035aeab01233037486b6b

dswarbrick: Perhaps I should have been clearer. I was not suggesting to use the constant from the btrfs package; merely pointing out where it currently exists.

Since this 512 byte sector size is definitely not btrfs-specific, the constant should be defined somewhere centrally (and ideally the btrfs code should import it from there also).

fs185143: https://github.com/prometheus/procfs/pull/658/commits/059cb7843a05f109c8a92d02f83180f165e69c91

dswarbrick: I'm not sure that the internal/fs package is the appropriate place for this constant either, given what the comment for FS says:
```
// FS represents a pseudo-filesystem, normally /proc or /sys, which provides an
// interface to kernel data structures.
```
The internal/fs package is basically just a wrapper for accessing items in procfs, sysfs and its siblings, given a specific base path.

The 512-byte sector size is also independent of filesystems in the Linux kernel. It is a basic, low-level size unit used by all `bio` operations in the kernel, which includes IO operations on raw block devices which have _no_ filesystem at all (e.g. swap partitions, mdraid).

Perhaps a little more discussion would be fruitful (especially with the maintainers), before throwing any more code at this PR.

fs185143: ok sure - @discordianfish any thoughts on the above?

fs185143: do we not think it belongs in `blockdevice/stats.go` or a new `blockdevice/constants.go`? the code you referenced from the linux kernel is in `linux/include/linux/blk_types.h`
> Block data types and constants.

discordianfish: Maybe we should just put it in <root>/fs.go - somewhere at the root level seem suitable given it's used at least by blockdevice/ and btrs/ @dswarbrick @SuperQ wdyt?

dswarbrick: I guess the main thing is that it is defined **once** for the whole project, and not duplicated in multiple places. Beyond that, I suppose it doesn't matter _that_ much where it goes. I think it's quite likely that we'll accumulate a few more constants like these that we borrow from Linux kernel header includes, assuming that they are not already defined somewhere in `os` or `unix` packages in the standard library.

fs185143: done

dswarbrick (approved):

fs185143:
@discordianfish mind taking another look at this please? thanks :smile:

discordianfish (approved):

fs185143:
@discordianfish thanks :) didn't seem to get merged automatically? I merged master in in case that was the issue, but now says a workflow is awaiting approval.

discordianfish:
@SuperQ needs to review this first

fs185143:
@SuperQ mind taking a look please?

SuperQ (approved):
Thanks!

SuperQ:
Would you please rebase this PR, it seems like the git history is corrupted.

fs185143:
> Would you please rebase this PR, it seems like the git history is corrupted.

done

fs185143:
@SuperQ is there anything else that needs to be done to merge this on my side?

SuperQ:
Looks great, thanks!

PR #659, opened 2024-08-22
Parse StartCode, EndCode, and StartStack in `Proc.Stat()`

pgimalac:
Add the following fields to the `ProcStat` type and read them from `/proc/<pid>/stat` in `Proc.Stat()`:
- StartCode
- EndCode
- StartStack

See fields 26, 27, and 28 in https://man7.org/linux/man-pages/man5/proc_pid_stat.5.html.

It enables categorizing regions from `/proc/<pid>/maps` depending on their address (ie. actual code vs something else, eg. memory mapped file).

### DIFF
--- a/proc_stat.go
+++ b/proc_stat.go
@@ -101,6 +101,12 @@ type ProcStat struct {
 	RSS int
 	// Soft limit in bytes on the rss of the process.
 	RSSLimit uint64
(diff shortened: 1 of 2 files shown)


discordianfish (approved):
Thanks!

SuperQ (approved):

PR #664, opened 2024-09-12
fix some fc-hba are missing the statistics/*

chijiajian:
Some FC-HBA cards are missing the /sys/class/fc_host//statistics/* directory and files, for example, the Huawei Technologies Co., Ltd. Hi1822 Family (2*16G FC) (rev 45) card with ID 3e:00.0. However, there are corresponding port_name, port_state, etc., under /sys/class/fc_host//. It is recommended to add a check in the parseFibreChannelStatistics function in the class_fibrechannel.


sysfs/class_fibrechannel.go
```diff
@@ -153,6 +153,10 @@ func parseFibreChannelStatistics(hostPath string) (*FibreChannelCounters, error)
 	var counters FibreChannelCounters
 
 	path := filepath.Join(hostPath, "statistics")
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		fmt.Printf("Directory %s does not exist, skipping.\n", path)
```
SuperQ, line 157: Please do not put `Printf` statements in this library.

SuperQ (changes requested):

sysfs/class_fibrechannel.go
```diff
@@ -153,6 +153,9 @@ func parseFibreChannelStatistics(hostPath string) (*FibreChannelCounters, error)
 	var counters FibreChannelCounters
 
 	path := filepath.Join(hostPath, "statistics")
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return &counters, nil
+	}
 	files, err := os.ReadDir(path)
 	if err != nil {
```
discordianfish, line 160: Why not check if the error is IsNotExist?

chijiajian: I think it's better to just check if the statistics directory exists, instead of worrying about other errors like permission or filesystem issues. If the directory isn't there, the  function can keep going instead of stopping for fibrechannel statistics not found

discordianfish (approved):

sysfs/class_fibrechannel.go
chijiajian: I have removed the corresponding Printf statements. Please review again. Thank you!

chijiajian (commented):
Fixed — now explicitly checking for fs.ErrNotExist and returning other errors.

chijiajian (commented):
fix: undefined: fs

chijiajian (commented):
Updated to check errors.Is(err, os.ErrNotExist). Thanks!

chijiajian:
Hi @SuperQ @discordianfish, I addressed the last review comment in a3c8569 by removing the redundant os.Stat call and handling os.ErrNotExist directly from os.ReadDir.

  The review threads are now resolved/outdated. Could you please re-run CI for the latest commit and take another look when you have a chance? Thanks!

SuperQ (approved):

sysfs/class_fibrechannel.go
```diff
@@ -154,6 +154,10 @@ func parseFibreChannelStatistics(hostPath string) (*FibreChannelCounters, error)
 	path := filepath.Join(hostPath, "statistics")
 	files, err := os.ReadDir(path)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			// statistics directory not present, skip
```
SuperQ:
```suggestion
```

PR #668, opened 2024-09-22
Forbid print statements

SuperQ:
Enable `forbidigo` to make sure print statements are not used in this library. This avoids injecting random log output into upstream user's log formatting.


kakkoyun (approved):
LGTM

PR #672, opened 2024-10-02
Add chip to the DRM parser class

Deezzir:
Adds a chip label to the `DRM` parser class so the data can be related to `hwmon` metrics when exported by the [node_exporter](https://github.com/prometheus/node_exporter).

For example, `node_exporter` produces the following metric:

```
node_hwmon_chip_names{agent_hostname="ubuntu-0", chip="0000:9d:00_0_0000:9e:00_0", chip_name="amdgpu", instance="ubuntu-0.maas"}
``` 

So, by having a `chip` label coming from the DRM parser, DRM metrics can be related to hwmon sensor metrics to build better dashboard with clear AMD GPU filters.


Deezzir:
@discordianfish and @SuperQ please review! Thank you

sysfs/class_drm_amdgpu.go
```diff
@@ -47,6 +52,38 @@ type ClassDRMCardAMDGPUStats struct {
 	MemoryVRAMVendor              string // The VRAM vendor name.
 	PowerDPMForcePerformanceLevel string // The current power performance level.
 	UniqueID                      string // The unique ID of the GPU that will persist from machine to machine.
+	HWmonChip                     string // The hwmon chip.
+}
+
+func cleanChipName(name string) string {
```
discordianfish, line 58: Why do we need to sanitize the name?

Deezzir: It's something that the hwmon does with the chip names as well, so the format is similar. 

See: https://github.com/prometheus/node_exporter/blob/e6a9cfbdcdaa21bf9676c6cd37bef8160227f423/collector/hwmon_linux.go#L74

discordianfish: Yes but in that case its used to construct the metric name, which has more limitations. Here its being used as label value, where I don't think we need this

Deezzir: Do you think it will still be the same and relatable by the chip from the hwmon? It seems like a good idea to align the result to what the hwmon does with its chip names

discordianfish: I guess in the node-exporter we should also just use the raw chip name and put it in a label instead of the metric name. But I think either way, here, this being used as label value, it should not get santized. But want tohave @SuperQ opinion on this as well.

Deezzir: A clarification - I don't care if it is sanitized or not; the only reason I am doing a sanitization here is to align the label `chip` coming from the hwmon metrics (which is also sanitized as per the permalink I provided above) to a new label I've added to the `node_drm_card_info` metric. 

I want to match, aka join a metric like `node_hwmon_fan_rpm` with `node_drm_card_info`  and have a connection between the two collectors. RN, there is no way to match the metrics coming from those collectors, although both provide metrics for GPUs. 

Sanitizing the label before the match makes sense because this is what's being done for the `chip` label in the hwmon collector. 

If you think we shouldn't sanitize the hwmon chip label in the node_exporter, I might provide a PR for that. As of now, I am aligning the behaviour with the existing solution.

discordianfish: Yeah makes sense, let's see what @SuperQ is thinking about this. I'm fine either way but moving chip to a label in the hwmon collector makes most sense to me.

Deezzir: @SuperQ Please take a look

Deezzir:
@discordianfish PTAL

sysfs/class_drm_amdgpu.go
discordianfish: I think we should move the chip name to a label in the hwmon collector.
Either way, we shouldn't sanitize the chip name here but in the node-exporter (if we decide to go that path)

discordianfish:
@Deezzir See my comment, sorry for the delay..

Deezzir:
@discordianfish If I understand your comments correctly, I should leave the logic of extracting the `devName` and `devType` and export those without sanitizing and creating a `chipName`. The logic for that should go into the `node_exporter`?

Deezzir:
@discordianfish can you please take a look again at the latest comment, so we can finalize the design?

Deezzir:
@discordianfish I've updated the PR to have unsatitized fields, PTAL

sysfs/class_drm_amdgpu.go
```diff
@@ -65,8 +81,10 @@ func (fs FS) ClassDRMCardAMDGPUStats() ([]ClassDRMCardAMDGPUStats, error) {
 			}
 			return nil, err
 		}
-		cardStats.Name = filepath.Base(card)
-		stats = append(stats, cardStats)
+		if cardStats != (ClassDRMCardAMDGPUStats{}) {
```
discordianfish: Why?

sysfs/class_drm_amdgpu.go
```diff
@@ -117,6 +135,10 @@ func parseClassDRMAMDGPUCard(card string) (ClassDRMCardAMDGPUStats, error) {
 	if v, err := readDRMCardField(card, "unique_id"); err == nil {
 		stats.UniqueID = v
 	}
+	if n, t, err := readDevInfo(card); err == nil {
```
discordianfish: Should we silently fail here?

sysfs/class_drm_amdgpu.go
Deezzir: Please see https://github.com/prometheus/procfs/issues/673

sysfs/class_drm_amdgpu.go
Deezzir: You mean something like this in case we couldn't retrieve the name & type?

```go
if err != nil {
    return ClassDRMCardAMDGPUStats{}, err
}
```

discordianfish: Yes

sysfs/class_drm_amdgpu.go
Deezzir: The line checks if the struct is empty (zero-initialized) and only appends to the `stats` list if it is not empty. I've also modified the test file for the module. Previously, the test would produce two struct instances: one valid and one with all elements zero-initialized. It doesn't make much sense to let the empty object be passed as a valid one.

Deezzir:
@discordianfish PTAL

Deezzir:
Thanks @discordianfish and @SuperQ, continuing in https://github.com/prometheus/node_exporter/pull/3145

PR #675, opened 2024-10-15
Add broadcom bnxt re support

antonycleave:
This adds all the exposed hw_counters metrics for the bnxt_re infiniband driver that does not also expose functional counters
e.g:
# cat /sys/class/infiniband/bnxt_re0/ports/1/counters/port_rcv_data
cat: /sys/class/infiniband/bnxt_re0/ports/1/counters/port_rcv_data: Invalid argument
# cat /sys/class/infiniband/bnxt_re0/ports/1/hw_counters/rx_bytes
7521781137064


antonycleave:
@discordianfish is there any reason why the hw_counters in this PR should not be added? it would make my life a lot easier to be able to pull from here.

Thanks!

PR #678, opened 2024-10-18
infiniband: do not make assumptions about counters based on HCA name

dswarbrick:
Some users have reported cases of systemd "predictable network interface naming" apparently also renaming the HCA device. This means we can no longer make assumptions about which counter(s) should be present based on the HCA name (i.e., `irdma*`, `mlx5_*`). The previous approach was quite brittle anyway, since there will undoubtedly be other IB / RoCE drivers in future which implement the` hw_counters` directory (but not the older `counters` directory).


discordianfish (approved):
Makes sense, LGTM

SuperQ (approved):

SuperQ:
Sorry, had to fix the commit since I didn't notice this before merging #687. This is a better solution.

PR #679, opened 2024-10-22
feat(mdstat): recognize reshape status

tamcore:
This change will properly catch raids that are currently undergoing conversion (reshaping), i.e. from RAID1 to RAID5 as in my case.


discordianfish (approved):
Thanks!

PR #686, opened 2024-12-09
feat: Read PCIE AER counters class/net

dasturiasArista:
Linux provides AER counters in the path /sys/class/net/<iface>/device/

This is split amoung 3 different files:
aer_dev_correctable
aer_dev_fatal
aer_dev_nonfatal

I would like to merge up this functionality since I would like to extend the node exporter to be able to report AER statistics and I wanted to follow the same pattern as the other class/net counters


dasturiasArista:
https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-pci-devices-aer_stats has more information about these counters

SuperQ (approved):
Great, thanks!

PR #687, opened 2024-12-12
Do not collect InfiniBand counters for Broadcom RoCE NICs

oneoneonepig:
Broadcom RoCE NICs (bnxt_re#) does not have the counter metrics exposed in `/sys/class/infiniband/<device>/ports/<port-num>/counter`.

```shell
$ strace -e trace=open,openat node_exporter --log.level=debug
...
1270794 openat(AT_FDCWD, "/sys/class/infiniband/bnxt_re0/ports/1/counters", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
...

$ ls /sys/class/infiniband/bnxt_re0/ports/1/counters
ls: cannot access '/sys/class/infiniband/bnxt_re0/ports/1/counters': No such file or directory
```

This will cause Node Exporter to never collect any InfiniBand metrics since it always return with error ENOENT

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -279,8 +279,8 @@ func (fs FS) parseInfiniBandPort(name string, port string) (*InfiniBandPort, err
 		return nil, fmt.Errorf("could not parse rate file in %q: %w", portPath, err)
 	}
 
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Thanks

PR #692, opened 2025-01-06
Build CPUInfo on Darwin

localhots:
`cpuinfo.go` is currently only built on Linux and it breaks Darwin builds for no apparent reason. 
I understand that there's no procfs on Darwin but it should still be possible to use macOS for development of projects that use this package.

This change allows it to be built on Darwin with a dummy `parseCPUInfo` implementation.

Please let me know if I need to change this solution in any way. Thanks!

### DIFF
--- a/cpuinfo.go
+++ b/cpuinfo.go
@@ -11,9 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build linux
-// +build linux
-
 package procfs
 
 import (
--- a/cpuinfo_darwin.go
+++ b/cpuinfo_darwin.go
@@ -1,0 +1,19 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


cpuinfo_darwin.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2020 The Prometheus Authors
```
SuperQ, line 1:
```suggestion
// Copyright 2025 The Prometheus Authors
```

cpuinfo.go
```diff
@@ -11,9 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build linux
-// +build linux
```
SuperQ, line 15: Removing this will likely break other non-linux non-darwin OSs like BSDs, etc.

SuperQ (changes requested):
I'm not sure this is the correct fix, it would likely break other OSs.

cpuinfo.go
localhots: I was worried about that too but couldn't find a better solution. Will you be able to suggest better build tags?

PR #693, opened 2025-01-09
Add support for per-interface SNMP6 stats

Vascko:
Hi Folks,

I'd love to extend the available IPv6 metrics in Prometheus' Node Exporter to include the per-interface stats from the /proc/net/dev_snmp6/ folder, so I started by adding the ability to get that data via this library.

Tests for the new methods as well as new testdata are included.

Fixed some old references to the fixture path in the docs as well.


net_dev_snmp6.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2018 The Prometheus Authors
```
discordianfish, line 1: 2018->2025

discordianfish (approved):
LGTM

discordianfish:
Also see linting error

net_dev_snmp6.go
Vascko: Thank you, fixed!

discordianfish:
Thanks!

PR #694, opened 2025-01-13
Dont check for temperatures in /thermal_zone where mode=disabled

Saumya40-codes:
### Relevant issue

Fixes: https://github.com/prometheus/node_exporter/issues/2980

### Current behaviour

- `node_exporter` checks for the contents of `temp` file even when `mode` is **disabled** due to which in the case when temp file contains a non desired  value the `thermal_zone` collector fails and no longer gives its related metrics 

### Proposed solution

- Have an early exit and dont check rest if `mode` is set to disabled

### Testing

- replacing `github.com/prometheus/procfs` in `node_exporter` to local dev setup of `procfs` and comparing the results in both

### DIFF
--- a/sysfs/class_thermal.go
+++ b/sysfs/class_thermal.go
@@ -63,6 +63,19 @@ func (fs FS) ClassThermalZoneStats() ([]ClassThermalZoneStats, error) {
 }
 
 func parseClassThermalZone(zone string) (ClassThermalZoneStats, error) {
+
+	// Optional attributes.
+	modeContent, err := util.SysReadFile(filepath.Join(zone, "mode"))
+	if err != nil && !os.IsNotExist(err) && !os.IsPermission(err) {
+		return ClassThermalZoneStats{}, err
+	}
+
+	if strings.TrimSpace(modeContent) == "disabled" {
+		return ClassThermalZoneStats{}, syscall.ENODATA
+	}
+
+	zoneMode := util.ParseBool(modeContent)
+
 	// Required attributes.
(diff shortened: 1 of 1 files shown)


sysfs/class_thermal.go
```diff
@@ -63,6 +63,19 @@ func (fs FS) ClassThermalZoneStats() ([]ClassThermalZoneStats, error) {
...
 
 func parseClassThermalZone(zone string) (ClassThermalZoneStats, error) {
+
+	// Optional attributes.
+	modeContent, err := util.SysReadFile(filepath.Join(zone, "mode"))
+	if err != nil && !os.IsNotExist(err) && !os.IsPermission(err) {
+		return ClassThermalZoneStats{}, err
+	}
+
+	if strings.TrimSpace(modeContent) == "disabled" {
```
SuperQ, line 73: Typically we avoid making decisions like this in the library. Rather, we simply expose the `mode` as a raw attribute and allow the downstream consumer to decide what to do.

Saumya40-codes: Ohkk

Maybe thats something which can be handled from `node exporter` itself only? or as stated we check while reading the `Zone temp` and if the error turns out to be of type `os.ErrInvalid` then we can return an empty struct  (and return an error when necessary (i.e. in the case of `os.IsNotExist(err)` && `os.IsPermission(err)`

SuperQ: Yes, we may need to refactor that error handling.

PR #695, opened 2025-01-28
Nvidia/Mellanox expose ROCE ECN information on sysfs on the path

dasturiasArista:
Nvidia/Mellanox expose ROCE ECN information on sysfs on the path
/sys/class/net/\<interface\>/ecn/\<protocol\>/

There are 2 protocols Reaction Point (rp) and Notification point (np)

For each of the protocols they have a list of attributes:
/sys/class/net/\<interface\>/ecn/\<protocol\>/params/\<requested attribute\>

Each protocol will also if ECN is enabled per priority (where X is the
priority):
/sys/class/net/\<interface\>/ecn/\<protocol\>/enable/X

This is documented here
https://docs.nvidia.com/networking/display/mlnxofedv571020/explicit+congestion+notification+(ecn)

The attributes are documented here:
https://enterprise-support.nvidia.com/s/article/dcqcn-parameters

### DIFF
--- a/sysfs/net_class_ecn.go
+++ b/sysfs/net_class_ecn.go
@@ -1,0 +1,387 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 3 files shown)


discordianfish:
LGTM in general but it makes me wonder where to draw the line what vendor specific stuff to include here and what not.. I feel this is probably relevant enough to be included but not sure. @SuperQ wdyt?

dasturiasArista:
Just to be clear this is just nvidia specific because other vendors aren't exposing these values through Sysfs. these are pretty generic values for ROCEv2. For example https://docs.broadcom.com/doc/NCC-WP1XX has information on broadcom's congestion control, which also uses ECN. Ideally other vendors would also expose these values in a sysfs path instead of relying on propriety utilities. I'm not aware if other vendors that implement rocev2 plan to in the future integrate with sysfs (if they already have), but this seems fairly generic in terms of ROCEv2

discordianfish:
@dasturiasArista Thanks for clarifying. I think this is a good argument for including it here.

PR #700, opened 2025-02-26
sysfs: add link_layer property to InfiniBandPort

thomasbarrett:
Common IB HCA models (e.g. Mellanox ConnectX 7) can operate in either InfiniBand mode or Ethernet mode. When operating in Ethernet mode, they are still detected as InfiniBand cards, but look unhealthy. The `link_layer` property can be used to determine which ports are InfiniBand vs Ethernet.


sysfs/class_infiniband.go
```diff
@@ -107,6 +107,7 @@ type InfiniBandHwCounters struct {
 // /sys/class/infiniband/<Name>/ports/<Port>
 // for a single port of one InfiniBand device.
 type InfiniBandPort struct {
+	LinkLayer   string
 	Name        string
 	Port        uint
```
SuperQ, line 112:
```suggestion
	Name        string
	Port        uint
	LinkLayer   string // String representation from /sys/class/infiniband/<Name>/ports/<Port>/link_layer
```

SuperQ (changes requested):
Minor nit, otherwise good to go.

PR #711, opened 2025-03-26
add SysBlockDeviceQueueStat

wjordan:
SysBlockDeviceQueueStat returns a single stat from `/sys/block/xxx/queue` where `xxx` is a device name and `stat` is a file. Allow fetching a single queue stat without parsing the entire queue sysfs structure.

See: https://github.com/prometheus/node_exporter/issues/3282#issuecomment-2755706786


discordianfish (commented):
Needs sign off but lgtm otherwise

SuperQ (changes requested):

Issue #698, opened 2025-02-19
Error Handling for Offline Thermal Zones


bnbhat:
## Description: 

When  `procfs` try to read the offline thermal zones, it gets stuck. please check this [issue](https://github.com/prometheus/node_exporter/issues/3071) for more details.

Device: 
- Jetson Orin Nano
- Linux 5.15.136-tegra
- node_exporter version 1.9.0

for ex, when it tries to read thermal_zone2, it gets stuck.
```bash
$ cat /sys/class/thermal/thermal_zone2/temp
cat: /sys/class/thermal/thermal_zone2/temp: Resource temporarily unavailable
```
It would be helpful if it could handle this scenario more gracefully.  

I have attempted to trace the error, here is the backtrace:

![Image](https://github.com/user-attachments/assets/18f3d445-f726-4174-a0ac-da56c58cfd20)

neoLsH:
On the procfs side, offline/unreadable thermal zones no longer fail the whole read: `parseClassThermalZone` collects per-attribute read failures into the `ReadErrors` field of `ClassThermalZoneStats` (via `errors.Join`), which came out of the "skip failed zones" work in #712 and the error-handling refactor in #794. Callers can then decide how to handle a zone that reported errors.

The remaining gap was on the consumer side: node_exporter's `thermal_zone` collector ignored `ReadErrors` and unconditionally exported `Temp`, so a zone whose attributes failed to read could show up as a bogus `0°C`. I've opened prometheus/node_exporter#3807 to skip zones that come back with a non-nil `ReadErrors`.

So I believe this is effectively addressed by #712/#794 here plus node_exporter#3807 — please correct me if there's still procfs-side behavior that's missing.

PR #712, opened 2025-03-28
Use SysReadFile for thermals and skip failed zones

kitschysynq:
Switches all calls in the thermal zones processing to use SysReadFile so
that unavailable zones won't stall forever waiting for them to become
available. Also continues processing in the case that one zone fails
with EAGAIN, as is the case for temporarily unavailable zones.

Addresses #698


SuperQ (approved):

SuperQ:
Thanks!

PR #720, opened 2025-04-25
chore: clean up golangci-lint configuration

mmorel-35:
#### Description

chore: clean up golangci-lint configuration


kakkoyun (commented):
While I don't see any harm on these changes, but I also don't understand why we need these?

This file was generated from the tool itself as a result of running the migrate command.

PR #731, opened 2025-06-17
nvme: Add ControllerID output

ShashwatHiregoudar:
exposing the ControllerID along with the existing parameters


SuperQ (approved):
Thanks!

PR #674, opened 2024-10-03
feat: expose MD raid component devices

robbat2:
Expose what component devices are part of a MD raid device, as well as the most common flags per-component. This will enable a future node_exporter metric showing which component of a RAID had failed.

This PR builds up top of PR 505, implementing the requested changes to detect the known RAID types on a given system.

Reference: https://github.com/prometheus/procfs/pull/505
Signed-off-by: Robin Johnson <robbat2@orbis-terrarum.net>

PR #737, opened 2025-07-13
Added stat for disks in "replacing" state

Finomosec:
When a disk is replaced using the command `mdadm /dev/md0 --replace /dev/sdb1 --with /dev/sde1` the disk ist marked with `(R)`.

This information is now exported as count DisksReplacing.

### DIFF
--- a/blockdevice/stats.go
+++ b/blockdevice/stats.go
@@ -21,10 +21,9 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


SuperQ:
This needs a rebase.

SuperQ:
I think this is solved by `MDStatComponent` in https://github.com/prometheus/procfs/pull/674.

PR #745, opened 2025-09-03
Out of range panic fix

artemseleznev:
The Svace static analyzer (https://www.ispras.ru/en/technologies/svace/) flagged this code as suspicious: it could lead to a panic (out of range) if len(out) == 0. It might happen if text would be e.g. "bla:bla"


SuperQ (commented):
There's definitely a corrupt input panic here. But do want to continue or error?

SuperQ (changes requested):
This should return an error, not continue.

SuperQ:
Fixed in #830

PR #755, opened 2025-10-17
Update supported Go versions.

SuperQ:
* Update minimum supported version to 1.24.0.
* Update CI tests for latest Go versions.
* Run Go modernize.


net_sockstat.go
```diff
@@ -141,7 +141,6 @@ func parseSockstatProtocol(kvs map[string]int) NetSockstatProtocol {
 	for k, v := range kvs {
 		// Capture the range variable to ensure we get unique pointers for
 		// each of the optional fields.
```
cristiangreco: I think this comment can go away now?

PR #763, opened 2025-11-07
Update sysfs/class_thermal: continue on EINVAL in parseClassThermalZone to ignore only invalid thermal zones which raise "invalid argument"

ccastiglione-reply:
Sum-up: ignore invalid thermal zones to avoid discharding of all thermal zone metric.

Reading sysfs entries in the thermal zone may return EINVAL if the underlying data is not yet available or is in an inconsistent state. Continue on EINVAL allows metrics collection to continue successfully for other valid thermal zones.

Example of error log: time=2025-11-05T15:26:06.536+01:00 level=ERROR source=collector.go:168 msg="collector failed" name=thermal_zone duration_seconds=0.171564844 err="invalid argument"


discordianfish:
@ccastiglione-reply Needs DCO sign off

SuperQ (approved):

PR #786, opened 2026-02-24
fix: handle 'delayed' state in sync_completed parsing

### LINKED ISSUE #770: sysfs.Mdraids() fails if any MD raid array has a delayed check or resync
If you have multiple software RAID arrays that share physical devices and you trigger a check or a resync for more than one of them, the Linux kernel MD layer notices this situation and normally delays the action on all but one device. For example:

````
Personalities : [raid1] [linear] [multipath] [raid0] [raid6] [raid5] [raid4] [raid10] 
md0 : active raid1 sdb2[0] sda2[1]
      972946432 blocks super 1.2 [2/2] [UU]
      [=========>...........]  check = 45.5% (442910144/972946432) finish=55.6min speed=158752K/sec
      bitmap: 2/8 pages [8KB], 65536KB chunk

md1 : active raid1 sdb3[0] sda3[1]
      3677184 blocks super 1.2 [2/2] [UU]
        resync=DELAYED
      
unused devices: <none>
````
When this is happening, /sys/block/md*/md/sync_completed will contain 'delayed' for any delayed arrays. This causes sysfs.Mdraids() to incorrectly fail, as the code assumes that sync_completed will contain either 'none' or '%d / %d', and attempting to parse 'completed' as the latter returns an 'expected integer' error.

### PR DESCRIPTION
**Pull request by mehrdadbn9**
## Summary
- Handle `delayed` as a special value in sync_completed parsing (similar to `none`)
- When multiple MD RAID arrays share physical devices and a check/resync is triggered for more than one, the kernel delays the action on all but one device
- In this case, `sync_completed` contains `delayed` instead of `none` or `N / M`

## Problem
The current code fails when `sync_completed` contains `delayed`:
```
error parsing mdraids: expected integer
```

## Solution
Add `delayed` to the list of special values that should not be parsed as integers.

## Testing
- Tested with manual test cases: `none`, `delayed`, `123 / 456`, `invalid`
- All cases handled correctly

Fixes #770
Related: prometheus/node_exporter#3500

### DIFF
--- a/sysfs/mdraid.go
+++ b/sysfs/mdraid.go
@@ -166,7 +166,7 @@ func (fs FS) Mdraids() ([]Mdraid, error) {
 			}
 
 			if val, err := util.SysReadFile(filepath.Join(path, "sync_completed")); err == nil {
-				if val != "none" {
+				if val != "none" && val != "delayed" {
 					var a, b uint64
 
 					// File contains two values representing the fraction of number of completed
--- a/sysfs/mdraid_test.go
+++ b/sysfs/mdraid_test.go
@@ -155,3 +155,26 @@ func TestMdraidStats(t *testing.T) {
 		t.Fatalf("unexpected Mdraid (-want +got):\n%s", diff)
 	}
 }
+func TestMdraidDelayedSyncCompleted(t *testing.T) {
+	fs, err := NewFS("testdata/fixtures/sys/issue770")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	mdraids, err := fs.Mdraids()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(mdraids) != 1 {
+		t.Fatalf("expected 1 mdraid, got %d", len(mdraids))
+	}
+
+	md := mdraids[0]
+	if md.SyncCompleted != 0 {
+		t.Fatalf("expected SyncCompleted=0 for 'delayed' state, got %f", md.SyncCompleted)
+	}
+	if md.SyncAction != "resync" {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by SuperQ — CHANGES_REQUESTED** [review]

File: sysfs/mdraid.go [hunk 1]
```diff
@@ -136,7 +136,7 @@ func (fs FS) Mdraids() ([]Mdraid, error) {
 			}
 
 			if val, err := util.SysReadFile(filepath.Join(path, "sync_completed")); err == nil {
-				if val != "none" {
```
dswarbrick: Even though "none" and "delayed" are currently the only two special cases (as of kernel v6.19.10), I wonder if this would be more robust if we performed a regex match, or flipped the logic around the `fmt.Sscanf` so that it would handle future special cases that don't match the `%d / %d` format.

mehrdadbn9:
@SuperQ DCO sign-off added (single squashed commit, `Signed-off-by` present and DCO check now passes).

While reworking this I also found and fixed a real defect in the PR itself: the regression test referenced `testdata/fixtures/sys/issue770` but the fixture had been committed at the wrong path (`testdata/issue770/...`), so `TestMdraidDelayedSyncCompleted` could never pass. I moved the fixture into the standard `testdata/fixtures.ttar` archive (the procfs fixture mechanism) and regenerated it. `TestMdraidDelayedSyncCompleted` and `TestMdraidStats` now pass.

Force-pushed `68b0e29`. The change handles `delayed` in `sync_completed` (mirrors `none`) so multi-array MD RAID resync no longer errors with `expected integer`. Fixes #770.

PR #824, opened 2026-06-17
blockdevice: add SysBlockDeviceRotational for targeted rotational reads

ruthwikkakumani:
## What

Add `FS.SysBlockDeviceRotational(device string) (uint64, error)` — a targeted
single-file reader for `/sys/block/<device>/queue/rotational`.

## Why

Required by `node_exporter`'s diskstats collector to fix a **10× scrape-latency
regression** on systems with many block devices
(prometheus/node_exporter#3282).

The previous implementation called `SysBlockDeviceQueueStats()` which reads
**~30 sysfs files per device** to populate a full `BlockQueueStats` struct,
consuming only the `Rotational` field. On systems with 1,000+ devices this
caused 30,000+ unnecessary file reads per scrape.

`SysBlockDeviceRotational` reduces this to **exactly 1 read per device**.

## How

Follows the same single-file read pattern as `SysBlockDeviceSize()` using
`util.ReadUintFromFile`. Returns `1` for rotational (HDD), `0` for
non-rotational (SSD/NVMe).

A unit test (`TestSysBlockDeviceRotational`) is included, using the existing
`sda` fixture (rotational=1) and verifying error behaviour for devices without
a `queue/rotational` file.

## Related

- node_exporter PR: prometheus/node_exporter#3686
- Regression issue: prometheus/node_exporter#3282

### DIFF
--- a/blockdevice/stats.go
+++ b/blockdevice/stats.go
@@ -492,6 +492,15 @@ func (fs FS) SysBlockDeviceSize(device string) (uint64, error) {
 	return procfs.SectorSize * size, nil
 }
 
+// SysBlockDeviceRotational returns the rotational value for the block device
+// from /sys/block/<device>/queue/rotational.
+// A value of 1 indicates a rotational device (HDD), 0 indicates a
+// non-rotational device (SSD, NVMe). An error is returned if the file
+// cannot be read or does not contain a valid integer.
+func (fs FS) SysBlockDeviceRotational(device string) (uint64, error) {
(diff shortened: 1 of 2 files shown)


SuperQ (approved):

ruthwikkakumani:
@SuperQ thanks for the approval! Could you also approve the CI workflow runs so the checks can complete? They're currently blocked on "2 workflows awaiting approval". Thank you!

PR #830, opened 2026-06-23
Fix crypto parsing panic

SuperQ:
Make sure that we don't panic when parsing invalid crypto files that have entries that don't start with a `name`.
* Continuation of https://github.com/prometheus/procfs/pull/745


ArthurSens (approved):
rubber stamping, I don't really understand this codebase

PR #832, opened 2026-06-30
Fix invalid format strings

HNO3Miracle:
Fix invalid format strings reported by current Go vet.

This changes []uint64 error diagnostics to use %v instead of %q, and reports the original string field for parse errors in net_wireless and proc_cgroup instead of formatting the parsed integer zero value.

Tested:
- make test


SuperQ (approved):
Thanks!

SuperQ:
Good find on some of those error returns that return the parsed value instead of the raw value.

Issue #831, opened 2026-06-29
stat.go: /proc/stat processes field overflow causes parseStat() to fail entirely


jsczrangel:
### Affected Code

`stat.go` — `parseStat()` function, `processes` case (lines ~177-180):

```go
case parts[0] == "processes":
    if stat.ProcessCreated, err = strconv.ParseUint(parts[1], 10, 64); err != nil {
        return Stat{}, fmt.Errorf("%w: couldn't parse %q (processes): %w", ErrFileParse, parts[1], err)
    }
```

### Root Cause

The `processes` field in `/proc/stat` represents "number of forks since boot" and is stored as an `unsigned int` in the Linux kernel. On long-running systems (300+ days) with high process churn, this counter wraps around and `/proc/stat` reports a negative integer string:

```
processes -1948031333
```

`strconv.ParseUint` cannot parse negative values, causing the entire `parseStat()` function to return an error. This breaks ALL consumers that depend on `Stat()`, including:

- **node_exporter CPU collector** — zero CPU metrics exposed
- **node_exporter stat collector** — zero stat metrics exposed

### Impact

This affects any system with >4 billion process forks since boot. In containerized environments (Kubernetes), containers start and stop frequently, accelerating the counter overflow.

### Reproduction

```bash
# On a system with uptime >300 days:
cat /proc/stat | grep "^processes"
# processes -1948031333  ← negative!

# Any code calling fs.Stat() fails:
# "couldn't parse \"-1948031333\" (processes): strconv.ParseUint: parsing \"-1948031333\": invalid syntax"
```

Tested and confirmed failing on:
- node_exporter v1.3.1 through v1.8.2 (all versions)
- procfs v0.21.0 (latest as of June 2026)
- Linux kernel 5.x, system uptime 388 days

### Proposed Fix

Use `strconv.ParseInt` instead of `strconv.ParseUint` for the `processes` field. In Go, converting a negative `int64` to `uint64` produces the correct unsigned representation via two's complement:

```go
case parts[0] == "processes":
    v, err := strconv.ParseInt(parts[1], 10, 64)
    if err != nil {
        // Don't fail the entire Stat() — skip this field gracefully
        break
    }
    stat.ProcessCreated = uint64(v)
```

This handles both cases:
- Normal: `ParseInt("12345")` → `int64(12345)` → `uint64(12345)` ✓
- Overflowed: `ParseInt("-1948031333")` → `int64(-1948031333)` → correct `uint64` ✓

### References

[...]

PR #836, opened 2026-07-02
fix: tolerate wrapped signed process counters in /proc/stat

happysnaker:
## Summary
- accept wrapped `/proc/stat` `processes` counters that appear as signed 32-bit decimal strings
- keep the existing fast path for normal unsigned values
- add regression coverage for the wrapped-counter case

## Why
Issue #831 reports systems where `/proc/stat` exposes `processes -2045677862`, which currently makes `parseStat()` fail and breaks all `Stat()` consumers. The Linux kernel still formats this field from `total_forks` using `%lu`, but real systems have been observed to surface the wrapped value as a negative decimal string, including prior `node_exporter` reports in node_exporter#1882.

This patch keeps the usual `ParseUint` path for normal hosts, then falls back to parsing a signed 32-bit decimal and reinterpreting it as the original wrapped `uint32` value. That preserves the underlying counter instead of failing the whole `/proc/stat` parse.

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -143,6 +143,22 @@ func parseSoftIRQStat(line string) (SoftIRQStat, uint64, error) {
 	return softIRQStat, total, nil
 }
 
+func parseProcessCreated(value string) (uint64, error) {
+	if v, err := strconv.ParseUint(value, 10, 64); err == nil {
+		return v, nil
+	}
+
+	// Some kernels have been observed to expose the wrapped fork counter as a
+	// signed 32-bit decimal string (for example "-2045677862"). Recover the
+	// underlying uint32 value instead of failing the entire /proc/stat parse.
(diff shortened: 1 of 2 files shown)


happysnaker:
Follow-up after digging up an older public repro:

- prometheus/node_exporter#1882 includes a full `/proc/stat` dump from a real affected host
- that host reported `Linux centos 4.14.105-19-0016 x86_64` and exposed:

```
processes -2045677862
```

So while we still do not have a complete matrix of affected kernels, there is at least one concrete historical kernel string plus one concrete negative `processes` sample already in Prometheus issue history.

I think that is enough to justify the parser hardening here: `procfs` is already seeing this invalid-but-real input in the field, and today we fail the whole `/proc/stat` parse because of one counter line. This patch keeps the existing unsigned fast path, and only falls back for the observed wrapped negative form.

PR #837, opened 2026-07-02
fix(sysfs): accept delayed mdraid sync_completed state

### LINKED ISSUE #770: sysfs.Mdraids() fails if any MD raid array has a delayed check or resync

happysnaker:
## Summary
- treat `sync_completed=delayed` as a valid mdraid state instead of failing the entire `Mdraids()` parse
- keep numeric progress parsing unchanged for the normal `a / b` form
- add a focused regression test that builds a minimal mdraid sysfs tree with `sync_completed=delayed`

## Why
When multiple md arrays share physical devices, Linux can expose `sync_completed` as `delayed` while a check or resync is deferred. The current parser only accepts `none` or `a / b`, so one delayed array causes `sysfs.Mdraids()` to fail outright.

This change treats `delayed` like the existing non-progress sentinel `none`: there is no numeric completion fraction to parse yet, but callers still get the rest of the array metadata.

Closes #770

### DIFF
--- a/sysfs/mdraid.go
+++ b/sysfs/mdraid.go
@@ -166,7 +166,7 @@ func (fs FS) Mdraids() ([]Mdraid, error) {
 			}
 
 			if val, err := util.SysReadFile(filepath.Join(path, "sync_completed")); err == nil {
-				if val != "none" {
+				if val != "none" && val != "delayed" {
 					var a, b uint64
 
 					// File contains two values representing the fraction of number of completed
--- a/sysfs/mdraid_test.go
+++ b/sysfs/mdraid_test.go
@@ -16,6 +16,8 @@
 package sysfs
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


happysnaker:
Maintainer context: Linux can expose `sync_completed` as the literal `delayed` when a check/resync is deferred because another md array sharing the same devices is already syncing. In that state there is no numeric progress fraction to parse yet, but the rest of the array metadata is still valid. This patch therefore treats `delayed` the same way the parser already treats `none`: keep returning the mdraid entry instead of failing `Mdraids()` for the whole scrape.

PR #839, opened 2026-07-02
sysfs: add InfiniBandDevice API

### LINKED ISSUE #823: Export InfiniBandDevice parsing or add device filtering to InfiniBandClass
Problem

On NVIDIA B100+ systems, some ConnectX-7 NICs are NVSwitch/Fabric Manager-managed Limited PFs (VPD marked SMDL=SW_MNG). These NICs' hardware registers are owned by Fabric Manager, not the host. Reading any file under /sys/class/infiniband/\<dev>/ports/\*/counters/ triggers kernel ACCESS_REG(0x805) firmware errors:

`mlx5_core 0000:xx:00.0: mlx5_cmd_out_err:834:(pid 8567): ACCESS_REG(0x805) op_mod(0x1) failed,status bad operation(0x2), syndrome (0x9a6171), err(-22)`

9 lines per device per scrape, flooding dmesg. See node_exporter#3434 (https://github.com/prometheus/node_exporter/issues/3434) for full context.

Why InfiniBandClass() can't be used

InfiniBandClass() eagerly parses all devices — parseInfiniBandDevice → parseInfiniBandPort → parseInfiniBandCounters — in a single call. The caller has no opportunity to filter devices before counter files are read. By the time the caller iterates over the returned devices, the firmware errors have already been triggered.

The caller can identify FM-managed devices beforehand (by reading /sys/class/infiniband/\<dev>/device/subsystem_device, which is world-readable), but cannot skip them because parseInfiniBandDevice is unexported.

**Request**

Either:

**Option A:** Export parseInfiniBandDevice (rename to InfiniBandDevice):

`func (fs FS) InfiniBandDevice(name string) (*InfiniBandDevice, error)`

This lets callers list the directory, filter, and parse per-device.

**Option B:** Add filtering to InfiniBandClass:

`func (fs FS) InfiniBandClass(excludePattern ...string) (InfiniBandClass, error)`

Option A is preferable — It is purely additive — InfiniBandClass() is unchanged, so all existing callers are unaffected. New users gain per-device parsing without any migration cost.

Workaround in node_exporter

We are currently using go:linkname to call the unexported parseInfiniBandDevice as a temporary bridge, which is fragile and not sustainable.

happysnaker:
## Summary
- add `FS.InfiniBandDevice(name)` as a public per-device parser
- keep `InfiniBandClass()` unchanged
- add regression coverage that the single-device API matches the corresponding `InfiniBandClass()` entry

## Why
Issue #823 describes environments where callers need to filter InfiniBand devices before reading per-port counters. Today `InfiniBandClass()` eagerly parses every device, so callers have no way to skip known-problematic devices before counter reads happen.

This patch takes the additive route proposed in the issue: expose the existing per-device parser as a small public API. Existing callers keep using `InfiniBandClass()` unchanged, while new callers can now:

1. list `/sys/class/infiniband` themselves
2. filter devices using their own policy
3. call `fs.InfiniBandDevice(name)` only for devices they actually want parsed

That keeps the change narrow and avoids adding filter semantics to `InfiniBandClass()` itself.

Fixes #823

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -159,6 +159,12 @@ func (fs FS) InfiniBandClass() (InfiniBandClass, error) {
 	return ibc, nil
 }
 
+// InfiniBandDevice returns info for a single InfiniBand device read from
+// /sys/class/infiniband/<Name>.
+func (fs FS) InfiniBandDevice(name string) (*InfiniBandDevice, error) {
+	return fs.parseInfiniBandDevice(name)
+}
+
 // Parse one InfiniBand device.
 // Refer to https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-infiniband
 func (fs FS) parseInfiniBandDevice(name string) (*InfiniBandDevice, error) {
--- a/sysfs/class_infiniband_test.go
+++ b/sysfs/class_infiniband_test.go
@@ -344,3 +344,25 @@ func TestInfiniBandClass(t *testing.T) {
 		t.Fatalf("unexpected InfiniBand class (-want +got):\n%s", diff)
 	}
 }
+
+func TestInfiniBandDevice(t *testing.T) {
+	fs, err := NewFS(sysTestFixtures)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	class, err := fs.InfiniBandClass()
+	if err != nil {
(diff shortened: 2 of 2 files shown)


SuperQ:
Thanks for the inspiriation. I merged #854 which includes this with a few more features.

PR #841, opened 2026-07-10
fix(procfs): report raw values in parse errors

HNO3Miracle:
## Summary
- report the original mountinfo field when mount or parent IDs cannot be parsed
- report the original network socket field for sequence, port, state, queue, UID, inode, and drop parsing failures
- add regression coverage for every corrected error path

## Why
The existing errors format the parsed integer destination, which is zero-valued when parsing fails. This hides the malformed input from callers. Go 1.26.4 also reports these remaining integer arguments as invalid for the %q format verb during downstream package tests.

This follows up on #832, which fixed the same pattern in net_wireless.go, nfs/parse.go, and proc_cgroup.go but did not cover mountinfo.go or net_ip_socket.go.

## Testing
- make test
- make lint

### DIFF
--- a/mountinfo.go
+++ b/mountinfo.go
@@ -98,11 +98,11 @@ func parseMountInfoString(mountString string) (*MountInfo, error) {
 
 	mount.MountID, err = strconv.Atoi(mountInfo[0])
 	if err != nil {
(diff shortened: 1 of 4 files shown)


SuperQ (approved):
Thanks!

PR #842, opened 2026-07-10
fix(mdstat): parse (W), (J) and (R) component device flags

sueun-dev:
`evalComponentDevices` fills the WriteMostly/Journal/Replacement fields from `match[3]`, but `componentDeviceRE` only captures `(\([SF]+\))?`, so `(W)`, `(J)` and `(R)` never match and those fields are always false. Combined flags break too: the kernel prints each flag as its own parenthesis group and `(F)` can follow `(W)`, so for `sde1[4](W)(F)` the old group matches nothing and even `Faulty` is lost.

The kernel emits `(W)`, `(J)`, `(F)`, `(S)`, `(R)` one group per flag (the md.c lines already linked in `evalComponentDevices`), so the group now captures a run of them: `((?:\([FJRSW]\))+)?`.

Added `TestEvalComponentDevicesFlags` covering each flag plus the `(W)(F)` combination — it fails before this change (`WriteMostly`/`Journal`/`Replacement` stay false, and `Faulty` is false in the combined case) and passes after. This covers the WriteMostly/Journal/Replacement entries of the TODO at the top of `mdstat_test.go`.

Checked: `go test ./...`, `go vet ./...`, `gofmt -l` (the unrelated `TestFileDescriptorsLen` fails on this non-Linux machine with and without the change).

### DIFF
--- a/mdstat.go
+++ b/mdstat.go
@@ -27,7 +27,7 @@ var (
 	recoveryLinePctRE    = regexp.MustCompile(`= (.+)%`)
 	recoveryLineFinishRE = regexp.MustCompile(`finish=(.+)min`)
 	recoveryLineSpeedRE  = regexp.MustCompile(`speed=(.+)[A-Z]`)
-	componentDeviceRE    = regexp.MustCompile(`(.*)\[(\d+)\](\([SF]+\))?`)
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
Thanks!

PR #844, opened 2026-07-15
procfs: add KernelTainted() to parse /proc/sys/kernel/tainted

arunsrini3082:
Add KernelTainted() method on FS that reads /proc/sys/kernel/tainted
and returns the raw taint value along with each known flag parsed into
a KernelTainted struct.

This is required by prometheus/node_exporter to implement a tainted
kernel collector, which was originally proposed in
https://github.com/prometheus/node_exporter/pull/1567 but not merged
because parsing was expected to live in procfs first.

The struct exposes:
- Value: raw uint64 from /proc/sys/kernel/tainted
- Bits: slice of KernelTaintBit with Index (bit position), Flag (letter
  code), Description, and Set (whether currently active)

Reference: https://www.kernel.org/doc/html/latest/admin-guide/tainted-kernels.html

### DIFF
--- a/kernel_tainted.go
+++ b/kernel_tainted.go
@@ -16,12 +16,6 @@
 package procfs
 
 import (
-	"bytes"
-	"fmt"
-	"io"
-	"strconv"
-	"strings"
-
 	"github.com/prometheus/procfs/internal/util"
 )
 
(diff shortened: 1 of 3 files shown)


SuperQ:
Something is wrong with the testdata change. 
image

kernel_tainted.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return parseTainted(bytes.NewReader(data))
+}
+
+// parseTainted parses the content of /proc/sys/kernel/tainted.
+func parseTainted(r io.Reader) (KernelTainted, error) {
+	data, err := io.ReadAll(r)
+	if err != nil {
+		return KernelTainted{}, err
+	}
```
SuperQ: Why are you converting the `[]byte` returned by `util.ReadFileNoStat()` into a reader just to read it all again?

```suggestion
	return parseTainted(bytes.NewReader(data))
}

// parseTainted parses the content of /proc/sys/kernel/tainted.
func parseTainted(data []byte) (KernelTainted, error) {
```

kernel_tainted.go
```diff
@@ -0,0 +1,115 @@
...
+	{"T", "The kernel was built with the struct randomization plugin."},
+	{"N", "An in-kernel test such as a KUnit test has been run."},
+	{"J", "Userspace used a mutating debug operation in fwctl."},
+}
+
+// KernelTainted reads /proc/sys/kernel/tainted and returns the raw taint value
+// along with each known flag parsed into a KernelTainted struct.
+// See https://www.kernel.org/doc/html/latest/admin-guide/tainted-kernels.html
+func (fs FS) KernelTainted() (KernelTainted, error) {
+	data, err := util.ReadFileNoStat(fs.proc.Path("sys", "kernel", "tainted"))
```
SuperQ: There's a helper function that simplifies this greatly.
```suggestion
	data, err := util.SysReadUintFromFile(fs.proc.Path("sys", "kernel", "tainted"))
```

arunsrini3082:
@SuperQ Thanks for the suggestions! I've applied both and reconciled them — since SysReadUintFromFile already returns uint64, I updated parseTainted to accept uint64 directly (no more io.Reader or string parsing). The "io", "fmt", "strconv", and "strings" imports are now removed, and the test updated accordingly. Pushed in the latest commit.

SuperQ (approved):

PR #851, opened 2026-07-22
nfsd: ignore unknown metric lines instead of failing

locker95:
## What

`ParseServerRPCStats` currently returns an error when `/proc/net/rpc/nfsd` contains a metric key it does not know about. That fails the whole parse, so callers such as node_exporter export **no** NFSd metrics for a single new kernel field.

Skip unknown keys and keep returning the stats we already understand. Known keys (including `wdeleg_getattr`) are unchanged.

## Why

Linux 6.6+ added `wdeleg_getattr` (handled in #574). The same failure mode will hit again the next time the kernel adds a line. Other parsers in this repo already ignore unknown prefixes for forward compatibility (e.g. PSI).

Related discussion: prometheus/node_exporter#2799

## Test

```
go test ./nfs/ -count=1 -run TestNewNFSdServerRPCStats
```

Adds a fixture case with a synthetic `future_kernel_stat` line alongside known metrics.

### DIFF
--- a/nfs/parse_nfsd.go
+++ b/nfs/parse_nfsd.go
@@ -76,7 +76,10 @@ func ParseServerRPCStats(r io.Reader) (*ServerRPCStats, error) {
 		case "wdeleg_getattr":
 			stats.WdelegGetattr = values[0]
 		default:
-			return nil, fmt.Errorf("unknown NFSd metric line %q", metricLine)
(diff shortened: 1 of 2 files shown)


SuperQ:
This error is there to avoid corruption. If there are new things to parse, they need to be added.

PR #854, opened 2026-07-27
sysfs: add InfiniBand device listing and per-device parsing

rafal-szypulka:
## Summary

Add APIs that allow callers to enumerate InfiniBand device names without
reading device attributes or counters, and then parse selected devices
individually:

- `FS.InfiniBandClassDevices()` lists device names under
  `/sys/class/infiniband` without parsing them.
- `FS.InfiniBandDevice(name)` parses one selected device.
- `FS.InfiniBandClass()` is refactored to compose these APIs while preserving
  its existing behavior.

## Motivation

`InfiniBandClass()` currently parses every device eagerly. This prevents
callers such as node_exporter from applying device filters before sysfs
counter reads occur.

On NVIDIA systems with Fabric Manager-managed or otherwise restricted
Mellanox PFs, merely reading some counters can trigger repeated
`ACCESS_REG(0x805)` firmware errors. Filtering the returned
`InfiniBandClass` is too late because the reads have already happened.

Separating enumeration from parsing lets callers:

1. enumerate device names;
2. apply their own include/exclude policy;
3. parse only accepted devices.

The existing `InfiniBandClass()` API and default behavior remain unchanged.

## Tests

- Verify that device names can be enumerated even when the device directories
  do not contain readable attributes or counters.
- Verify that `InfiniBandDevice(name)` returns the same data as the
  corresponding `InfiniBandClass()` entry.
- Linux test package cross-compiles successfully.

Related to #823 and prometheus/node_exporter#3434.

## Existing work

#839 already proposes the additive `InfiniBandDevice(name)` API. This change
includes that capability and additionally provides list-only enumeration so
callers do not need to reproduce procfs sysfs path handling outside this
package.

### DIFF
--- a/sysfs/class_infiniband.go
+++ b/sysfs/class_infiniband.go
@@ -136,9 +136,9 @@ type InfiniBandDevice struct {
 // The map keys are the names of the InfiniBand devices.
 type InfiniBandClass map[string]InfiniBandDevice
 
-// InfiniBandClass returns info for all InfiniBand devices read from
-// /sys/class/infiniband.
-func (fs FS) InfiniBandClass() (InfiniBandClass, error) {
+// InfiniBandClassDevices returns the names of the InfiniBand devices found in
+// /sys/class/infiniband without reading any device attributes or counters.
+func (fs FS) InfiniBandClassDevices() ([]string, error) {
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
Thanks!

PR #855, opened 2026-08-01
procfs: fix parseStat comment and reuse line variable

### LINKED ISSUE #845: A small nitpick and a redundant method call in func parseStat()
**A small nitpick:**

At line 181, https://github.com/prometheus/procfs/blob/master/stat.go#L181, the comment is passed as:
`// parseStat parses the metrics from /proc/[pid]/stat.`

however, parseStat parses the system-wide /proc/stat file rather than /proc/[pid]/stat, so the comment appears to be inaccurate.

**Repetitive method calling:**
In the same function:
```
buf := make([]byte, 0, 8*1024)
	scanner.Buffer(buf, 1024*1024)

	for scanner.Scan() {
		line := scanner.Text()
		parts := strings.Fields(scanner.Text())
```
(Line https://github.com/prometheus/procfs/blob/master/stat.go#L197)
Since `line `already contains the current scanner text, this could instead be:
`parts := strings.Fields(line)`

They both are very small issues but worth noting.

If accepted I would like to make a PR for these changes.

zanarellidev:
## Summary
Fixes the inaccurate `parseStat` comment (it parses `/proc/stat`, not `/proc/[pid]/stat`) and reuses the `line` variable for `strings.Fields`.

Fixes #845

## Test plan
- [x] `go test -run 'TestStat$' .`

### DIFF
--- a/stat.go
+++ b/stat.go
@@ -178,7 +178,7 @@ func (fs FS) Stat() (Stat, error) {
 	return procStat, nil
 }
 
-// parseStat parses the metrics from /proc/[pid]/stat.
+// parseStat parses the metrics from /proc/stat.
 func parseStat(r io.Reader, fileName string) (Stat, error) {
 	var (
 		scanner = bufio.NewScanner(r)
@@ -194,7 +194,7 @@ func parseStat(r io.Reader, fileName string) (Stat, error) {
 
 	for scanner.Scan() {
 		line := scanner.Text()
-		parts := strings.Fields(scanner.Text())
+		parts := strings.Fields(line)
 		// require at least <key> <value>
 		if len(parts) < 2 {
 			continue


stat.go
```diff
@@ -178,7 +178,7 @@ func (fs FS) Stat() (Stat, error) {
 	return procStat, nil
 }
 
-// parseStat parses the metrics from /proc/[pid]/stat.
+// parseStat parses the metrics from /proc/stat.
```
SuperQ, line 181: Why is this comment incorrect?

zanarellidev: Because `FS.Stat()` in `stat.go` reads `/proc/stat` (system-wide metrics via `fs.proc.Path("stat")`), whereas `/proc/[pid]/stat` is the per-process stat file parsed separately in `proc_stat.go` (`Proc.Stat()`). The docstring in `stat.go` was previously copied from `proc_stat.go`.

SuperQ: Oh, sorry, I thought we used this parser for both modes of operation.

PR #862, opened 2026-08-26
Add resctrlfs package

rtreffer:
Parse the monitoring data of the resctrl filesystem, the kernel interface to Intel RDT, AMD PQoS and ARM MPAM. All vendors share the same driver and file layout, so one parser serves all vendors.

MonData returns the per domain counters of the root control group (llc_occupancy, mbm_total_bytes, mbm_local_bytes). A counter is nil when the hardware lacks the feature or reports "Unavailable". L3MonInfo exposes the monitoring capabilities from info/L3_MON.

This patch only adds read support for monitoring data.

Signed-off-by: René Treffer <treffer@measite.de>

## Reviewers

**AI disclosure**: This code was written with the help of AI.
**I did read through the code and everything looks as expected**.

I have tested this with a new `node_exporter` collector (`resctrl`).


SuperQ (approved):
Seems legit.

Dealing with `Unavailable` is tricky. But I think returning `nil` is fine.

PR #464, opened 2022-09-04
Replace float64 with *float64 for proc_netstat, proc_snmp and proc_snmp6

nikakis:
This PR is related to this issue #434 which is needed for https://github.com/prometheus/node_exporter/issues/2336

Signed-off-by: Nikos Kakavas <nikakis@protonmail.com>

PR #863, opened 2026-08-27
netstat: model legacy TcpExt counters removed from recent kernels

neoLsH:
## What

Add 10 legacy `TcpExt` counters to the `/proc/net/netstat` model and parser:
`PAWSPassive`, `TCPPrequeued`, `TCPDirectCopyFromBacklog`,
`TCPDirectCopyFromPrequeue`, `TCPPrequeueDropped`, `TCPLoss`,
`TCPFACKReorder`, `TCPForwardRetrans`, `TCPHPHitsToUser`,
`TCPSchedulerFailed`.

## Why

These counters are absent from the current kernel's `TcpExt` table (none of
them are in linus/master `include/uapi/linux/snmp.h`), but older kernels still
print them in `/proc/net/netstat`. node_exporter is switching its netstat
collector to these parsers (prometheus/node_exporter#3796), and would
otherwise silently drop these metrics for users whose
`--collector.netstat.fields` regex selects them.

## Behavior

- Purely additive: the fields are `*float64` and stay `nil` when the kernel
  does not print them (all modern kernels), matching the optional-field
  semantics from #464.
- Tests cover both the "older kernel prints them" and "modern kernel doesn't"
  cases.

## Out of scope

- `IcmpMsg` / `Icmp6Msg` per-type counters: the kernel only prints non-zero
  `InTypeN`/`OutTypeN` entries, so the set is host-dependent and open-ended —
  a fixed struct can't cover it (a map, like `NetDev`, or an unknown-keys
  escape hatch would). Separate follow-up.
- The `StatValue` refactor floated in prometheus/node_exporter#3796: changing
  all netstat/snmp field types would be a breaking API change for existing
  consumers; separate discussion.

Note on scope: this PR closes the *backward* gap (counters older kernels print that
were never modeled). procfs's `TcpExt` also lags the *current* kernel in the other
direction (e.g. `TCPDelivered`, `TCPAckCompressed`, `TCPZeroWindowDrop`,
`TCPBacklogCoalesce`, `BeyondWindow`, `TCPTimeoutRehash`, `TCPDSACKRecvSegs`,
`TCPMigrateReq*`, `TCPAO*` are printed by modern kernels but not modeled). That
forward lag is procfs's ordinary upkeep, pre-exists this change, and affects all
consumers; it is deliberately out of scope here rather than bundled in.

### DIFF
--- a/proc_netstat.go
+++ b/proc_netstat.go
@@ -146,6 +146,20 @@ type TcpExt struct { // nolint:revive
 	TCPMTUPFail               *float64
 	TCPMTUPSuccess            *float64
 	TCPWqueueTooBig           *float64
+
+	// Legacy TcpExt counters dropped from recent kernels' TcpExt table;
+	// modeled so consumers retain metric parity with older kernels that
+	// still print them in /proc/net/netstat.
+	TCPLoss                   *float64
+	PAWSPassive               *float64
+	TCPForwardRetrans         *float64
+	TCPSchedulerFailed        *float64
+	TCPPrequeued              *float64
+	TCPDirectCopyFromBacklog  *float64
+	TCPDirectCopyFromPrequeue *float64
+	TCPPrequeueDropped        *float64
+	TCPFACKReorder            *float64
+	TCPHPHitsToUser           *float64
 }
 
 type IpExt struct { // nolint:revive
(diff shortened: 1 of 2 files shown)


SuperQ:

SuperQ (approved):

PR #866, opened 2026-09-02
sysfs: handle cpufreq files with <unknown> values

### LINKED ISSUE #304: Handle cpufreq files with <unknown> values.
Some systems produce `<unknown>` values for cpufreq files.

See: https://github.com/prometheus/node_exporter/issues/1710

neoLsH:
Fixes #304

## Description

The kernel can write the sentinel `<unknown>` to cpufreq files (e.g. `scaling_cur_freq`) when the current frequency cannot be sampled (reported in #304, with the downstream symptom in prometheus/node_exporter#1710: the whole cpufreq collector stops exporting when a single CPU hits the sentinel).

Currently `parseCpufreqCpuinfo` treats that as a parse error and fails the whole read, so `SystemCpufreq()` returns an error and consumers lose **all** cpufreq metrics, not just the one unavailable value.

This change treats the exact `<unknown>` sentinel like a missing file: the affected `*uint64` field is left `nil`, which is the established contract for unavailable values in `SystemCPUCpufreqStats` (e.g. `cpuinfo_cur_freq` is only readable as root, so absent files already produce `nil`, and consumers check for `nil`).

Detection mirrors the `errors.As` idiom already used in `net_class.go` for `EINVAL`: only a `*strconv.NumError` whose input is exactly `<unknown>` is skipped, so any other malformed value still returns an error.

## Testing

- `TestSystemCpufreq` now covers the sentinel: the `cpu1` fixture's `scaling_cur_freq` contains `<unknown>` and must parse to `ScalingCurrentFrequency: nil` (the missing-file path is still covered by `cpu0`'s absent `cpuinfo_cur_freq`).
- Verified locally with `GOOS=linux go build ./...`, `go vet`, `go test -c`, `gofmt`, and by extracting the modified `fixtures.ttar` with `./ttar -x`. The Linux-only sysfs tests themselves can't run on my macOS dev machine — looking forward to the CI results.

### DIFF
--- a/sysfs/system_cpu.go
+++ b/sysfs/system_cpu.go
@@ -16,6 +16,7 @@
 package sysfs
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


neoLsH:
@SuperQ friendly ping — treats the kernel's `<unknown>` sentinel in cpufreq files like a missing file (nil) so one CPU's unsampleable frequency doesn't fail the whole SystemCpufreq read. Fixes #304 and the downstream node_exporter#1710. CI is green.

